PR #10, opened 2016-11-21
Boot from rebuilt ITS

larsbrinkhoff:
This makes the build boot from the rebuilt ITS.  It also includes the new exec DDT and NSALV.

I had to update SYSTEM; to version 1647 from the PI distribution, or else there would be a crash in the final dump of the file system.


larsbrinkhoff:
I'm not sure if the DB configuration is up-to-date with the latest changes in CONFIG 202.  I just copied it from NEWSYS;CONFIG 193.

Anyway, I figure we will be updating it later anyway.

larsbrinkhoff:
@eswenson1 Do you think this is a good format for a pull request?

larsbrinkhoff:
You were there, so you would know better than I whether certain conventions make sense or not.  I'm not against having the binary in SYSTEM;.

PR #24, opened 2016-11-26
Added MLDEV/MLSLV devices. 

eswenson1:
These allow access to the file system using the device DB: both locally and from remote systems.


larsbrinkhoff:
Looks good, thanks!

build/build.tcl
```diff
@@ -394,6 +394,16 @@ respond "*" ":link sys;ts f,sys;ts name\r"
...
 
+respond "*" ":midas device;atsign mldev_sysen2;mldev\r"
+expect ":KILL"
+
+respond "*" ":link device;jobdev db,device;atsign mldev\r"
+
+respond "*" ":midas device;atsign mlslv_sysen2;mlslv\r"
+expect ":KILL"
+
+respond "*" ":link device;tcp syn123,device;atsign mlslv\r"
```
larsbrinkhoff: So MLSLV already works over TCP/IP? I guess we can write an MLDEV for Linux, then!

eswenson1:
Yes. This would be pretty cool to have!

larsbrinkhoff:
Note.  In ITS 724, MLDEV is served by SYS; ATSIGN OTHER.

larsbrinkhoff:
Wrong.  Make that ITS 768.  ITS 724 has NETP=0 and thus no IMP or MLDEV.

PR #25, opened 2016-11-26
Added ITS support for CHTN and CFTP (Telnet and FTP over Chaosnet).

eswenson1:
To actually use these programs, however, the emulator needs to get
updated to support Chaosnet (over UDP, for example).


larsbrinkhoff:
Thanks!

PR #30, opened 2016-11-27
Added SRCCOM program and documentation.

eswenson1:
SRCCOM is a source/binary comparison program.
SRCCOM also support source merges.


larsbrinkhoff:
Thanks!

PR #32, opened 2016-11-27
Build the DIR device.

larsbrinkhoff:


eswenson1:
Is a link to the ATSIGN DIRDEV not needed -- something like SYS; JOBDEV DIRDB?

larsbrinkhoff:
Apparently not.  I tested it like this:

    ^R dir:sys;only links

and that worked.

eswenson1:
Ok, thanks. On ES-ITS I have that link -- but I can't remember where I saw to do that. I'll merge as is, and if I can find a reason for it, I'll enter a ticket.

PR #57, opened 2016-12-01
Added PWORD and PANDA. Set up initial password database.

eswenson1:
Did not link sys;atsign hactrn to sysbin;pword bin, so logins
are not required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 EMULATOR ?= simh
 
-SRC = system syseng sysen1 sysen2 sysnet kshack dragon channa midas _teco_ emacs rms klh syshst sra mrc ksc cstacy gren bawden emacs1 _mail_ l lisp liblsp libdoc comlap lspsrc nilcom rwk inquir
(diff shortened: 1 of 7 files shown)


larsbrinkhoff:
@BobKerns wrote PWORD.  Earliest timestamp I find is 1978-03-27, for ZZ; PWORD TTD.  Three days later there is a RWK; PWORD 1143.  The high version number suggests it had been in the works for a while.

BobKerns:
I don't recall how much elapsed time passed before putting it into use. Some of the activity would be due to evolving discussions of exactly what behavior it should have, some went to making it bulletproof, and some toward trying to make it simultaneously easy to use for someone encountering ITS for the first time, but not too different from using DDT—obviously incompatible requirements.

And I learned early to save my work often—in the modern world, I make frequent git commits. It came in handy yesterday when I wanted to revert some temporary code.

Still, "in the works for a while" is undoubtedly an accurate observation.

PR #63, opened 2016-12-01
Rebuild EMACS

larsbrinkhoff:
This builds EMACS, broken up into a few steps:

- First, move the EMACS binaries from the minsys tape, to the sources tape.  They are still located in the bin directory.

- Then, add a minimum amount of :EJ files to rebuild EMACS :EJ, PURIFY :EJ, and EINIT :EJ, using the EINIT library invoked from EMACS.  The corresponding source files are also added.

- The dumping code depends on the AUX library, so regenerate that.  Finally, load EMACS; EMACS TECO into TECO and run the dumping code.

One last thing to fix, is that the resulting EMACS thinks it's version 163.  Nothing has changed from 162, so I'd like to keep that version.


larsbrinkhoff:
@eswenson1 Please hold off merging this until I figure out how to set the version.

larsbrinkhoff:
There, it's finished.

Note that EMACS isn't completely done yet.  I'll add all libraries and documentation later.

PR #72, opened 2016-12-03
Added SEND/REPLY, as replacements for DDT's :SEND.

eswenson1:


larsbrinkhoff:
Does a binary in the SYS directories take precedence over a built-in colon command?

eswenson1:
I don't think so. You have to disable DDT's :SEND in order to use this. See the doc on SEND for how. If I recall, you have to change the value of ..SNDFLG. I'm not at a computer now to check.

-- Eric

eswenson1:
From INFO; SENDER >:

If you want to use SENDER, then what you do is put in ..SNDFLG any non-
zero value.  This tells DDT that ":SEND" should look for the program
named SEND, and not use the built-in one.  Then you just need a link or
translation of TS SEND to the binary (KP;SENDER BIN on MC), and you're
set.

Issue #68, opened 2016-12-02
DIRED - directory editor


larsbrinkhoff:
SYSEN2; DIRED 187  
.INFO.; DIRED ORDER  
.INFO.; DIRED INFO  
.INFO.; DIRED RECENT

This is unrelated to DIRED in EMACS.

PR #134, opened 2016-12-05
Added DIRED (not the one from EMACS, but the MIDAS one).

eswenson1:


eswenson1:
This PR resolves issue #68

Issue #93, opened 2016-12-04
CROCK - running analog clock


eswenson1:
SYS1; TS CROCK
SYSEN1; CROCK 1008

larsbrinkhoff:
Oh, I want this!

larsbrinkhoff:
It also wants SYSEN1; SYSTEM >, but that's in SYSENG.

When it runs, there's an error:

    ILOPR; 320>>XCT 437(10)   452/   33

eswenson1:
The PI distrubution binary runs for me without ILOPR.

larsbrinkhoff:
I get the same error when I copy the PI binary to the DB system.  Did you run it in PI or DB?

larsbrinkhoff:
I'm running in SIMH, so maybe that's why.

larsbrinkhoff:
Done in #149.

PR #149, opened 2016-12-05
CROCK - analog watch

larsbrinkhoff:
.INSRTs SYSEN1; SYSTEM >, so needs that to link to SYSENG; SYSTEM >.

I'd like to fix the source code to .INSRT the right file, but I'll submit the unmodified file for now.


larsbrinkhoff:
Note that the system clock has to be set for CROCK to work.

Fixes #93.

PR #157, opened 2016-12-06
Add contribution guidelines.

larsbrinkhoff:
@eswenson1 I'd like you comments on this.  Anything to change, or to add?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,19 @@
+## Contribution Guidelines
+
+- Do not add files with personal, private data.
+
+- Initially, we try to stay close to the "Public ITS" distribution.
+
+- A pull request should adress one, and only one, concern.  For
+  example, adding a program or bug fix.
+
+- Individual commits within a pull request should consist of atomic
+  changes.  A single commit should neither be in a partial state, nor
+  should it contain multiple unrelated changes that could be split
+  into separate commits.
+
+- Add or update documentation in the pull request.  That includes both
+  README files in the repository, and files put into the ITS disk
+  image.
+
+- Do use the git history rewriting operations to modify pull request.


eswenson1:
Seems ok to me. I'd probably include something to the effect that initially we are trying to mirror the PI distribution in terms of organization. Later, we may wish to start deviating/reorganizing to make for a saner system. But for now, it makes things easier if we can compare the DB and PI builds (file-by-file, directory-by-directory) to measure progress and determine when we have succeeded.

larsbrinkhoff:
I added a short bullet point about that.

You are free to change, add to, or merge this PR if you like.

eswenson1:
Are you ready to have me merge, once travis is happy?  I'm ok with this.

larsbrinkhoff:
Yes, merge away!

larsbrinkhoff:
> we are trying to mirror the PI distribution in terms of organization. Later, we may wish to start deviating/reorganizing to make for a saner system. But for now, it makes things easier if we can compare the DB and PI builds (file-by-file, directory-by-directory) to measure progress and determine when we have succeeded

I feel it's better to do the updates as we go, because the amount of work to get to a perfectly finished state replicating PI is so large.  We may end up never getting there.  And then the reorganizations will never be done.

I'd suggest we try to eat the cake and have it too.  Add a commit that records the files in a state that replicates the original system, and then another commit with updates.  It can be an individual choice whether this update is done right away, or much later.

eswenson1:
I actually think we should mirror the PI distribution with respect to all the executables and basic functionality/doc. I don't think it is necessary to match all other files. There is a ton of useless stuff in the PI distribution, as well as lots of old versions of sources and binaries. I don't think we should duplicate that. Also, i think we'll know when we are at the point of reorganizing. Once we've produced all the executables and binaries that we can, and we've included all relevant doc, we can declare ourselves ready to clean up. We should enter tickets for all cleanup to document our intent and expectations. 

Anyway, that's just my opinion. I'm ok with not doing it that way if you feel strongly about it.

-- Eric

larsbrinkhoff:
I do have some strong feelings, but I'm fine with you working that way.  I'd like to do some cleanups on the way, but it's not urgent so I can wait for a while.

Issue #6, opened 2016-11-18
Use xfiles to build


larsbrinkhoff:
Currently, the build is scripted entirely from the host, using **expect**.

Change this to build as much as possible with xfiles.  In some cases xxfile may be necessary.

PR #160, opened 2016-12-06
Build with xfiles

larsbrinkhoff:
**WORK IN PROGRESS** - do not merge.

Note that only those programs which build using only input to DDT can use xfiles.  For now, I have made a single commit for each program, but I think I'll make it one big commit later.

Comments, @eswenson1?


eswenson1:
I'm not crazy about this scheme. It was much easier to go to one place (build/build.tcl) to do all the build steps (yes, I know, one big global function isn't great and modularization is good).  Now, there all the individual pieces are spread out over oodles of files. Some stuff is build in build/build.tcl and other stuff is built in xfiles. You have to search to find out where something is built. Sometimes there are files that happen to be part of bootstrapping some piece of software, but are actually independent prerequisites and might be used for other pieces of software. For example, to build INQUIR, I needed a bunch of FASL files, which I built from source. However, while these are prerequisites for getting INQUIR built, they really are completely independent of INQUIR and other programs also depend on them. Putting them in an INQUIR-specific build file is arbitrary.  Hiding prerequisites in some XFILE obscures their independence. 

Also, I don't like the fact that programs that require input other than from DDT have to be built using a different mechanism.  (Is this really true, by the way?)

In any case, I understand that it is nice that parts of the build can be done in ITS. I just don't like that we have to split things up into those that can be built in ITS and those that can't, and that all the parts that can are split up into so many files.  I'll get over it, though.

larsbrinkhoff:
> I just don't like that we have to split things up into those that can be built in ITS and those that can't

Fair enough, I agree with that.  I could refrain from merging anything until this is fixed.

>  (Is this really true, by the way?)

I haven't tested it myself, but it's a documented limitation of :XFILE: *"Note that the execute file affects only DDT;  it does not also supply input to other programs.  Programs that normally read input from the terminal will continue to do so, even when invoked by an execute file."*

XXFILE is supposed to solve this, by running a job in a pseudo TTY and sending all input there.  But I haven't fully understood how to make it work yet.

larsbrinkhoff:
Fixes #6.

larsbrinkhoff:
@eswenson1 I'm still updating this branch with added programs.

I have now figured out how to handle terminal input when programs are built: Add a filename translation from TTY: to a file containing the input.

eswenson1:
Very cool. Hadn't occurred to me to use DDT translations for this. Good thinking, and a bit obvious in retrospect!

larsbrinkhoff:
I spoke too soon.  The .TTYMAC directive probably opens the TTY every time it's used.  So for every call, it should get the first line of the input file.  This doesn't work if .TTYMAC is called several times, which is the case for many programs.

Also, I don't know why, but EMACS doesn't cooperate at all with a translation for TTY:

larsbrinkhoff:
Almost there, it's mostly just the Lisp stuff left.  Apparently, it's hard to do Lisp dumps from an xfile.

I think calling all the xfiles from one big ALL XFILE will speed up the build.  As it is now, we spend much of the build time in the `type` procedure, waiting for ITS to echo characters back.

larsbrinkhoff:
Now that XXFILE seems to work, maybe it's time to revisit this.  What do you say @eswenson1, would you oppose moving most of the build scripting from TCL to XXFILE?

eswenson1:
Yes, we should. But I tried a couple of times to do this unsuccessfully. Can you get some build with input requested (like a Midas build that prompts) switched over and committed and I’ll copy for my stuff?

-- Eric

larsbrinkhoff:
The gains as I see it:
- Build ITS programs from ITS, not a foreign scripting language.
- Speed up the build.  Having expect gently feed a command line to ITS takes a few seconds maybe?
- Uniform scripting language, not split between x(x)files and expect.

eswenson1:
For these one-lines, does it really speed up the build?  ITS has to spawn a STY for each invocation of XXFILE, right?

larsbrinkhoff:
Right, using XXFILE is likely not much faster than expect.  But for those one-lines, I think :XFILE can do the job, and I think that's faster.

larsbrinkhoff:
I think this is too old to be useful now.  It's still something worth persuing, but if so let's use a clean slate.

PR #161, opened 2016-12-07
Rearrange some binaries and directories

larsbrinkhoff:
Minor cleaning.


eswenson1:
The diff for build.tcl suggests you will create sys3 twice.

larsbrinkhoff:
God catch, thanks.  I've updated the last commit.

PR #184, opened 2016-12-08
Generate some EMACS documentation

larsbrinkhoff:
Runs "Document EMACS" in EINIT to use the ABSTR library to write documentation files into EMACS;

This requires some other libraries to be in place first.


larsbrinkhoff:
@eak Do I understand correctly that you can't load both PURIFY and IVORY into the same EMACS instance?

PR #185, opened 2016-12-08
Copy everything in the "user" directory to ITS.

larsbrinkhoff:
But exclude it from version control.  `user/.gitignore` is just a placeholder; whithout it, git will remove the empty directory.  (Just like the ITS Salvager!)

You can put your init files there, or any other (personal) data you want in the system, but not in the repository.


larsbrinkhoff:
I have been missing my LARS; LARS EMACS.

eswenson1:
I don't have any issue with this change.  However, I find it very frustrating that I don't have my EJS directory automatically created when I boot a new version, and some of my INIT files. I frequently boot with a custom KLH10 INI file that mounts a tape from ES-ITS, and then I use load (manually) to load up my EJS directory. While this change is nice, we end up getting things in a USER directory -- and you still have to create your home directory and populate it -- unless, of course, you log in as "user".  I guess I could get used to that.

larsbrinkhoff:
> we end up getting things in a USER directory

No, you get every directory that is located inside the Unix `user` directory.  So you can mkdir `user/ejs` and add your files there.  This works just like the `src`, `doc`, and `bin` directories.  Their subdirectories appear as UFDs in ITS.

> Can I merge this?

Yes.

We should probably have a way to signal that a PR is ready for merging.

eswenson1:
Thanks. Merged. The @ PR is ready to merge too.

larsbrinkhoff:
@eswenson1 I have my user files checked into an orphan branch called user/lars.  That way, they don't clutter the master branch, but are still available in the repository.

PR #45, opened 2016-12-09
STINK - Linker

larsbrinkhoff:
SYS1; TS STINK  
SYSEN2; STINK 200  
.INFO.; STINK DOC
.INFO.; STINK RECENT


larsbrinkhoff:
I disassembled the built 200 with the binary 201.  It was a really small fix, so I'll include it in this pull request.

Fixes #189.

PR #191, opened 2016-12-10
Added LUSER: requests help from registered list of logged-in users.

### LINKED ISSUE #98: LUSER - Provide help to users
SYS1; TS LUSER => SYSBIN; LUSER BIN
SYSENG; LUSER 85

eswenson1:
Resolves #98.  Note, that with the unmodified sources, it is pretty useless, since none of the hard-coded lists users is likely to be logged in.  Same with the hard-coded list of machines.


larsbrinkhoff:
If you like, you can add new programs somewhere in the middle of `build.tcl`.  I try to find a section with similar programs.  That way, it's less likely there will be a merge conflict.

PR #193, opened 2016-12-10
Added PRUFD.

### LINKED ISSUE #101: PRUFD - list files on disk volume
SYS1; TS PRUFD
SYSEN2; PRUFD 10

eswenson1:
Resolves #101.  Includes patch to replace LPT device with TTY device (version in PI must have been patched).


eswenson1:
Ok. I'll redo. However, you realize it won't run without the patch, right?

larsbrinkhoff:
You can have two commits in this pull request.  The first with unchanged files, the second with the fix.  That way, it will work when the pull request is merged.

eswenson1:
Ok. Will do.

-- Eric

eswenson1:
There are now two commits. The first just builds PRUFD from sources. The second patches it to use TTY: rather than LPT:  I entered #195 for the second commit.  #101 is for the first commit.

larsbrinkhoff:
The PRUFD source file seems to be broken.  See the end:  
https://github.com/PDP-10/its/blob/master/src/sysen2/prufd.10#L404

PR #197, opened 2016-12-10
11SIM - PDP-11 emulator.

larsbrinkhoff:
There are rumors there was two binaries, TS PDP11 and TS PDP45, but I see no evidence for the latter in the snapshots.

The 11/45 emulator is built by setting the symbol `45P` to `1`.  Is it possible to do that without changing the source file?


eswenson1:
It looks like 45P is a assembly-time constant rather than a memory location in the image. Therefore, it is not generally possible to change this post-assemble, since that constant could be used to conditionally assemble very different code. I haven't looked to see all the places where this constant is actually used -- if only trivial changes results from setting this one way or the other, those affected memory locations could be patched. On the other hand, perhaps it is better to query for the value of that constant at assembly time so that the same source can be assembled into two separate binaries.

larsbrinkhoff:
Right, good idea, and consistent with what other programs do.

larsbrinkhoff:
Tested by :LOADing LIVE BIN assembled from BUDD; LIVE PALX and single-stepping it.

larsbrinkhoff:
Running LIVE BIN also works, once I figured out how to operate it.  Editing the grid works fine, but hitting `g` doesn't do anything.  Anyway, a good test that the PDP-11 emulator works.

Thanks @philbudne

philbudne:
A simulation under a simulation under a simulation!
Odd that it doesn't "run".... Where can I see my old home (BUDD) dir?

larsbrinkhoff:
> Where can I see my old home (BUDD) dir?

We only add files on a case-by-case bases as needed to build or document programs, as there may be privacy concerns otherwise.

Since you kinda requested them, here are your files:
https://github.com/PDP-10/its/tree/user/budd

larsbrinkhoff:
@budd Your program works in the 11/45 version of the emulator.

philbudne:
ISTR I ran it on my 11/03, so maybe something non-basic snuck in.
I couldn't find any EIS (ASH, ASHC, MUL, DIV) instructions....

p

PR #200, opened 2016-12-10
Make the KLH10 dskdmp script start automatically.

larsbrinkhoff:
This is to make it easier for users to start ITS with the current set of scripts.


larsbrinkhoff:
For some reason, I now get a lot of MFDCLB in DSKDMP.

larsbrinkhoff:
This won't do.  I'll have to change the usage instructions instead.

In my experience KLH10 is more finicky like this than SIMH.

PR #244, opened 2016-12-12
Added LOSS device.

### LINKED ISSUE #243: LOSS device
DEVICE; JOBDEV LOSS
DEVICE; JOBDEV LOS => DEVICE; JOBDEV LOSS
SYSENG; LOSS 9

eswenson1:
Resolves #243


README.md
```diff
@@ -151,6 +151,7 @@ There's a [DDT cheat sheet](doc/DDT.md) for Unix users.
    - LOADP, displays system load.
    - LOCK, shut down system.
    - LOOKUP, looks up user info in INQUIR database.
+   - LOSS (device)
```
larsbrinkhoff: What does it do?

eswenson1:
Yeah. I did that because I didn't know what it did. I thought it was a means of analyzing a .LOSE, but then I had my doubts. It's probably going to require that we study the source more.

eswenson1:
I've looked at the code a bunch and don't really understand what it does and how it works. @larsbrinkhoff, if you have a chance, can you look at it?

larsbrinkhoff:
I looked at it too, but I'm at a LOSS.

eswenson1:
:-)  When I try to run it as a normal job, it fails immediately trying to open the BOJ device. I don't understand the BOJ device -- even after having read RMS' writeup on it (SYSDOC; JOB 100).  My guess is that you cannot run it as a normal job -- and I don't know how to debug a JOBDEV handler (like JOBDEV LOSS) when it is invoked "normally" (as a device).

eswenson1:
This is a device to translate error codes to human-readable strings. For example:

^R loss:error 5

prints DIRECTORY FULL.

other FN1s are supported besides ERROR. See the source for the supported set.

eswenson1:
Also FN1 = INT and IOC  are supported.

larsbrinkhoff:
Thanks!  There's also STATUS.  And it looks like it should be possible to make a file listing of the LOSS: device, but I don't get anything.

PR #259, opened 2016-12-13
List of ITS machines.

larsbrinkhoff:
Unsure whether emulators should be included.  SV and UP got an honorable mention for being online for many years.


larsbrinkhoff:
Fixes #181.

PR #317, opened 2016-12-18
UNTALK - split-screen comm-link program

larsbrinkhoff:
SYS2; TS UNTALK
AI: GREN; UNTALK 87


larsbrinkhoff:
I made a mistake.  It should be named TS UNTALK, or else it will not connect two parties together.  Fixed in #677

PR #334, opened 2016-12-19
Update README.md.

larsbrinkhoff:


larsbrinkhoff:
Fair enough.  Would it be better to remove all periods?

eswenson1:
Probably. But clearly not a very important change.

PR #347, opened 2016-12-19
Fix filenames

larsbrinkhoff:
Change .INSRT filenames to point to the current locations of files.


larsbrinkhoff:
Good point.

eswenson1:
I think, perhaps, we should create a SYSEN3 directory and put all the sources we extract from random archives into that directory. We should probably also move some of the random included files from KLH, KSC, etc. directories into SYSEN3.  It's my opinion, that if a binary is in SYS, SYS1, SYS2, or SYS3, then the source should be in SYSENG, SYSEN1, SYSEN2, or (SYSEN3).

eswenson1:
What do you think of my idea of creating SYSEN3 and putting all these sources there?  I think we should also consolidate all the klh/ksc/mrc/gren/etc. INSRT files too, but this will require us to rename some of them to avoid collisions (e.g. MACROS).

larsbrinkhoff:
> I think, perhaps, we should create a SYSEN3 directory and put all the sources we extract from random archives into that directory.

Yes, good idea.

> We should probably also move some of the random included files from KLH, KSC, etc. directories into SYSEN3.

I think it's part of its historical charm that some files reside in individual user directories.  On the other hand, it makes for a more maintainable and logical system if all system sources are in the SYSEN directories.  I suppose the FOO; MACROS > would have to get some special treatment.  Stay with the users, or go in e.g. SYSMAC?

> It's my opinion, that if a binary is in SYS, SYS1, SYS2, or SYS3, then the source should be in SYSENG, SYSEN1, SYSEN2, or (SYSEN3).

I've been thinking the same thing.

eswenson1:
While I like the historical charm, I don't like the ugliness of having INSRT files in all these random directories, and, quite frankly, that charm is really only appreciated by a few.  If you bring up an ITS system, it is a big ugly to have bawden, eak, gren, ksc, mrc, rms, sra, alan, cstacy, ecc, gz, klh, and rwk directories (I know host of these people and hope that I won't offend them greatly by my blasphemous comment).  There is no logical reason why anyone would look in these directories for system sources, and the fact that there are relevant sources there can only be discerned through examining the source or by attempting to assemble/compile programs.  I think we could have a document included in the distribution that maps the original locations to the new locations to maintain the history.  (Geez, maybe we could even have a script that creates all these personal directories and copies the files back to them for those who really wanted it that way!).

eswenson1:
I didn't answer your SYSMAC issue.  I think all those INSRTs should go in SYSEN3 and get renamed to avoid collisions.  So, for example, GREN; MACROS > could become SYSEN3; GREMAC >, or something like that.

larsbrinkhoff:
Ok, updated the pull request.  I dropped the .TEMP. commit.

larsbrinkhoff:
I didn't move any files.  The move from JSR; STNDRD MACROS to SYSENG; JSF MACROS was already present in AI/MC/PI.

PR #352, opened 2016-12-20
Create SYSEN3 and move some files there

larsbrinkhoff:
SYSEN3 contents become:  
DPDEV 19  
GREN MACROS  
NICNAM 31  
NICWHO 30  
TYPE8 2  
WHOJ 26


eswenson1:
Have you verified that ITS treats SYSEN3 as a system directory, making note on the console of writes to that dir? Just wanted to verify that ITS didn't have a hard coded list of directories that we needed to update (rather than code to treat any directory beginning with SYS as one)

larsbrinkhoff:
No, I haven't.  I didn't know ITS treated system directories differently.

eswenson1:
I suspect the ITS implementor "did it right" and just treated any directory that started with the three letters "SYS" as a system directory.  Any time a program writes to these directories, a system message (see SYSMSG) is written to the operator console and stored in the SYSMSG log.  The user performing the operation is identified (as the "culprit") to aid in tracking down unauthorized changes to the system (or aid in tracking down why something is found to be broken).

larsbrinkhoff:
I tested it now, and you're right.  I created files in various directories, and only those beginning with SYS triggered the console logging.  Including SYSEN3.

eswenson1:
Great!  I was hoping we didn't have to find a hardcoded list of directory names in the ITS sources and update that!

PR #199, opened 2016-12-20
11SIM: build the 11/20 version

larsbrinkhoff:
And then add a SYS; TS PDP45.

See #197.


larsbrinkhoff:
I previously had the /45 version built, but to the wrong directory and filename.  Now it mirrors PI:
- SYS; TS PDP45 is the 11/45
- SYS; TS PDP11 is the 11/20

PR #372, opened 2016-12-21
Added ACOUNT. This is a dummy account program that allows users people

### LINKED ISSUE #285: acount
This is a dummy ACOUNT program that allows people coming from "safe" sites to login without an account. Probably not relevant.

sys;ts acount
ar9: kmp; dummy acount

eswenson1:
coming from "safe" sites to login without an account.
Resolves #285.


larsbrinkhoff:
Can you add to the commit message, which archive the source comes from?

PR #373, opened 2016-12-21
Update list of ITS machines.

larsbrinkhoff:
@Pul-Ess @bictorv I'd be happy to add more information about SV and UP.


bictorv:
What info do you want? (See up.update.uu.se/up.html for hardware info?)

larsbrinkhoff:
I suppose mostly just the year it started running, but you already provided that.

PR #408, opened 2017-01-10
Y2K bugs

### LINKED ISSUE #26: Y2K problem
The year 2016 in file listings is printed as 116.  This applies to both :LISTF and :FIND.

### LINKED ISSUE #405: BINPRT Y2K bug
Prints year 2017 as `;7`.

larsbrinkhoff:
:LISTF calls a monitor routine.  ITS"UQL5A extracts the 7-bit year field and prints the contents.  Since years are stored offset from 1900, this looks like a 2-digit year for years less than 2000.  I added 1900 to print a four-digit year in all cases.

BINPRT calls a routine in DATIME.  DATIME"TMDT2 is supposed to print a two-digit year.  To handle values over 99, I do a modulus 100.

FIND does the date printing itself.  I changed from two-digit years to four by adding 1900 in DATPRT and REFPRT.

GMSGS - truncate year to two digits.

DIRED - print year as four digits.  This only applies to the current file prompt, not a directory listing.

DUMP - print year as four digits.

INIT - print year as four digits.  Remove hardcoding for 20th century.

Fixes #26.  Fixes #405.


larsbrinkhoff:
Oh, I forgot.  Should I bump the version numbers?  For ITS, I'd like to accumulate more changes before doing that.

eswenson1:
You should bump the version numbers. I'll wait until you do before merging.

larsbrinkhoff:
I'll take this opportunity to scan the system for more occurrences of `[330700,,`

larsbrinkhoff:
I am, but there may be more places to fix which I can add to this pull request.  In particular, if I'm already touching DATIME, I can look there.

eswenson1:
Hmm... I see changed versions of existing files (with the same version FNAME2). That suggests you didn't create a new file with an incremented FN2.

eswenson1:
FIND, DATIME, and ITS are all modifications to existing (versioned) files.  We should create new files with incremented FN2s.

larsbrinkhoff:
Yes, I will increment the versions.

larsbrinkhoff:
But I don't want to do a ton of small edits and end up with e.g. ITS 1872.

eswenson1:
Ok, I’ll wait to merge until I see this.

larsbrinkhoff:
Versions are now incremented.  Four more programs fixed.

eswenson1:
Travis build failed.

-- Eric

PR #415, opened 2017-01-17
Utility for printing binary files

larsbrinkhoff:
Read binary files, and output each word in many different formats:

- Twelve octal digits
- Two halfwords, like AOBJN pointers
- SIXBIT
- ASCII
- Squoze

Suggested name: OD, if that isn't too Unixy.

This isn't anything that can't be done in DDT, it would just be slightly more convenient.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -723,6 +723,14 @@ expect ":KILL"
 respond "*" ":midas inquir;inqupd bin_inquir;inqupd\r"
 expect ":KILL"
 
+# od
+respond "*" "complr\013"
+respond "_" "liblsp;_libdoc; od\r"
+respond "_" "\032"
+type ":kill\r"
+respond "*" ":lisp sysen3;od (dump)\r"
+expect ":KILL"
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
I wrote one in Maclisp.

Putting the FASL file in SYSBIN is probably wrong.  Where should it go?

eswenson1:
Well, for standard system packages the FASL file goes in the LISP; directory.  However, take a look at LIBLSP and LIBDOC, and read the -READ- -ME- there.  Author-maintained LISP code seems to have gone in LIBDOC.

larsbrinkhoff:
Hold on, I'll add half-word display too.

larsbrinkhoff:
Depends on #436.

eswenson1:
Still holding on, right?

larsbrinkhoff:
It should be done now.  I wanted to use FORMAT ~< ~>, so I had to bring in LISP; FORMAT BRACK.

PR #440, opened 2017-01-19
MAGDMP and MAGFRM - handle files on a tape

larsbrinkhoff:
SYSENG; MAGDMP 27 says it best:

;MAGDMP is a standalone program loader/dumper for magtape.  
;The format of binary programs is the same as the ITS "SBLK"  
;format, as popularized by Midas.  The tapes are normally  
;created under timesharing with the MAGFRM program.


larsbrinkhoff:
@rcornwell These are now to be included in the ITS build.

larsbrinkhoff:
I'm not yet sure how to build MAGDMP.  Here are the options:
- Paper tape?  If so, add a RIM10 header.
- KL10? No, we don't need that, at least not for now.
- TM10A or TM10B?
- 340 display?  No, not now.

larsbrinkhoff:
I think RIM10 format will be good, given that the primar target is a KA10 emulator.

rcornwell:
Ok. The KA10 sim supports readin mode on the TMA/TMB tape interface. On TOPS10 I use BOOTM.REL as the boot block. RIM mode would probably also work since that is basically what readin mode is. The first word has a address/word count and data is placed there. 

I believe booting via Paper tape is also supported, however I have not tried it. 

KA10 sim supports readin of rim10,  .SAV, and .EXE files. It is basically the same code as Bob's KS10 for simH.

larsbrinkhoff:
I'm trying to figure out how, exactly, to create a file which the emulator will accept.

Right now, I'm looking at how the LOAD command reads a RIM file.  Apparently, the load_rim function builds 36-bit words from 6 octets read from the file.  And the octets has to have the 0200 bit set.  Ok, I can do that.

aap:
That's paper time format. Paper tape works in ascii and binary mode. In ascii mode the lower 7 bits are transferred, in binary mode 6 octets which must have have b10 in the upper two bits are combined into one 36 bit word.

larsbrinkhoff:
Thanks, after converting to paper tape binary format, the emulator accepts that layer of the file.   Next, I had to hack the emulator to accept the peculiarities of the ITS RIM format.

Now, it loads the file to 777200, and in the end there's a JRST 777200.  Good.  But "go" does nothing.  I was hoping for a MAGDMP prompt.

rcornwell:
What does instruction history show? Also try go 777200.

What is the difference between the ITS RIM format and RIM10?

larsbrinkhoff:
Nope.  Location 7772000 is zero.

> What is the difference between the ITS RIM format and RIM10?

The checksum calculation is different; there is a additional ROT.  Also, the very first block doesn't have a checksum.  In my fork, there's a magdmp.rim you can examine.

larsbrinkhoff:
Ok, more fixes.

Now the CPU goes to 777200, which is a CONO APR,635550.  But single-stepping the CPU never proceeds from there.  It's stuck at that location.

rcornwell:
It looks like MAGFRM will create a readin tape (just not for MC), The tape looks like it has a boot loader in it. You should be able to just attach this tape and type "Boot mt0". The  boot loader will read in the first record into memory at the correct location, then transfer to it. If MAGFRM does not write the boot loader to a separate record you might have to fix this.

larsbrinkhoff:
Wow, I didn't dare hope it would have that capability.  I'll certainly try to use it for booting.

Next would be to try building SALV to make a disk structure, and DSKDMP.

Issue #447, opened 2017-01-21
Make build more portable


aap:
I tried to build ITS for simh on FreeBSD earlier, but ran into some problems.

itstar and tapeutils need patching of tapeio.c because MTSETBLK and MTSETDENSITY are not defined in sys/mtio.h under FreeBSD. This fix worked for me:
```
#include <sys/mtio.h>
#ifdef __FreeBSD__
#define MTSETBLK MTSETBSIZ
#define MTSETDENSITY MTSETDNSTY
#endif
```

FreeBSD doesn't ship GNU make under the name 'make' but 'gmake'. The current build needs GNU make but has 'make' hardcoded in the makefile.

I ran into trouble with GNU make on both FreeBSD and an old raspian (versions 4.2.1 and 4.0 resp.) where the environment variable ${PWD} was empty. Invoking the build with `` gmake PWD=`pwd` `` worked in both cases but I still wonder what the problem is.

Lastly (for now?) itstar seemed to have build broken files on my raspbian and a friend's FreeBSD. sources.tape and minsys.tape have different checksums from the working ones (or is that expected?). The install fails like this:
```
...
DB     DUMP   SYSENG DIRDEV 86     DELRNM 10:10:51
DB     DUMP   SYSENG _DUMP_ OUTPUT WRITE  10:10:52
DB     DUMP   SYSENG SPELL  741    DELRNM 10:10:53
DB     DUMP   SYSENG _DUMP_ OUTPUT WRITE  10:10:53
DB     DUMP   SYSENG $CALL  MACRO  DELRNM 10:10:53
DB     DUMP   SYSENG _DUMP_ OUTPUT WRITE  10:10:54
DB     DUMP   SYSENG FORMAT 305    DELRNM 10:10:54
DB     DUMP   SYSENG _DUMP_ OUTPUT WRITE  10:10:54
ERROR:  - NON-EXISTENT DIRECTORY
20627>>.OPEN 7,4076   DB     DUMP   SYSENG LFN    3      DELRNM 10:10:55
```
You can get my broken source.tape here: http://aap.papnet.eu/share/broken_source.tape

larsbrinkhoff:
Thank you.  For the tapeio.c issue, can you please submit pull requests to the PDP-10/itstar and brouhaha/tapeutils repositories?

larsbrinkhoff:
Does FreeBSD have $PWD defined in the shell?

aap:
[...]

PR #448, opened 2017-01-21
Update the minsys DUMP from 442 to 444.

larsbrinkhoff:
The new version fixes the bug restoring a link as the first think in a directory.


larsbrinkhoff:
Problem reported in #447.

PR #452, opened 2017-01-22
KA ITS pager notes from Dave Conroy.

larsbrinkhoff:
These are notes in a very raw form, but we'll take what we can get.


larsbrinkhoff:
@rcornwell @dabridgham This may be helpful to you.

dabridgham:
Yes, indeed.  Thanks for the mention.

PR #443, opened 2017-01-24
Export GitHub issues

larsbrinkhoff:
There is a lot of valuable information in the issues and their comments.  We shouldn't rely on GitHub to store this information for us.

The GitHub API provides a way to extract the issues, pull requests and comments:  
https://api.github.com/repos/PDP-10/its/issues?state=all  
https://api.github.com/repos/PDP-10/its/pulls?state=all  
https://api.github.com/repos/PDP-10/its/issues/160/comments


larsbrinkhoff:
For now, we can manually run this script once a month or so.

aap:
I'm wondering whether we should use the wiki for some things instead? Wikis are git repos.

larsbrinkhoff:
Yes, everything that is labelled "note" could very well be a wiki page instead.  It's just that it's so convenient to create issues.

PR #418, opened 2017-01-27
STINKR - The new ITS linker

larsbrinkhoff:
SYS2; TS STINKR  
SYSENG; MULSEG INSERT  
.INFO.; STINKR INFO

Just a PDUMP binary.  No source that I can find.  Not even mentioned in the old backups?!?

There's a sample link script in C; CLIB STINKR.


larsbrinkhoff:
I believe the source code has been found.  I hope it can be released into the wild.

PR #128, opened 2017-02-02
REDRCT - redirect IP routing

larsbrinkhoff:
.; TS REDRCT  
SYSNET; REDRCT 21

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ A list of [known ITS machines](doc/machines.md).
    - PWORD, replacement for sys;atsign hactrn that requires registered logins.
    - QUOTE, prints out a random quote.
    - REATTA, reattaches disowned jobs to terminal.
+   - REDRCT, recirect IP routing.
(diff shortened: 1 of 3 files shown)


eswenson1:
What does this do?

larsbrinkhoff:
I don't know!  I haven't analysed it.

larsbrinkhoff:
Here's what it says when run without JCL:

```
Usage is:

  :REDRCT <old gateway> <optional new gateway>

Gateways can be given as hostnames, or in decimal octet form (as in
"10.0.0.77").  If the new gateway is omitted, REDRCT will simply pick a
likely-looking gateway from ITS's list of main gateways.

REDRCT will ask for confirmation once after it interprets its command, to be
sure it properly understood what you asked it to do.  Then for each routing
table entry it finds that uses the old gateway, it will ask for confirmation
before clobbering it with the new gateway.
```

Looks like it maps in system pages and rewrites tables in core.

PR #104, opened 2017-02-02
SALV - old directory salvager

larsbrinkhoff:
SYS1; TS SALV
SYSTEM; SALV 286 (missing), but SYSTEM; SALV 331 exists


larsbrinkhoff:
For the standalone version, I picked the ML machine.  AI has a disk controller which isn't yet supported by the KA emulator.  MC shouldn't work, because it's a KL.  DM may work.

larsbrinkhoff:
Good question!  It was because I copied the file into the repository, but it was already there.

Fixed and rebased.

PR #478, opened 2017-02-03
.MAIL.; NAMES cleanup

larsbrinkhoff:
I think .MAIL.; NAMES > should be cleaned up.  Also, .MAIL.; NAMES INFO should be added.


larsbrinkhoff:
Also added TURIST POLICY, because it was referenced in NAMES.

PR #406, opened 2017-02-05
CVTUNV - convert MACRO universal file to MIDAS file

larsbrinkhoff:
MIDAS; CVTUNV 4

No installed binary?


larsbrinkhoff:
I suggest we just include the file as-is in the MIDAS directory.

Also added some MIDAS documentation files.

PR #299, opened 2017-02-05
DATE - print date and time

larsbrinkhoff:
SYS1; TS DATE

Sources are missing for this, however, SV has:

HACK; DATE 2

which, although when assembled creates a very different executable appears to work correctly and improves upon the SYS1; TS DATE binary in the PI distribution in that it can handle years post 1999 correctly. Not sure if we want to use this source, but we should consider it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ A list of [known ITS machines](doc/machines.md).
    - CROCK, analog watch.
    - CROSS, cross assembler for micros.
    - CTIMES, Chaosnet time server.
+   - DATE, print date and time.
    - DATSRV, server for RFC 867 Daytime protocol.
    - DCROCK, digital watch.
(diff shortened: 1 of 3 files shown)


eswenson1:
Sounds good.

Pul-Ess:
On Sun, 5 Feb 2017, Lars Brinkhoff wrote:

>
>I disassembled AI: SYS1; TS DATE and SV: SYS1; TS DATE. The symbol tables
>are almost identical, so I would think @Pul-Ess's HACK; DATE 2 would be
>quite close to the original source.

HACK;DATE is disassembled from TS DATE, with a small y2k fix, IIRC.

 	/Paul

larsbrinkhoff:
Great, thanks!

PR #232, opened 2017-02-05
Expect script doesn't define "abort"

larsbrinkhoff:
When there's a timeout, build.tcl calls abort, but there's no such procedure defined.


eswenson1:
Yay!

PR #271, opened 2017-02-06
Fix NSALV unit number conversion bug

larsbrinkhoff:
John Wilson's [build.info](https://github.com/PDP-10/its/blob/master/doc/build.info#L35-L42) says there's a bug in NSALV.  Our source doesn't have the fix, but we may well be running his fixed binary.  We should check this, and if so, fix our source code too.

There are two places to check:
- bin/boot/salv.rp06, and
- build/klh10/@.nsalv-260-u


larsbrinkhoff:
It would certainly be good to check this with Alan or Ken, but they don't seem to have much spare time.

I have also tried to contact John Wilson, but no success.

PR #487, opened 2017-02-07
Y2K fixes for PEEK and DIRDEV.

larsbrinkhoff:
From @Pul-Ess  and @bictorv.


bictorv:
The comment in my version of the patch (see http://up.update.uu.se/PATCH%3bDIRDEV%20COMPAR) says it allows for a four-digit year.

eswenson1:
Oh, I see.  Thanks.

PR #484, opened 2017-02-10
Unix MLDEV/MLSLV

larsbrinkhoff:
I have started an MLDEV for Linux.  For now, it sends a NOOP to ITS, and gets the expected NOOP response back.


larsbrinkhoff:
Added support for writing files.  Still only text files, though.

Issue #402, opened 2017-01-08
Not building successfully using KLH10


b4:
:PDUMP DSK:.MAIL.;COMSAT LAUNCH
*:kill
*:job comsat
!
*:load .mail.;comsat launch
*debug/'NEW$:   0   -1
mfinit$g
.VAL 0; DEATHV+2>>JRST DEATHV+1
/
:kill
invalid command name "abort"
    while executing
"abort"
    invoked from within
"expect -exact $w"
    (procedure "respond" line 2)
    invoked from within
"respond "*" ":kill\r""
    (procedure "initialize_comsat" line 6)
    invoked from within
"initialize_comsat"
    (file "../build.tcl" line 606)
    invoked from within
"source ../build.tcl"
    (file "build/klh10/build.tcl" line 54)

SIMH does not bomb here

larsbrinkhoff:
The build went fine in Travis CI:  
https://travis-ci.org/PDP-10/its/jobs/190050150#L3923

Running COMSAT RFINIT needs networking, so I'm guessing there is a problem in that area.  I don't know if it's a problem with KLH10 configuration, or your host system.

SIMH skips this part, because we don't support networking with that emulator.

larsbrinkhoff:
Both Eric and I use Linux to run the emulators.  I don't know how well KLH10 supports FreeBSD.

eswenson1:
This does not bomb for me under KLH10 under Ubuntu Linux.  It seems COMSAT isn't initializing properly in your environment.

b4:
I'm running on ubuntu here - i'll investigate networking and see what's going on.

b4:
Trying 10.12.3.10...
Connected to 10.12.3.10.
Escape character is '^]'.
DB-ITS

DB ITS.1647. DDT.1546.
TTY 11
2. Lusers, Fair Share = 98%

ITS is up - but still bombing on COMSAT init

.VAL 0; DEATHV+2>>JRST DEATHV+1
 LOGIN  11TLNT 0 005001 13:16:41
 LOGOUT ___006 11 13:17:02
 LOGIN  11TLNT 0 005001 13:17:02
invalid command name "abort"
    while executing
"abort"
    invoked from within

eswenson1:
Did you make sure to update all the places where the IP address must be set correctly?  It is best to try this where the ip address of the machine is 192.168.1.100 (or whatever I have the IP address set to). Otherwise, you have to patch/update a bunch of places.

[...]

PR #461, opened 2017-02-13
Make IP number configurable

larsbrinkhoff:
Make it easier to specify the target IP number when building.


larsbrinkhoff:
I removed the binary host table from SYSBIN; it's now built from scratch.

@b4 Can you see if this solves #402?

eswenson1:
Looks like this failed. Fix?

larsbrinkhoff:
Yes, apparently some problem with the SIMH build.  I'll look into it and request a review when it's done.

larsbrinkhoff:
Fixed.

eswenson1:
I think the solution of building the binary host table the way you're doing it is fine. I suspect the way I was building it the right way of doing it once the system is up and running and the host table is in use.  But for bootstrapping the system from zero, I think the way you're doing it is better.

b4:
With a but of delay - boot drive in my VM box is failing and will need replaced.

larsbrinkhoff:
@b4 Whenever you are ready, of course.  You should just have to edit the top of the Makefile to change the networking configuration.

b4:
*comsat$j!
*$l .mail.;comsat bin
*bughst/'NEW$:   .IOT U4,3131   1203001144
domgat/   0   1203001005
tcpgat/   0   1203001005
debug/   -1   0
xvers/   -1   0
purify$g
:PDUMP DSK:.MAIL.;COMSAT LAUNCH
*:kill
*:job comsat
!
*:load .mail.;comsat launch

it did not crash!

However...what format is the IP address?  That doesn't seem to be octal as that's most definitely not 10.12.2.100 ;)

larsbrinkhoff:
Great!  Yes, 1203001144 is octal for 10.12.2.100.

Quick check in Emacs: (+ (ash 10 24) (ash 12 16) (ash 2 8) 100) gives the answer 168559204 (#o1203001144, #xa0c0265)

PR #529, opened 2017-02-16
NUDIR - create a user directory

larsbrinkhoff:
This is my own invention.  Makes it easier to create directories without having to remember the phrase ..NEW. (UDIR).

It also leaves a -READ- -THIS- file in the directory, to ensure it won't get deleted by the salvager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ A list of [known ITS machines](doc/machines.md).
    - NAME, Shows logged in users and locations, aka FINGER.
    - NETIME, network time dragon.
    - NICNAM/NICWHO, look up someone in the ARPAnet directory.
+   - NUDIR, create user directory.
    - NWATCH, small watch display.
    - OBS, observe system activities.
    - OS, realtime TTY spy.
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
I'm changing the name to NUDIR.  It's supposed to be a double entandre.

larsbrinkhoff:
Before the ..NEW. (UDIR) feature existed, I believe directories had to be created by patching ITS at location NUFDTB.  It's a table of directory names that are created automatically when references.  The first item is 0, which is why I think it was for patching.

So now I'm considering renaming NUDIR to NUFD in homage to ancient ITS usage.

eswenson1:
I assume NUFDTB existed in old ITS, but not in current ITS, right?  I don't see it in any of the files in the SYSTEM; directory.

There is an AUTOCR table in the SYSTEM; DISK >, but there are no zero-entries.

larsbrinkhoff:
Right.  NUFDTB was replaced by AUTOCR.  I'm guessing ..NEW. (UDIR) was introduced at the same time.

larsbrinkhoff:
The latest version we have with NUFBTB is SYSBIN; ITS CPY from May 1974, which is ML ITS 864.  The earliest sighting of AUTOCR is in SYSTEM; DISK 1012 from November 1976.

PR #521, opened 2017-02-17
Remove duplicated files

larsbrinkhoff:
KSHACK; KSDEFS 194 has a copy in  SYSTEM;

SYSEN2; MLDEV has two versions: 103 and 104.


larsbrinkhoff:
It could be that KSHACK was the offial home for KSDEFS, but I think KS ITS has moved from the hack stage and deserves its place in SYSTEM.

PR #532, opened 2017-02-17
Global ITS network

larsbrinkhoff:
Who wants to join?

- [x] @larsbrinkhoff's NO
- [x] @bictorv's UP
- [x] @eswenson1's ES
- [x] @b4's SJ
- [ ] @Pul-Ess's SV
- [ ] Lennox' CC (@blipvert?)
- [x] @livingcomputermuseum
- [ ] Other secret machines?


bictorv:
Again, not BV, but UP. :-)

larsbrinkhoff:
Sorry, UPdated.

larsbrinkhoff:
Note, this isn't ready for merging yet.  It's just available for review and commentgs.

eswenson1:
Sure, include mine.

larsbrinkhoff:
ES added.

b4:
Contemplating joining to be a way to test handling of very long domains (how many words is a 29-character domain entry? ;) )

larsbrinkhoff:
I see there was an ITS called CRETACEOUS-PARK, that's almost as bad.  Anyway, an ITS name should be shortened to two characters.  That ought to be enough for everyone.

b4:
Oh, the ITS name part is short - I claim "SJ" for San Jose.

The long domain part for the HOSTS file is just me wanting to look for anything that expect a short domain and barfs.

larsbrinkhoff:
That's fine with me.  Do you have any other information you want to be added, such as IP number, Chaosnet address?

b4:
Will sort the rest out in the morning - I can create repos from my phone, but not get a non-IPv6 IP from it!

b4:
What are the length limits on chaosnet addresses?

IP will be 50.131.218.138

larsbrinkhoff:
> What are the length limits on chaosnet addresses?

It's just 16 bits, divided into 8 bits for the subnet number, and 8 for the node.  We're using subnet 6, stolen from MIT.

b4:
Can I go with 7100 for a host address? (7 - California LATA prefix, 100 was the address for DEC-WRL per the pickle experiment paper)

larsbrinkhoff:
Sure, let's go with that!  I made a guess that sj.gewt.net would be the hostname.

larsbrinkhoff:
I think we are good to go.  Awaiting build status.

larsbrinkhoff:
Ooops, hold on, I should update MLDEV too.

larsbrinkhoff:
Oh, I already did that.  $P

larsbrinkhoff:
I now have 4.1BSD up and running most of the time...

MINITS works too, but for some reason it won't see any incoming packets when running on the same machine as the KS10 and VA

bictorv:
Cool! What kind of packet filtering do the VAX/MINITS emulators use? For example, you can't see BPF-filtered pkts to/from the same host.

bictorv:
Is PUMA routed over NO, by the way? Last time MX-11 saw PUMA (2 days 20 hrs ago) it was over NO, which seems... bad/strange.

larsbrinkhoff:
Re packet filtering, nothing in particular.  They run on SIMH, in which the CH11 device is implemented with the internal TMXR API.  Let's not go into details, but it provides UDP connections.

No, PUMA isn't (to my knowledge) routed over NO.  But maybe there's some strangeness which explains my problems.  They share the same host machine, but uses different UDP ports of course.  PUMA doesn't see incoming packets, so I don't have it running continuously.

Will ITS even do routing?

bictorv:
ITS understands routing, and NO should get routing info from MX-11, so if NO got a packet from PUMA to someone else it might send it to MX-11.

Will PUMA see incoming packets when NO/PREP are not running? (this goes beyond the current issue...)

larsbrinkhoff:
That explains why packets come from UP directly even though I'm just talking to router.aosnet.ch.

...hey that could explain why the packets are lost.  Possibly TMXR filters out anything that isn't coming from the expected peer, in this case router.aosnet.ch.

Here's what incoming UDP packets to 42044 looks like:

```
13:15:48.772426 IP UP.Update.UU.SE.42042 > Ubuntu-1604-xenial-64-minimal.42044: UDP, length 32
 0x0000:  4500 003c 0760 4000 3011 ff08 82ee 13d7
 0x0010:  ac1f 0164 a43a a43c 0028 71dd 0101 0000
 0x0020:  0100 1006 061f 0000 0663 842d 0001 0000
 0x0030:  5453 5441 5355 061f 0620 5620
```

And `netstat -nua` shows the open ports:

```
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State      
udp        0      0 0.0.0.0:42042           0.0.0.0:*                          
udp        0      0 0.0.0.0:68              0.0.0.0:*                          
udp6       0      0 2a01:4f8:c0c:104f:42043 2001:470:27:7e6:::42042 ESTABLISHED
udp6       0      0 2a01:4f8:c0c:104f:42044 2001:470:27:7e6:::42042 ESTABLISHED
```

42043 is vax780, working fine.

bictorv:
router.aosnet.ch (aka MX-11) is the cbridge process which runs on port 42042 on the host UP.Update.UU.SE; another process is the ITS system UP, which only connects to cbridge.

My guess would be that simh/TMXR does some filtering, and it might depend e.g. on who starts first (prep or puma)?

What do the routing tables on prep/puma look like?

larsbrinkhoff:
I seem to recall now that the recv() call in SIMH never indicates any incoming packet.  But maybe I should check that more carefully to be sure.

It seems to work fine if I run from another host.

I don't know what the routing tables look like or how to see them.  But MINITS does have a DUMP-ROUTING-TABLES contact.

PR #538, opened 2017-02-18
Make hostname and Chaosnet address configurable

larsbrinkhoff:
Adds hostname and Chaosnet address as Makefile variables that will configure the built ITS.

Update KLH10 to a version that supports @bictorv's CH11.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ SIMH=${PWD}/tools/simh/BIN/pdp10
 ITSTAR=${PWD}/tools/itstar/itstar
 WRITETAPE=${PWD}/tools/tapeutils/tapewrite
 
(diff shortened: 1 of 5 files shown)


larsbrinkhoff:
I'll add that too!

larsbrinkhoff:
No, it's not tested yet.  Actually, we can hold off merging this until we see that it works.

larsbrinkhoff:
I have been testing this during the weekend.  Setting hostname and Chaosnet address does work, and it doesn't break TCP/IP networking.  Enabling Chaosnet in KLH10 *seems* to work somewhat, but I haven't been able to connect to @bictorv's UP.

For the purpuse of this pull request, I consider the test good enough.

larsbrinkhoff:
UPDATE: CFTP is working!

PR #541, opened 2017-02-19
Run ICHECK after full dump

larsbrinkhoff:
This sets the "dumped" bit on all files.  Or whatever the bit is called that is output as "!" in directory listings.

ICHECK checks the tape number.  When it sees 0, it asks for confirmation.   Maybe the tape number should start at 1.  Or maybe 0 is good for this purpose.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -1389,6 +1389,9 @@ type ":dump\r"
 respond "_" "dump links full\r"
 respond "TAPE NO=" "0\r"
 expect "REEL"
+respond "_" "rewind\r"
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
This raises the question of how to number tapes.  With a physical machine and real magnetic tape reels, I suppose there would be numbered tape archive.  If someone is serious about maintaining their ITS, they should probably also number their tape images.

When a dump is made, the tape number is recorded in SYSENG; MACRO TAPES.  It's a binary file, and I don't know the format.  Or which other programs can read or write it.

PR #91, opened 2017-02-19
TTYSWP - Terminal Swapping Program

larsbrinkhoff:
Source missing for SYS1;TS TTYSWP


larsbrinkhoff:
I don't know how to use this.  The code works with numbers that may be the TTY number, or maybe a user index.

If you can make it do something useful, we can add some comments to document it.

PR #241, opened 2017-02-20
JOBDEV CHA

larsbrinkhoff:
DEVICE; JOBDEV CHA
DEVICE; JOBDEV CHAOS
MC:DCP; CHADEV 1 (missing source)


larsbrinkhoff:
I added the binary, and the link.

I'm not sure what its purpose is.  If you do `^r cha:`, you get a list of active Chaosnet connections, same as :peek K.

CFTP to DB doesn't work.  I was hoping this would fix it, but it didn't.  I suppose ATSIGN CHAOS or CHAOS FILE isn't quite on board yet.

PR #551, opened 2017-02-23
Clean up host table

larsbrinkhoff:
Remove local IP number for ES-ITS.  It is in conflict with the other IP for ES-ITS.

While at it, clean up the rest as well.  I hope we'll soon have DNS support.


larsbrinkhoff:
I removed it because I see it as a local configuration for your network.

We can keep it now if you want, but in that case, I'd like to later move it to another place which is not checked into the repository.  Maybe similar to how you can put local user directories in `user`.

larsbrinkhoff:
As of now, this is my local configuration for NO.nocrew.org:  
https://github.com/PDP-10/its/compare/lars/chaos

Do you need a gateway name configured as well?

eswenson1:
Well, without a gateway host configured, the it's won't be able to send any outbound mail or route ip traffic to any other host, right?

-- Eric

PR #535, opened 2017-02-17
SYSHST additions

larsbrinkhoff:
The HOSTS3 program had binaries in both SYSHST; TS HOSTS3 and SYSHST; HOSTS3 BIN.  The former makes it much easier to build the host table.

Also adds SYSHST documentation files.

PR #562, opened 2017-02-24
Reinstate SYSHST; HOST3 BIN, add SYSHST; TS H3MAKE.

larsbrinkhoff:
There is a problem with renaming SYSHST; HOSTS3 BIN in #535: M3MAKE relies on that name.

This reverts the rename, and adds a link from TS HOSTS3 instead.  Also adds a link for TS H3MAKE.


eswenson1:
What is that problem?

larsbrinkhoff:
M3MAKE needs the HOSTS3 program in SYSHST; HOSTS3 BIN.  It tries to open that file.

PR #556, opened 2017-02-24
SYS; ATSIGN CHAOS

larsbrinkhoff:
@CHAOS is currently assembled to SYSBIN and linked from DEVICE; ATSIGN CHAOS.  The link should be from SYS; ATSIGN CHAOS instead.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -536,7 +536,7 @@ expect ":KILL"
 # Chaosnet support
 respond "*" ":midas sysbin;_syseng;@chaos\r"
 expect ":KILL"
-respond "*" ":link device;atsign chaos,sysbin;@chaos bin\r"
+respond "*" ":link sys;atsign chaos,sysbin;@chaos bin\r"
 
 respond "*" ":link syseng;netwrk 999999,sysnet;netwrk >\r"
 


larsbrinkhoff:
This is probably one cause for Chaosnet services not working!

larsbrinkhoff:
@b4 You may want to install this change on SJ.

b4:
ITS or sleep...I choose ITS ;) Give me a few.

larsbrinkhoff:
It should only take a minute to move the file or create a link.  I don't think a reboot should be necessary, but I'm not sure.

b4:
I wanted to do a full clean and rebuilt as I needed to try and track down a bug.

b4:
Okay. Installed.

larsbrinkhoff:
Great, CFTP to SJ works now.

b4:
Wait...for CHAOS, should the port be forwarded to the ITS IP or the host IP?

b4:
What speeds?

larsbrinkhoff:
I have UDP port 42042 set up to be forwarded from the host to ITS.  That seems to work.  But now that you mention it, I'm not sure it's necessary.

Speed is abysmal: CFTP reports around "30 KBaud".

b4:
I will want to investigate what is causing it to be that bad.

It could be the internet, or I misconfigured I/O priorities.

larsbrinkhoff:
I also tried to SUPDUP into SJ.  I could login just fine.  Response to typing was really slow, but I could make myself a home directory.  But then the connection seemed to hang.  So yes, there's something to investiate.

PR #516, opened 2017-02-25
DIRCPY - copy a directory to another

larsbrinkhoff:
AR2: RWK; DIRCOP 1

No binary installed.  But I think this would be a good thing to have.


larsbrinkhoff:
@BobKerns This is your program I've been manhandling.  I added JCL parsing and made it handle directories on any devices.

PR #569, opened 2017-02-28
SN - snoop terminal output

larsbrinkhoff:
SYS2; TS SN

Source missing, but the binary isn't very big so it shouldn't be that hard to recreate from disassembly.


larsbrinkhoff:
I don't see anything promising named SN \<number>, so it looks like it will remain "source missing".

PR #574, opened 2017-03-02
Document the ITS character set.

larsbrinkhoff:
I'm not sure what exactly the ITS character set is.  This pull request adds a file to document that set, but I'd appreciate if people would review it and send in corrections.

### DIFF
--- a/doc/_info_/char.its
+++ b/doc/_info_/char.its
@@ -1,0 +1,36 @@
+ITS character set.  It's similar to the SAIL character set, but not
+identical.  These are the exceptions to standard ASCII.
+
+^@	centered dot
+^A	downarrow
+^B	alpha
+^C	beta
+^D	logical and
+^E	logical not
+^F	epsilon
+^G	pi
+^H	lambda
+^I	gamma
+^J	delta
+^K	uparrow
+^L	plus-minus
+^M	circle-plus
+^N	infinity
+^O	partial delta
+^P	subset of
+^Q	contains
+^R	intersect
+^S	union
+^T	for all
+^U	there exists
+^V	circle-X
+^W	equivalence (two way arrow)
+^X	back arrow
+^Y	forward arrow
+^Z	not equal (=/)
+^[	lozenge (diamond)
+^\	less than or equal 
+^]	greater than or equal
+^^	equivalence (three bars)
+^_	logical or
+^?	integral


larsbrinkhoff:
The first draft is identical to the SUPDUP character set documented in [RFC 734](https://github.com/PDP-10/supdup/blob/master/doc/rfc734.txt).

Things I'm particularly unsure about are:
- Do codes 136 and 137 retain their ASCII glyphs ^ and _, or should they display as arrows as in "old ASCII"?
- Do codes 175-176 copy from SAIL (ALTMODE and }) or ASCII (} and ~)?

dabridgham:
I suspect ITS and therefore SUPDUP originally used the SAIL character set and it was changed for the RFC but that's just speculation.

larsbrinkhoff:
I'm guessing the canonical ITS character set would have been the one used by the Knight TV.  But I don't know exactly what that looked like.

larsbrinkhoff:
I found the Knight TV font, and it's the the same character set as documented in this pull request.

MMcM:
The SAIL version of SAIL characters from MONCOM is https://archive.org/stream/monitorcommand00brucrich#page/116/mode/1up or http://www.saildart.org/allow/sail-charset-utf8.html.

As you can see, the differences are:

* The MIT SAIL is all graphic characters; control is a modifier. Stanford has the ASCII characters used by TOPS-10 in the regular places.

* Underscore and caret are in the standard ASCII places at MIT, but swapped with arrows at SAIL.

larsbrinkhoff:
Thank you, @MMcM.  The Knight TV font, SUPDUP RFC 734, and the CADR character set all agree.

I found a variation of the SAIL charset in RFC 698.  It looks like the usual SAIL, plus ITS characters added in various places.

The table below highlights the differences.

| OCTAL | SAIL | RFC 698 | ITS
| --- | --- | --- | ---
| 000 | | · | ·
| 011 | | γ | γ
| 012 | | δ | δ
| 013 | | ∫ | ↑
| 014 | | ±| ±
| 015 | | ⊕| ⊕
| 030 | _ | _ | ←
| 032 | ~ | ~ | ≠
| 033 | ≠ |≠ | ◊
| 136 | ↑ | ↑ | ^
| 137 | ← | ← | _
| 175 | ALT | ◊ | }
| 176 | } | } | ~
| 177 | | ^ | ∫

PR #577, opened 2017-03-03
Remove trailing garbage in PRUFD source.

larsbrinkhoff:
@bictorv saw that the file looked strange.  It's another case of a damaged file, but in this case it's just a benign addition of part of another file.


larsbrinkhoff:
Actually, is a duplication of a part of itself.  Looks like 512 words, maybe a disk block?

PR #579, opened 2017-03-08
Support idling in emulators

larsbrinkhoff:
Both SIMH and KLH10 supports saving host CPU cycles when the target operating system is idling.

- For SIMH, just throw a configuration toggle, and it detects the idle loop automatically.
- For KLH10, configure a special device and patch the operating system to use it.


bictorv:
I don't know about response times, but the idle time calculations go crazy with the idling instruction. It can likely be hacked around, if anyone cares enough.

PR #586, opened 2017-03-26
NO has moved to another IP number.

larsbrinkhoff:


larsbrinkhoff:
NAME is quite picky about the network configuration.  I have seen this failure before, but I don't remember what it was.

PR #589, opened 2017-09-13
Update to latest version of supdup

larsbrinkhoff:
Now with the option to use the extended ITS/SAIL character set.


lokedhs:
Looks good to me. Should the -u and -U flags be documented?

larsbrinkhoff:
Yes, they should.

PR #598, opened 2017-12-19
SALV doesn't agree with DUMP tape format

larsbrinkhoff:
When running `TRAN$G` from the old SALV with a tape made with DUMP, it says `BAD HEADER` for every file on the tape.

When running DSKDMP with a disk formatted with the old SALV, it says `MFDCLB`.  EDIT: Fixed, see below.


larsbrinkhoff:
SALV is the old KA/KL salvager.  It probably didn't receive updates when other software changed.  The KAs went away around 1984, and the KL in 1988.  I see the newer KS NSALV added the new format in 1989.

PR #605, opened 2018-01-04
Fixes for old Salvager

larsbrinkhoff:
Problems fixed when using the old Salvager with @rcornwell's KA10 simulator:

- Empty the input buffer after reading link info.  Failing to do so would mess up reading the following file.
- Set the link flag in the UFD when writing a link to disk.


larsbrinkhoff:
I actually copied the link setting code segment verbatim from NSALV, but sure, it's good to add comments.

larsbrinkhoff:
Most of SALV is rather short on comments, but now there's a little more.

eswenson1:
As an ongoing principle, I think we should always add comments as we change code.  Let’s not let the evils of our ancestors inculcate bad habits in ourselves.  :-)

PR #647, opened 2018-02-01
Minimal set of binary files for F40 FORTRAN IV compiler.

larsbrinkhoff:
By trial and error, I found these binary files to be needed to run the F40 compiler and load the resulting object code:

DECSYS; F40 SHR  
DECSYS; FORLIB REL  
DECSYS; JOBDAT REL  
DECSYS; FOROTX SHR  
DECSYS; LOADER SHR

I have tested this by compiling and running ADVENT.  See #167.


larsbrinkhoff:
I tried to use the JOBDAT REL in SAIL, but it didn't work with this (probably very old) version of LOADER.

larsbrinkhoff:
I'm labelling this source missing, even though sources do exist.  But I don't think it's appropriate to copy them into the ITS image.

eswenson1:
Why not include it in ITS build?

larsbrinkhoff:
Phil B mentioned something about runnning build scripts with GALAXY.  I thought it sounded like it wouldn't quite be worth the trouble.

PR #167, opened 2018-02-01
ADVENT (Colossal Cave) - adventure game

larsbrinkhoff:
SYS3; TS ADVENT

Also known as [Colossal Cave](https://en.wikipedia.org/wiki/Colossal_Cave_Adventure).

I don't think we have the source code in AI, MC, PI, or SV.  But there's Fortran source code here: http://www.ifarchive.org/if-archive/games/source/adv350-pdp10.tar.gz


eswenson1:
How would I find out whether any of my old ITS files is available in the MIT tape archive?

Would that be through Devon? I’ve asked him for some of the missing Macsyma sources but haven’t heard back from him yet.

larsbrinkhoff:
I think Brad Parker can get it done.  But the key is that people ask for
their personal files.

eswenson1:
I wrote to Brad about missing Macsyma files, and he directed me to Devon. I’ll try both.

eswenson1:
I wrote to Brad and Devon about recovering any files in MC: EJS;.

larsbrinkhoff:
Great, I hope something comes out of it!

larsbrinkhoff:
Interesting article about Adventure.

http://www.digitalhumanities.org/dhq/vol/001/2/000009/000009.html

PR #660, opened 2018-02-13
Recreated sources no longer needed

larsbrinkhoff:
@eswenson1 We recreated a few source files, but this is the only replacement as of now, right?


larsbrinkhoff:
STINK 201 in #45 is identical to our recreated version.  It was just a one-character change from version 200.

larsbrinkhoff:
SYSENG; DRAGON 63 isn't a replacement for our SYSENG; DRAGON 206.

larsbrinkhoff:
MRC; TEN50 22 is mostly cosmetic changes compared to our TEN50 1.  Not sure which to prefer.

eswenson1:
Yeah, the dragon source I recreated assembles to an identical binary as we got from PI ITS distribution. The most recent DRAGON source we recovered is very different and produces an entirely different binary.

larsbrinkhoff:
I added MRC; TEN50 22 too.  I think it's an improvement over TEN50 1.

PR #662, opened 2018-02-13
Standalone versions of TECO and STINK.

larsbrinkhoff:
Here are standalone TECO and STINK.  I don't know which versions.  The file timestamps are 1972-12-22 and 1973-06-12, so at least that old.  There are just binaries, no sources.

I have tested them lightly with the KA10 simulator.  They both do console I/O.  I got TECO to say "UNIT UNABLE", so it looks like it's trying to access DECtape.

TECO seems close to what's descibed in AIM-81.  STINK doesn't seem close to AIM-90.


larsbrinkhoff:
Minor point of contention: these files will end up on the MINSYS tape.  They have no use in boostrapping ITS, so an argument could be made they shouldn't be there.

larsbrinkhoff:
Also had to increase the timeout for the final dump.

larsbrinkhoff:
No, STINK isn't very useful on the MINSYS tape.  It's just that it's in the `.` directory, and all of that is included in MINSYS.

eswenson1:
Actually, now that you’ve included standalone Midas, I see the value of TECO and stink.

larsbrinkhoff:
Got TECO 335 which seems to be able to run standalone and work with paper tapes.

I'll see if it can be built and used to replace this binary.

Issue #449, opened 2017-01-22
MACDMP


larsbrinkhoff:
You are in a maze of twisty little programs, all alike:

SYSENG; MACDMP 10D256  
SYSENG; MACDMP 10U256  
SYSENG; MACDMP 6U256  
SYSENG; MACDMP 6U32  
SYSENG; MACDMP 6U64  
SYSENG; MACDMP MOBY  
SYSENG; MACDMP MOBY1  
SYSENG; MACDMP RIM10  

SYSEN1; 10DMP 13 - full moby  
SYSEN1; 10D16K 1 - 16K memory  

SYSEN2; TENDMP 2 - both microtape and TD10  
SYSENG; TENDMP 12 - same as version 2 except microtape is default  
SYSENG; TENDMP MAC - original DEC version  

SYSEN2; MACTAP F71 - library routines

Standalone tool for loading/dumping files on DECtape.

aap:
Have you come across any code for formatting DECtapes? Formatting is one thing in my emulation that I haven't implemented yet and I have some trouble imagining how a program would even do it, having some code would certainly help me.

larsbrinkhoff:
I have now made an extensive search for "dectape", and nothing obvious turned up.  Maybe it's in a MACDMP subroutine.

larsbrinkhoff:
By the way, do you mean some low-level formatting, or initializing a tape with a MACDMP format?

larsbrinkhoff:
10D16K is just 10DMP with one line changed.

TENDMP is 10DMP with many edits and additions.

aap:
I mean the low level formatting; writing the timing and marking tracks. I talked to Greenblatt and he said they didn't have anything of that sort...maybe programs supplied by DEC, but I'm not too familiar with DEC software. One other thing that I find strange is that both @rcornwell and Greenblatt said DECtape formats for different architectures were rather compatible. I can't imagine how that would work as the 36 bit format has an additional mark code (25, Block Space) that the 18 bit and 12 bit formats (which are otherwise identical) don't. And this code is quite important for reading too...not only used to extend the block number to 36 bits.

larsbrinkhoff:
Over all, I don't see a lot of references to DECtape at all in ITS material.

larsbrinkhoff:
@aap @rcornwell 
[...]

PR #663, opened 2018-02-13
Build standalone version of MIDAS.

larsbrinkhoff:
The latest version of MIDAS still retains the ability of standalone operation. 

Not tested at all.  I don't see much in the way of I/O instructions, so I'm guessing input and output goes from/to memory.  There's probably an AI memo for this.

Together with #662 and exec DDT, this forms a complete toolchain for standalone programming:

- Editing with TECO
- Assemling with MIDAS
- Linking with STINK
- Debugging with DDT

These can be run from DSKDMP, but they probably want access to a DECtape unit.  So MACDMP would be nice too, see #449.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -343,6 +343,11 @@ respond "*" ":midas sys1;ts salv_system;salv\r"
 respond "time-sharing?" "y\r"
 expect ":KILL"
 
+# standalone midas
+respond "*" ":midas /t .;@ midas_midas;midas\r"
+respond "with ^C" "TS==0\r\003"
+expect ":KILL"
+
 # magdmp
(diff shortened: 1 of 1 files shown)


eswenson1:
Now I see the value of standalone STINK from other pull request. I didn’t see standalone MIDAS when I commented on the other PR.

larsbrinkhoff:
Turns out setting TS==0 doesn't give us a standalone MIDAS.  It will still set ITSSW==1.

Setting ITSSW==0 elicits the message "So what monior is MIDAS supposed to run under?"

So either I'll have to revert this, or find an older version of MIDAS which can run standalone.  I found version 329 in Saildart which doesn't make any complaint when no target operating system is selected.  But I'm not sure that means it can run standalone.

PR #667, opened 2018-02-14
FAIL improvements

larsbrinkhoff:
Simplify use of STINK.  No need to double-ALTMODE every command; just the last.  No need to write out a FAIL BIN file, just exit to DDT.

Build SAIL; JOBDAT REL from source code.


larsbrinkhoff:
Apparently it *is* necessary to double-ALTMODE.  Or else the build script won't work.  Don't know why.

PR #666, opened 2018-02-14
Reconstruct OARCDV source code

larsbrinkhoff:
The source code for the old-format DEVICE; OARCDV BIN seems lost.  Rumor has it it was built from ARCDEV 66.  Version 68 is for the new format, and is very different.

We found ARCDEV 23, though.  It looks very similar to OARCDV.  It should be possible to disassemble OARCDV and fill the gaps in version 23 to bring it up to 66 equivalent.


larsbrinkhoff:
Oops, forgot to add to build script.  Coming up.

This new ARCDEV 66 builds to an identical binary, except the metadata parts.

eswenson1:
Very nice!  What are some examples of old-format archives that we have?

larsbrinkhoff:
These:  
AI: BAWDEN; AR69 SPCWAR  
AI: KSC; AR1 TIMER  
AI: KSC; AR2 HAKPRG  
AI: MOON; ARC.D'TRMF

larsbrinkhoff:
Note that ARCDEV 23 calls the signature word 777777777777 the new format.  So presumably there could be four formats:

- New: ARC!!1
- Old: ARC!!!
- Older: 777777777777
- Ancient: ?

larsbrinkhoff:
I better save this somewhere.  The old OARDV BIN has no information which source it was built from.  But there's this:

> Date: 29 February 1984 18:56-EST
> From: Kent M Pitman <KMP @ MIT-MC>
> To: BUG-ITS @ MIT-MC
> 
> I spoke with CStacy on the phone and he told me some bits I didn't have.
> 
> He saved the old arc device as DEVICE;OLD ARCDEV. It was built from
> AI: SYSENG; ARCDEV 66, which is earlier than the oldest source (68) on
> MC: SYSENG; ... the implication being that MC: SYSENG; ARCDEV 68,
> although it has been around for 5 years, has never been compiled and/or
> tested.

larsbrinkhoff:
I have now compared ARCDEV 66 against OARCDV BIN.  They are not the same!  So the information in the email from 1984 is not correct.

larsbrinkhoff:
SYSEN2; DMARCD 30 is in fact identical to my reconstructed ARCDEV!

larsbrinkhoff:
Hmm, maybe I was too hasty.  It could well be that the OLD ARCDEV in the email was assembled from ARCDEV 66.  But that's not the "old arc device" we need as OARCDV BIN.

larsbrinkhoff:
About the archive file formats.  DMARCD 30 indicates 777777777777 and ARC!!! is the same format.  ARCDEV 21 and 22 has an assembly time switch for ARC or DSK directory type, so DSK is the oldest format.  DEVICE; JOBDEV LIB was used to handle the DSK format.

- Oldest: DSK, first word random.
- Old: ARC!!! or 777777777777.
- New: ARC1!!.

PR #653, opened 2018-02-14
EMACS RMAIL

larsbrinkhoff:
The RMAIL program uses EMACS; RMAIL :EJ, which is a link to EMACS; [RMAI] >. [RMAI] is built from EMACS1; RMAILX > and EMACS1; RMAILZ >. But the latter is missing.


larsbrinkhoff:
There, source code found.

I could not figure out a way to make Generate Library write out to EMACS; [RMAI] >.

Issue #502, opened 2017-02-08
PT - print long-named file


larsbrinkhoff:
SYS3; TS PT  
AR3: CSTACY; PT 8

SYSEN2; PRINT 1 seems to be a slightly newer version, so maybe this is obsolete.

PR #623, opened 2018-02-15
PT - page test

larsbrinkhoff:
SYSENG; PT 2  

Optionally tests the quantum timer.  Maybe use it to check @rcornwell's KA10 simulator?

Not the same as PT in #502.


larsbrinkhoff:
I see there used to be a .; @ PT which probably is this program.  The binary has timestamps from as early as 1973, so that would be before the KL port.

@rcornwell have you been able to run this?  Does it need DDT to be loaded first?

rcornwell:
Yes I have gotten it to run, there is no need to run DDT. I am getting some errors in the checking but have not had time to trace them down.

PR #281, opened 2018-02-19
Spacewar - game

larsbrinkhoff:
Source code in BAWDEN; AR69 SPCWAR

https://en.wikipedia.org/wiki/Spacewar!


larsbrinkhoff:
The standalone version has been tested.  The timesharing version hasn't.

larsbrinkhoff:
@philbudne 

I'm trying to use the ASD ship (starting at upper left), but I have a hard time doing anything to avoid crashing into a star.  Are the thrusters very weak on this particular ship?

larsbrinkhoff:
Other than that, the KA10 build runs to completion and includes @ SPCWAR which works.  Good job everyone!

philbudne:
I'm not aware of any differences between ships, but I haven't tried
very hard to play the game (just tried to verify the controls work)
or looked deeply into the sources.

I find I have to lean pretty hard on the thrusters to avoid being pulled
into a star when the game starts... And even then I'm often whipped around
one star and thrown into another!

philbudne:
@larsbrinkhoff

Are you able to turn the ships?  With x11.c I find I have to click in
the window before keyboard events work

larsbrinkhoff:
I do click randomly in the window.  I'm not sure it does anything other than to put focus on the window.  I was guessing maybe a light pen input would start the game or something.

A and S works to turn my favourite ship (i.e. the key set I remember).

larsbrinkhoff:
We should have KLH teach us a lesson in PDP-10 gaming!

larsbrinkhoff:
I believe those are Spacewar consoles in the background.  Sussman obviously just won a round.

sussman-pdp6

larsbrinkhoff:
Hello @Tominator2,

I saw your web page about Spacewar controllers.  For your information, this GitHub project hosts PDP-6 Spacewar, PDP-11 Dazzle Dart and a few other programs using similar controllers.  Rather than building new devices, I'm adding support for USB game controllers to the simulators running the games.

Tominator2:
Hi Lars,

Thanks very much for letting me know and great work!  I haven't seen that
picture with Gerald Sussman before but the controllers look very much like
those in the Creative Computing article from 1981:
https://archive.org/stream/creativecomputing-1981-08/Creative_Computing_v07_n08_1981_August#page/n65/mode/1up

larsbrinkhoff:
I talked to Richard Greenblatt a few weeks ago.  He told me the very first Spacewar consoles for the PDP-1 was the best ever made.  All subsequent copies never felt as good; and they were either too large or too small.

larsbrinkhoff:
We found a file called WAR 44, which is an early version of PDP-6 Spacewar.  Best guess is around 1968.  It's appearance is much more plain than the elaborate 1976 version.  Source code and labels are reminiscent of the PDP-1 version.

larsbrinkhoff:
Information from SAIL people is that their version of PDP-6 Spacewar was ported around 1967-1968.  (This is close to the MIT version; was there a connection?)  It used a Type 30 display that was previously used with the PDP-1.

A later version from 1971 runs on both processors in parallel and draws on a III display.

larsbrinkhoff:
Text from "Galaxy" magazine, August 1963.  (This obviously can't be about the PDP-6 version.)

spacewar1963

larsbrinkhoff:
Another glimpse of the PDP-6 Spacewar console.

wcncls

From here: https://youtu.be/zOP1LNr70aU?t=429

larsbrinkhoff:
Michael Beeler writes:
> Here is some history on the Spacewar consoles. I wish I knew who made the original ones at the PDP-1.  Greenblatt is correct, they were great.  I heard that one of the officials at the PDP-1 came in one morning and found several MIT students who were not qualified PDP-1 users hanging around playing Spacewar.  So it was banned, and the consoles were hidden.  I don’t know what ever happened to them; perhaps they were thrown out.  Some time later, Spacewar was ported to the AI Lab PDP-6.  Clearly we needed some consoles.  So I designed as good a replica of the original ones as I could, and made them. Pine sides, masonite top, telephone switchboard toggle switches and pushbuttons from TMRC.  Slightly improved by a rubber strain relief sleeve.  As close to the dimensions and sloping top as I could remember.  Cinch-Jones plugs on the end of the cables, that plugged into sockets in the general purpose devices interface.

philbudne:
When I first "wired up" keystrokes to I/O bits in the SimH PDP-1 simulator
ISTR there were only four bits per controller turn left, turn right, fire
torpedo, fire engines, and hyperspace was signalled by both turn signals.
I assumed the controller had separate push buttons for turning.

When I updated SimH for the PDP-6/10 version of the game, there were
five bits per controller: one was added for hyperspace, which makes
sense for the drawings/photos of controllers with two double throw,
center off toggles.

If the original PDP-1 thruster switch were double pole (two
electrically independant double throw switches operating in tandem),
then wiring the turn lines to the separate poles on the unused throw
of the "thruster" switch would have been trivial.  And I guess, if it
wasn't a double pole switch, all that would have been needed would
have been a pair of diodes to prevent shorting the turn lines when
attaching them to the hyperspace pole....

I'm certainly curious to know if there's any first-hand recall of how
the controllers on ANY PDP-1 (CHM included) were/are constructed!

p

larsbrinkhoff:
The 60s version ("WAR 44") of PDP-6 Spacewar uses another I/O device
with 18 bits per console.

philbudne:
> The 60s version ("WAR 44") of PDP-6 Spacewar uses another I/O device
> with 18 bits per console.

a half-word per player?????  Here's what I implemented in the glue for
the 340 display on the PDP-6/10 (four nine bit bytes with four unused
bits):

/*
 * map 32-bit "spacewar_switches" value to what PDP-6/10 game expects
 * (four 9-bit bytes)
 */
/* bits inside the bytes */
#define CCW     0400                    /* counter clockwise (L) */
#define CW      0200                    /* clockwise (R) */
#define THRUST  0100
#define HYPER   040
#define FIRE    020

/* shift values for the players' bytes */
#define UR      0               /* upper right: enterprise "top plug" */
#define LR      9               /* lower right: klingon "second plug" */
#define LL      18              /* lower left: thin ship "third plug" */
#define UL      27              /* upper left: fat ship "bottom plug" */

ISTR the PDP-6/10 I/O device showed up with different bit-polarity
than the PDP-1 interface (one was active low, the other active high).

larsbrinkhoff:
Yes, a half word, though not all bits are used by Spacewar.  I don't even know what all the used bits do.

Device code 724 can be DATAI'd to get 36 bits.  Each half describes one console.  There are bits for:

- Rotate clockwise
- Rotate counter clockwise
- Faster (thrust)
- Slower (brake)
- "Button", Torpedoes
- "Xmit data", same as torpedoes
- "Pedal", hyperspace
- "General cancel", self destruct
- Time 0
- Time K
- Beacon beam

One of the inputs is called pedal, which is interesting.  I don't know what "time 0" or "time k" do.  The beacon is a beam that shoots from the front of the ship, but beyond looking very cool I don't know how it affects the game.  The labels are taken from WAR 44 source code.  It's in the ITS repository.

I have a simulation of this device on some branch.  It's using USB game controllers for inputs.

larsbrinkhoff:
Another interesting factoid is the the "nine-bit" consoles share the I/O device with the color scope and a "robot console".

Depending on some CONO bits, DATAI can come from different sources.  And DATAO goes to the color scope XY outputs.

PR #685, opened 2018-02-27
Build KA ITS

larsbrinkhoff:
This builds KA ITS on @rcornwell's KA10 simulator.

The configuration is based on the KA10 ML, but with many peripherals stripped.  Now there's just the console, two RP03 drives, and one TM10A drive.


eswenson1:
Nice!

PR #688, opened 2018-02-27
MAGDMP

larsbrinkhoff:
Fix MAGDMP not to echo control characters.  Echoing ESC doesn't look good.

In the KA build, run MAGFRMP to make a bootable MAGDMP tape.

Work around "APR ERROR IN NULL JOB" in build script.


larsbrinkhoff:
Oops, I closed this instead of merging it.

PR #696, opened 2018-03-02
Write files in . to DSK0:

larsbrinkhoff:
This way, there's no need to bring the second pack online in DSKDMP.


eswenson1:
And despite not understanding the DSKDMP commands, I'm guessing the removed line has to do with not having to reference another disk pack.

larsbrinkhoff:
Yes, this is related to the KA10.  We use two RP03 packs.

If DSKn device is specified ITS will place files randomly on the two packs.  But when DSKDMP starts, it only brings the first pack online automatically.  So you have to type L$1$ to load files from the other pack.

eswenson1:
I take it we use two RP03 packs instead of one RP06 packs (like we do on the KS) because of some limitation on the KA? Are 03s smaller than 06s, and one 03 is just two small?

larsbrinkhoff:
As far as I can see, KA ITS was never used with RP06 drives.

We based the configuration on ML, because because it's more conventional than both AI and DM.  And ML used RP03s.  During ITS bringup, @rcornwell discovered that two packs numbered 2 and 3 were needed.  I haven't checked why exactly.

RP03s are just 10 Mword, compared to RP06's 37.

PR #540, opened 2018-03-03
COMRED - Lisp library for TOPS-20 command line 

larsbrinkhoff:
Maclisp DUNNET uses a library to emulate the TOPS-20 command line.  This library used to be on OZ ISIS:

The API, as used by DUNNET, is quite simple:

- (comred-initialize)  
  Initialize library.
- (defspec symbol () help-text error-text)
  Define a command named by the symbol.
- (comspec-add-items symbol list t)
  Adds all subcommands in list to the command named by symbol.
- (comred symbol)
  Prompt, and read a subcommand for the command named by symbol.
- (comred-force-guideword text)
  Print guide text.  Do not break the line.
- (let-comred text . body)  
  Set prompt to text for the first call to comred inside the body.

There are two predefined commands which can be used with comred:
- confirm  
  Signals that the current command line is complete.  Advance cursor to next line.
- text-string
  Read a plain text string without any subcommands.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -762,6 +762,12 @@ type ":kill\r"
 respond "*" ":lisp libdoc;od (dump)\r"
 expect ":KILL"
 
+# comred
+respond "*" "complr\013"
+respond "_" "liblsp;_libdoc; comred\r"
+respond "_" "\032"
+type ":kill\r"
+
 respond "*" ":link inquir;lsrtns 1,syseng;lsrtns >\r"
 
 respond "*" ":midas inquir;ts lookup_inquir;lookup\r"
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
Some progress locating the original TOPS-20 COMRED library.  It seems to have been split in two files: COMRED.LSP with functions, and MACROS.LSP with ... macros!

I see these files:  
\<ZZZ.DARTS>MACROS.LSP.1 - source code for macros.  
\<PAO.R>MACROS.FASL.1 - compiled.  
\<PAO.R>COMRED.FASL.1 - compiled.
\<RAY.COURSE.ISIS>COMRED.FASL.1 - compiled.  
\<RAY.COURSE.ISIS>ISIS.TAGS.3 - lists all definitions in COMRED and MACROS.

So unfortunately no source code for the COMRED file, but maybe enough for a close reconstruction.

Issue #606, opened 2018-01-05
SALV and NSALV Y2K problem?


larsbrinkhoff:
I see some files with timestamps in the early 1900s, mostly in `.` and `SYS`.  It looks like the ones put there by SALV or NSALV.

larsbrinkhoff:
This could be the same as #486.

larsbrinkhoff:
This happens with KLH10, but not with KA10.  I'll check SIMH.

larsbrinkhoff:
It does happen only with KS10, but NSALV is fine.  The problem is that the bootstrap ITS binary has a Y2K bug when creating files.  (KA ITS uses a fixed version.)

larsbrinkhoff:
Fixed by #699.

PR #699, opened 2018-03-05
Update bootstrap ITS.

larsbrinkhoff:
The ITS used to bootstrap the build has a Y2K problem when creating files.  This leaves some bad file timestamps in the final image.


larsbrinkhoff:
This fixed #606.

larsbrinkhoff:
The new binary is built from src/system/its.1648 in this repository, using the DB RP06 configuration.  I just took it from the finished image.

Up until now we have been using ITS 1633 built by Alan Bawden in 1990.

eswenson1:
Ah, thanks.  I didn't realize that we hadn't already been using the one we had built from source.

PR #705, opened 2018-03-06
Resolves #704. Renames SYS; PURQIO 1 to SYS; PURQIO 2154

### LINKED ISSUE #704: SYS; PURQIO has incorrect version number
SYS; PURQIO >, as currently produced, is SYS; PURQIO 1.  This is actually version 2154.  The file should get renamed to SYS; PURQIO 2154. This will also ensure that the next version dumped will get version 2155.

eswenson1:
Causes newly created lisp to name dumped image as SYS; PURQIO 2154

rather than SYS; PURQIO 1.  The ITS convention is to have the PURQIO version number
match the version of the Lisp dumped.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -738,7 +738,7 @@ respond "*" ":job lisp\r"
 respond "*" ":load .temp.;*lisp bin\r"
 respond "*" "\033g"
 respond "*" "purify\033g"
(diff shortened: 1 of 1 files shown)


eswenson1:
Thanks. I've updated my local Git config to use my personal email address.

PR #144, opened 2018-03-26
RABBIT - Scheme compiler

larsbrinkhoff:
OSCHEM; TS RABBIT  
OSCHEM; RABBIT INFO  
ALAN; RABBIT 570  
SCHEME; HOWTO RABBIT


larsbrinkhoff:
I should probably use DUMPIT somehow to write out a TS RABBIT.

eswenson1:
I’ve been commenting out instances of (NEWIO T) in DECLARE special forms. The current lisp compiler doesn’t accept it any more. I suspect that this was required in Lisps that predated the introduction of NEWIO or at least when this feature was an option. Now it is builtin/standard. For some reason it is rejected now. In any case, just commenting it out allows the code to be compiled. The result runs fine, as far as I’ve been able to tell.

-- Eric

larsbrinkhoff:
I considered it, but I didn't want to add more workarounds at this point.

eswenson1:
But that means that building ITS won't actually get the scheme compiler built, and therefore claiming that the ticket is "fixed" is misleading!  :-)

larsbrinkhoff:
Sure, it's a known defect!

eswenson1:
Ok, then I'm sure you'll enter a ticket for it (or not close this ticket), right?

PR #765, opened 2018-03-26
Fix typo in emulator interrupt key.

atsampson:


larsbrinkhoff:
Thanks!

You can merge this yourself when the build is done.  In the spirit of the original ITS, everyone in the PDP-10 group can write to this repository.

PR #463, opened 2018-03-27
CHAOSNET ARPA/NCP/TCP - Chaosnet to Arpanet/Internet gateway

larsbrinkhoff:
SYSNET; ARPA 66  
DEVICE; CHAOS ARPA  
DEVICE; CHAOS NCP  
DEVICE; CHAOS TCP

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -502,6 +502,13 @@ respond "*" ":link device;jobdev chaos,device;jobdev cha\r"
 
 respond "*" ":link syseng;netwrk 999999,sysnet;netwrk >\r"
 
+# CHAOS ARPA/NCP/TCP gateway
+respond "*" ":midas sysnet;_arpa\r"
+expect ":KILL"
+respond "*" ":link device;chaos arpa, sysnet;arpa bin\r"
+respond "*" ":link device;chaos ncp, sysnet;arpa bin\r"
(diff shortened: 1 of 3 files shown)


eswenson1:
Why are these PRs saying they were opened about a year ago?

larsbrinkhoff:
Because I'm reusing the old issues rather than opening new pull requests.

eswenson1:
How do you do that?  I thought you had to create PRs for any new merges to master from changed non-master branches.

larsbrinkhoff:
There is an API to convert an issue into a pull request.  I have a shell script to do that in my git-scripts repository:  
https://github.com/larsbrinkhoff/git-scripts/blob/master/git-hub#L54-L81

eswenson1:
Why would you want to do this?  Seems to me that while the issue might be old, the PR (the request to merge) is new.  Why would you want the date on the PR to be that of the date the issue was created?  Is it so that we don't use up integer values too quickly (or issues and PRs)?  :-)

larsbrinkhoff:
I do that t keep the discussion about the issue and the pull request in one place.  So that information about the same thing is less spread out.

eswenson1:
Is that an approach you'd like the rest of us to use too?

larsbrinkhoff:
I'm not overly concerned.  I think people can do what they're comfortable with.

PR #780, opened 2018-03-31
Reconstruct source code for FACTOR.

### LINKED ISSUE #96: FACTOR - factors integers
SYS1; TS FACTOR
.INFO.;FACTOR ORDER

Source missing?

atsampson:
Based on a disassembly, so any misunderstandings in the comments are my own fault.

It turns out that this uses a straightforward trial division approach, although it has a couple of neat tricks to speed up its inner loop. To be fair, most of the better factoring algorithms we'd use today hadn't been invented in 1977.

I've put the reconstructed source in src/sysen1, but that's a fairly arbitrary choice; any better suggestions?

Fixes #96.


eswenson1:
I'd probably put it in sysen3, under the premise that this is not a very critical piece of software.  That's where we've been putting some of the random sources for programs where we weren't sure what the correct sysenX directory was.

atsampson:
Sounds good; moved to src/sysen3.

PR #783, opened 2018-04-03
Build ITS and some programs with 340 display support

larsbrinkhoff:
- Fix for 340 in KA10 simulator.
- Enable 340P in ITS.
- Set %TT340 bit for the console.  Some programs check this to decide whether to use the 340 or not.
- Assemble PEEK with 340 support.
- Prepare for assembling MAGDMP with 340 support, but don't throw the switch just yet.


eswenson1:
We hit the time limit again!  We've got to split up this build!

larsbrinkhoff:
I restarted the job, and now it went through.

eswenson1:
But that’s cheating! :-)

larsbrinkhoff:
I didn't add much, so I guessed the longer time was a fluke.

I rember the good old days when a full built took ten minutes...

PR #793, opened 2018-04-07
Reconstruct JOTTO

atsampson:
Based on the WAITS copy of the source. The build process is a bit hairy, but it does work; I've played a couple of games against it successfully. (It won.)

Questions for reviewers:

- Is everything in the right directories? (I don't know what the original source directory was so I've put JOTTO > and JOTTO DICT in GAMES. TS JOTTO was in SYS1 on AI/MC.)

- I've dumped the final executable using :PDUMP as with other similar programs, but that results in a 25% bigger binary than the original SBLK (large tables with runs of zeros in). Is it worth making it SBLK instead?


larsbrinkhoff:
Thanks!  I'll take a look at this later today.

Something to consider, though.  Sometimes it's possible to dig around and eventually find the original ITS file.  So if you're in a hurry to assemble a source file go ahead and reconstruct it.  If not, you may hold of a while and see it it pops up.

atsampson:
Yep, if the real source shows up later then all the better - it should be possible to drop it in directly, although I guess we'll need to check that the case-insensitivity changes have actually been made in the source. I'm interested in how early games like this were exchanged between institutions so I don't mind doing a bit of duplicate detective work.

We could simplify the build process by having BEG0 read a single filename string, and open it using the SOPEN call (and fix the missing TTY open at the same time). I'm happy to implement that but I'll give it a few days in case the original source turns up in the NGAME stuff EJS is looking at.

Issue #665, opened 2018-02-13
Muddle


larsbrinkhoff:
Making notes here.

larsbrinkhoff:
According to STINK files, these should be linked together:

- PURE
- SPECS
- CONST (not in MUDXXX STINK)
- LDGC
- UTILIT
- UUOH
- MUDEX
- MAPPUR
- CORE
- ATOMHK
- INTERR
- NFREE
- GCHACK
- READCH
- AGCMRK
- READER
- PRINT
- BUFMOD
- ARITH
- MAPS
- PRIMIT
- STBUIL
- EVAL
- DECL
- MAIN
- MUDSQU
- FOPEN
- PUTGET
- CREATE
- SAVE
- AGC
- AMSGC
- SECAGC
- INITM

larsbrinkhoff:
There's also MUDITS.

And SECAGC says it's for multi sections.

larsbrinkhoff:
UUOH:

```
UUOTBL+6        6        0.     1-039   .FATAL  Shadowing a pseudo-op
DSAVA1+1        566      0.     8-002   No END statement
        in IFE Starting at 7-362

**Comment by larsbrinkhoff** [comment]
MAPPUR:

```
FXUPGO+2        207      0.     4-682   No END statement
        in IFE Starting at 1-335

larsbrinkhoff:
ATOMHK:

```
RLOOK1+11       1050     0.     6-054   NOATMS  Undefined

**Comment by larsbrinkhoff** [comment]
INTERR:

Lots of symbols undefined.

**Comment by larsbrinkhoff** [comment]
READCH:

```
CIMAGE+14       1207     0.    12-002   No END statement
        in IFE Starting at 11-447

larsbrinkhoff:
EVAL:

larsbrinkhoff:
MAIN:

```
                GL+0     3.    12-004   Symbol table full
Unterminated successful bracketed conditionals
The first was at 12-004 of file MAIN

**Comment by larsbrinkhoff** [comment]
FOPEN:

```
CHNRNM+16       2207     0.    27-171   CHNRM1  Undefined
GTINTC+10       4456     1.    39-017   Stray )
PTINTC+12       4505     1.    39-048   Stray )

larsbrinkhoff:
SAVE:

```
FASTR1+5        176      0.     1-322   Externals multiplied

**Comment by larsbrinkhoff** [comment]
AMSGC:

```
[...]

PR #800, opened 2018-04-10
Zork source code

larsbrinkhoff:
I think the Zork source code is interesting even though there's no way to run it yet.  (See #665.)

Also, we have a version that is a slight update over the "latest version" published by Bob Supnik.

Does anyone think it would be a bad idea to just make a ZORK directory with the source files?


larsbrinkhoff:
Sorry, it appears I diffed the wrong way.  Supnik's version is from 1981, the other one from 1979.

larsbrinkhoff:
I renamed Supnik's 1981 files (which all has a .mud file extension) to match the ones from 1979.  I bumped the edited file version number by one which is probably historically incorrect, but better than having them unchanged.

larsbrinkhoff:
Hello @archivisth, 

This is the pull request where I added the two Zork versions.  The last commit will show the differences.

larsbrinkhoff:
A runnable January 1978 *binary* was installed in #1951

PR #784, opened 2018-04-10
Enable DECtape

larsbrinkhoff:
@rcornwell is fixing DECtape (TD10) support in his KA10 simulator.

When done, set NEWDTP==1 and NUNITS > 0 in KA ITS.

By default, ITS will use the MACDMP format for files on a tape.  There is also a way to read raw blocks.


larsbrinkhoff:
Now, in order to use a bootstrap ITS without channel 1 multiplexing, I had to modify the build scripts a little.  The bootstrap ITS uses a different simulator configuration from the one used to run the newly built ITS.

PR #801, opened 2018-04-10
Import the FACTOR 26 source

atsampson:
Also fix a bug in the KL mode it adds, and update FACTOR ORDER to match.

I've tested that it works with both KL10==0 and KL10==1. (Whether it's actually faster on a real KL10 is a question I can't easily answer!)


larsbrinkhoff:
The original binary was assembled for KL10==0, and this is now assembled with KL10==1, right?

atsampson:
Yes, in effect -- setting KL10==0 would produce the same inner loop code that AI's TS FACTOR binary had. (It just controls whether the code runs in the ACs or not; it doesn't use KL10-specific instructions or anything.)

atsampson:
Sounds reasonable - updated!

PR #698, opened 2018-04-16
KA ITS to use TM10B

larsbrinkhoff:
TM10A raises an interrupt for every word transferred.  And there's a limit to the interrupt rate ITS is prepared to handle.  TM10B uses a DMA thingy, so the data rate can be made higher.


larsbrinkhoff:
Since TM10B uses direct memory access, it has a potential for higher transfer speeds than TM10A which makes an interrupt for every work.  However, tt this point, the KA10 simulator doesn't have an option to increase the data rate.

rcornwell:
There will not be a increase in speed since the speed of the tape device was independent of the type of controller used.

Issue #827, opened 2018-04-22
STINK bug fixes for linking Muddle


larsbrinkhoff:
STINK 201 can't link Muddle.  The version bundled with Muddle works.

atsampson:
The version of STINK bundled with Muddle 106 is a TENEX port of STINK 121T. We had to fix some of the added conditionals to get it to build on ITS again.

The other change I applied was the following in `LOCGLO`:

```
-	JUMPGE D,[JRST 4,.]	;NO SYMBOL THERE
+	JUMPGE D,[PUSHJ P,RPB	;Expunge for nonexistant symbol - ignore
+	          JRST DATABK]
```

This was because the more recent MIDAS I used initially to build Muddle would emit a loader directive for EXPUNGE even when the symbol being expunged didn't exist. I don't think it's needed any more now we're using MIDAS 73 (and it could be done as a binary patch rather than a source modification easily enough).

Issue #826, opened 2018-04-22
MIDAS bug fixes for assembling Muddle


larsbrinkhoff:
@atsampson has found a number of problems in MIDAS when using it to assemble Muddle.

Check #665 for gory details.

Also, he's been using MIDAS 323 because 328 and later doesn't work.  So see what the difference is and fix it.  (77, 289, and 308 also work.)

larsbrinkhoff:
I think the next verison of MIDAS should be 459 rather than incremented for every bugfix or git commit.

larsbrinkhoff:
Also add the INFO; MIDAS ARCHIV file.

larsbrinkhoff:
MIDAS 323 can multiply a global symbol with a constant.  See e.g. MUDSYS; SAVE 176, `; POINT TO REAL CHANNELS SLOTS`.

MIDAS 458 won't do that.  It'll complain saying "Externals multiplied".

atsampson:
To summarise the changes we made to MIDAS 323 when we were using it for building Muddle:

- Break `.ALSO RET]` into two lines, so it would compile with MIDAS 458.
- Add `XMOVEI` to the symbol table, since the Muddle source used it (it doesn't any more) - 458 has this.
- Set the `IRCONT` flag for all two-word linker directives so they don't get split across a block boundary - 458 needs this fix.
- Define the right symbol name at the start of a `CONSTANTS` area - 458 needs this fix.

We're now using MIDAS 73 for Muddle, so none of these are strictly necessary, although it would be nice to include the last two in 459 for posterity.

It may well be that the other older code we're building with 323 would actually build with 73/77, so 323 itself might not be necessary.

larsbrinkhoff:
Unless 323 is necessary to build 77?

atsampson:
At present we build 73 using a binary of 77 - neither 73 nor 77 will build with 323.

Issue #12, opened 2016-11-21
Bring Zork back home


larsbrinkhoff:
Rich Alderson says he has the Muddle interpreter/compiler? for TOPS-20.  Binary only.  It could be disassembled and ported back to ITS.

![zork](https://user-images.githubusercontent.com/775050/34245278-7f9f83b4-e629-11e7-9bdd-6ea34a97c2e2.jpg)

larsbrinkhoff:
There may be another copy on twenex.org.

larsbrinkhoff:
@ethandicks Did you use a PDP-10 MDL interpreter when you ported Zork to z-code?

ethandicks:
On Mon, Nov 21, 2016 at 2:33 PM, Lars Brinkhoff
<notifications@github.com> wrote:
> @ethandicks Did you use a PDP-10 MDL interpreter when you ported Zork to z-code?

Not at first - I just read the MDL code, got the MDL books from MIT
Press and learned enough MDL to understand what the code was doing.
Fortunately, despite the massive visual differences between Inform and
MDL, because MDL is the ancestor of ZIL which compiles to Z-machine
byte code, and Inform compiles to Z-machine byte code, when one needs
to, say, generate a room with attributes and properties, the resulting
byte code is quite similar, so it's somewhat easy to "rephrase" a MDL
problem into Inform.

I did later do some side-by-side testing by running the original
36-bit Zork on one of the machines on the Internet with
open-to-request accounts.  I did not compile the MDL into a running
game, but I did run 20+ year old game files to elicit various
responses.

So the Inform port is a re-creation, but it's a section-by-section
re-creation with very similar logic and structure, but not 100%
identical.  It's using the Inform parser code, not the MDL parser
code, for example, but with various responses and some structure
backported from the MDL to Inform.

-ethan

larsbrinkhoff:
Zork MDL source code from Bob Supnik: http://simh.trailing-edge.com/games/zork-mdl.zip

Muddle manual: https://github.com/taradinoc/mdl-docs

ethandicks:
I did not get my MDL sources from Bob (I got them directly from one of the original contributors nearly 25 years ago) but they are identical.

eswenson1:
[...]

PR #830, opened 2018-04-23
MDL - MDL 56 interpreter

### LINKED ISSUE #665: Muddle

atsampson:
This is a tidied-up version of what @larsbrinkhoff and I have been doing in branches of the Muddle repository.

Fixes #665. Related to #12, #826, #827.


README.md
```diff
@@ -104,6 +104,7 @@ Some major applications:
 - Emacs, editor
 - Mac Hack VI and Tech II, chess programs
 - Maclisp, interpreter and compiler
+- Muddle, interpreter
```
larsbrinkhoff: I'm torn between calling it MDL or Muddle.  @36bit seems to prefer Muddle for presumably good reasons, but most manuals seem to say MDL.

doc/programs.md
```diff
@@ -102,6 +102,7 @@
 - MODEMS, modems gragon.
 - MSPLIT, split a file into smaller parts.
 - MTBOOT, make bootable tapes.
+- MUDDLE, MDL interpreter.
```
larsbrinkhoff: Since the executable is named TS MDL, I suppose this could be "MDL, Muddle interpreter."  Or whatever the proper name is.

eswenson1:
As far as I remember, the language was MDL, but the interpreter was called MUDDLE. TS MDL seems wrongdoings me. The executable was called TS MUDDLE.

larsbrinkhoff:
I'll note that the 1973 version includes a TS MUDDLE.  TOPS-20 executables are named MDL10*n*.EXE.  Adam, you found a reference to TS MDL in some manual?

src/mudsys/midas.324
```diff
@@ -13076,6 +13076,8 @@ SQUOZE 10,CLEARB
 CLEARB
 SQUOZE 10,ADJBP
 IBP
+SQUOZE 10,XMOVEI
+SETMI
```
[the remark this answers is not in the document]
atsampson: There are quite a few KL10 instructions in newer MIDAS's synonym table (search for XMOVEI). I could add them all here if we had a use for them?

README.md
atsampson: I was sticking to the terminology in the 55/105 manual: "The MDL programming language began existence in late 1970 (under the name Muddle) as a successor to Lisp ...".

I guess if we resurrect the 1973 version in the future, that would be Muddle and this is MDL.

atsampson:
TS MDL came from Chapter 1 of the 55/105 manual:

> Somehow get the interpreter running -- the program in the file SYS:TS MDL in the ITS version or SYS:MDL.SAV in the Tenex version or SYS:MDL.EXE in the Tops-20 version. [...] The program will be referred to as just "MDL" (or "the interpreter") from here on.

There's a comment in the source that suggest TS MUDDLE was a previous name, though:

> ; HACK FOR TS FOO linked to TS MUDDLE

We could install it under both names?

src/mudsys/midas.324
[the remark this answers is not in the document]
larsbrinkhoff: No need, at least not now.  I think it would be more useful to try getting MIDAS 458 in shape.

atsampson:
> Why remove the assignment to A rather than comment it out?

Good catch - I'd first rewritten it to compute a better limit for A, which didn't work, and forgot to put the original line back. I'll fix that.

larsbrinkhoff:
Maybe MUDSAV; TS MDL56, and then link SYS3; TS MDL and SYS3; TS MUDDLE to that?

atsampson:
Sounds sensible - that'd also allow us to have multiple versions installed, if we ever needed (say) a MDL54 with different behaviour for some application.

src/mudsys/save.177
```diff
@@ -319,7 +319,7 @@ IFN ITS,[
 ]
 
 ; NOW CYCLE THROUGH CHANNELS
-	MOVE	C,[-N.CHNS*2,,CHNL1]	; POINT TO REAL CHANNELS SLOTS
```
larsbrinkhoff: Does multiplying a global work with the MIDAS "323" + STINK "2" combo?  If so, maybe this can be dropped.

atsampson: Yes, it works:

```
002672:  000000000002  multiply next global reference by this                   
002673:  663013473135  reference minus global: 'N.CHNS', flags=15               
002674:  000000000000                           ;"      "                       
```

I'll drop that commit.

larsbrinkhoff: I see the code that checks this is very simiilar in 323 and 458.  See MULTP1.  Better check that the generated code actually works.  If so, maybe it would be a fix to copy that code from 323 to 459!

atsampson: I've added a testcase for this to [my collection](https://github.com/atsampson/its-tests/tree/master/midas). At some point I'd like to try all of those tests on the various versions of MIDAS/STINK we have...

src/mudsys/mud56.stink
```diff
@@ -0,0 +1,36 @@
+MPUREL
```
larsbrinkhoff: The default FN2 for STINK scripts passed on the command line is LOADER.  But since STINK "2" doesn't take a JCL, it doesn't matter much now.

PR #838, opened 2018-04-25
Fix high core purification.

### LINKED ISSUE #831: MDL can't purify high address space
@atsampson:

> MDL's PURIMP routine, which adjusts the memory map before dumping the final interpreter binary, fails on its final CORBLK call. We've stubbed this out for now, but this means that some pages which should be read-only are writable.

atsampson:
The PURIMP routine splits the page map into three parts: writable data
at the bottom, a large gap in the middle (which the GC will manage
later), and read-only code/data at the top. There's one CORBLK call for
each of these.

The final call, which gave pages BOT/2000 to 400 as the range, failed
with %EROPG when it hit the first unmapped page after the end of the
pure data. Experimentation shows that, at least with current ITS, this
is how CORBLK normally behaves.

Fix by calculating the final page from RHITOP rather than assuming 400.

Fixes #831.


atsampson:
If the pages are marked as pure, then they can be shared between multiple Muddle jobs (:CORPRT shows how many jobs are using each page, which is pretty neat).

Unfortunately it doesn't fix the GC errors I was seeing...

PR #109, opened 2018-04-25
LIMERI - displays random limeric

larsbrinkhoff:
SYS2; TS LIMERI
SYS3; TS . => SYS2; TS LIMERI
SYS2; TS LIMMER => SYS2; TS LIMERI
SYSEN1; LIMERI 1 (PI binary compiled from MC: SYSEN1; LIMERI 33, but that source not in PI).


larsbrinkhoff:
I assumed `:cwd eak` would do the trick, but it didn't.  .INSRT LIMS > takes the default from the source file, i.e. SYSEN1.

larsbrinkhoff:
@eswenson1 Reveiw ping.

PR #834, opened 2018-04-26
Add FAIL link script and documentation

larsbrinkhoff:
SAIL; FAIL -READ-  
SAIL; FAIL MANUAL  
SAIL; FAIL PUB  
SAIL; FAIL LOADER


larsbrinkhoff:
I think ? is there already, hidden on the third line.

larsbrinkhoff:
Note: the previous STINK script was my invention, which is why it looks strange.  The updated one is from an actual backed up file.

PR #614, opened 2018-05-03
SUDS

larsbrinkhoff:
There are SUDS files scattered here and there.  But not much of SUDS itself.

We'd like to have SUDS.  @ams inquired about it too.


build/sail.tcl
```diff
@@ -16,3 +16,102 @@ respond "*" ".jbsa/strt\r"
...
+respond "*" "\032:kill\r"
+
+# SUDS drawing program
+respond "*" ":cwd draw\r"
+respond "*" ":fail d\r"
+respond ";^C" ";\003"
+respond "*" "\032:kill\r"
+respond "*" ":stink d\r"
+respond "STINK." "\177"
+respond "??" "\033\0331L decsys; decbot bin\r"
```
atsampson: This stanza to dump out a file with the DECUUO bootstrap gets repeated several times - worth adding a helper function for it?

doc/programs.md
```diff
@@ -109,6 +111,7 @@
 - NAME, Shows logged in users and locations, aka FINGER.
 - NETIME, network time dragon.
 - NICNAM/NICWHO, look up someone in the ARPAnet directory.
+- NODIPS, SUDS...
```
atsampson: "SUDS wirelister (without DIP definitions)"?

build/sail.tcl
larsbrinkhoff: Maybe.  And in general there are many other places in the build scripts that could benefit from calling helper functions.  I'm not exactly in love with TCL, so I haven't been eager to work on the scripts.  On the other hand, no other great solution has presented itself so maybe we're stuck with expect.

larsbrinkhoff:
As part of building WL, NODIPS BIN is pdumped to DATDRW, so I'm assuming it could be a useful program in its own right.  But maybe it's not, I don't know.

larsbrinkhoff:
For those following this GitHub issue, I have some news.

We have found SUDS schematics for the Rubin 10-11 interface.  They can be loaded into SUDS, written to a plotter file, and then in several steps converted to a PNG files.  They are published with Tom Knight's blessings.  
https://stuff.offog.org/its/ten11/

This confirms that the basics of SUDS are working.

larsbrinkhoff:
I can add that we have not been successful converting Dick Helliwell's ADVENT DRW in the same way.  There are some error messages loading it, and the resulting plotter file is almost empty.

larsbrinkhoff:
340 display, found by @mattwyrm.  Is it SUDS in action?

https://www.youtube.com/watch?v=Hm3UY1o7Bdo&t=4m16s

larsbrinkhoff:
Chapter  SCALD II descibes D briefly:  
http://www.bitsavers.org/pdf/llnl/scald/SCALD_II_UsersMan.pdf

PR #858, opened 2018-05-05
XXFILE and TAA; UUOS

### LINKED ISSUE #13: XXFILE
XXFILE .insrts the file TAA; UUOS >, but there is no such file anywhere.

Disassembly reveals that TAA probably were at DM, so we're most likely out of luck.

### LINKED ISSUE #238: Reconstruct TAA; UUOS
Try to reconstruct TAA; MACROS well enough to assemble SYSEN1; XXFILE MTA224 and SYSEN2; MSEND 140.

atsampson:
@larsbrinkhoff and I had both looked at reconstructing @taa01776's TAA; UUOS to get XXFILE working; this merges both versions together. It can be used to build an TS XXFILE that's identical to the AI one (220), and a new TS XXFILE (224) built from source appears to work.

I've added XXFILE to the end of misc.tcl, but we may want it earlier to use it as part of the build process for other programs, if that's easier/faster than expect (see #160).

Fixes #13. Fixes #238. Helps with #119, but TAA; INP also needs reconstructing for that.


eswenson1:
Anyone know how to insert a ^Z (control-z) in a file from emacs?  I wanted to try out XXFILE, but I needed to send a ^Z to the process (e.g. LISP, or DECSYS; F40 and can't figure it out.  The obvious (^Q ^Z) doesn't work -- for obvious reasons.

larsbrinkhoff:
First ^Q, then ^_, finally ^Z.

eswenson1:
While we're at it, if you create a STY, and are providing input to the STY, and then, say, logout, how do you get back to the original hactrn?  I tried ^_ and various possible characters to no avail.  I wasn't sure where we have the doc on STYs in the file system (didn't see a STY > in .INFO. or INFO.

eswenson1:
Thanks. Now I just have to work out what the sequence is to type. I think it is ^\AZ but I have to do the math and try it out to know for sure. The esc char is ^\ followed by two chars. The second is Z to send a ^Z, and the first sets the control, shift, etc meta bits. I think @ means none, and A means control.

eswenson1:
I use reattach all the time. I have multiple active sessions with ES-ITS — because I forget that i’m Logged in somewhere — so I use reattach to clean up — consolidate. I only use it with my uname as a parm so as to reattach an old hactrn.

eswenson1:
I had tried ^_^Z and that didn't work.  I'll try the other three when I get a chance.

eswenson1:
Well, I tried all of these and the only one that worked was ^\^Z -- and only when logged out of the STY.  If I had a HACTRN (logged in or not logged in), ^\ evokes the JOB? prompt.  ^_^Z outputs ?? while you have a HACTRN.  ^_032 displays a ^Z and that displays ??.  ^_^B^Z displays ??.  So I don't know how to "get back" to your original session from the STY when the STY has a HACTRN (logged in or logged in).

PR #576, opened 2018-05-06
Fonts

larsbrinkhoff:
FONTS;  
FONTS1;  
FONTS2;  
.INFO.; KST FORMAT (maybe link to XFONT; KST FORMAT)

These are binary KST files, but can be converted to readable AST files with FED.


larsbrinkhoff:
Here are the files in FONTS, FONTS1, and FONTS2.

atsampson:
The content looks good. Do these directories need adding to the list in Makefile?

There's a description of the KST format in .INFO.; FED ORDER, with the note:

> CAVEAT: Several bugs in the description of KST format have been fixed, by EBM, 11/10/79.  I looked carefully at the source code for FED and XGPSPL, and tested things out to verify the fixes, but naturally cannot take complete responsibility since I did not write the programs.

larsbrinkhoff:
Thanks.  Added to Makefile.

larsbrinkhoff:
[AI WP 72](https://dspace.mit.edu/bitstream/handle/1721.1/41107/AI_WP_072.pdf): XGP Font Catalog.

larsbrinkhoff:
[AI WP 197](https://dspace.mit.edu/bitstream/handle/1721.1/41034/AI_WP_197.pdf): The Last Whole XGP Font Catalog.

Also has a description of the KST format.

larsbrinkhoff:
I suppose we should get this in now, since people actually are using fonts.

larsbrinkhoff:
I forgot to flesh out the commit message.  But maybe there wasn't much to add.

atsampson:
It occurred to me that the resolution of the XGP isn't far off that of a modern 4k monitor. So for anyone who'd like to try some of these fonts in their terminal emulator, here's [a quick hack](https://github.com/atsampson/its-tests/blob/master/tools/ksttobdf) to convert KST files to BDF files. 20FG KST is quite nice.

t

eswenson1:
If we have the capability now, to create a file from ITS that can be converted (somehow) and transferred to a linux host for printing on a modern-day printer, would someone please provide instructions for doing this?  I can't figure out how to do this from what I've read here and elsewhere.

atsampson:
Sure - here's an example.

- Create `HELLO TJ6` as follows:

```
.xgp
.font 0 40vshd
HELLO WORLD!
```

- `:tj6 /x hello`, which writes output for the XGP as `HELLO XGP`. (Or you could do something else that produces XGP output; I've also tried SUDS into SCNV, and @ with fonts specified.)

- `:versa hello xgp`, which converts it to a Harvard scan file `.TEMP.; HARSCN 1`.

- FTP `HARSCN 1` over to your host machine in image mode. (Or use a tape and itstar, in which case say -Wits below instead of -Wbin.)

- `harscntopbm -Wbin harscn.1 | pnmsplit - out.%d.pbm`, which will produce a PBM image file for each page. harscntopbm is in [this pull request](https://github.com/larsbrinkhoff/pdp10-its-disassembler/pull/19).

- The PBM images will probably need to be rotated (for TJ6 and @, by 90 degrees; for SUDS, by 180 degrees); you can do something like `for x in *.pbm; do pnmflip -r90 "$x" | ppm2tiff /dev/stdin "$x.tif"; done` to rotate and convert to TIFF. If you want a PDF file, you can then do `tiffcp -c g4 out.*.tif all.tif; tiff2pdf -p letter -x200 -y200 -ro -o all.pdf all.tif`.

eswenson1:
This process didn't work for me. I ended up with two TIFF files -- the first one, when viewed on my Ubuntu machine seemed empty.  The second one had the H of Hello and W of World with a space in between.  So the output was "H  W!"

Does the font only support uppercase letters?  My input was "Hello World!"

larsbrinkhoff:
@atsampson, this would be a perfect beginner HOWTO like doc/hello-midas.md.

atsampson:
Yes, that sounds like it worked correctly - that font does only have capital letters and some symbols in, and TJ6's output starts with a blank page.

A HOWTO sounds like a nice idea to me too. :-)

eswenson1:
@larsbrinkhoff I was totally wrong, above.  I had originally created two test tj6 files -- one with circumflex-f-1 and one with control-f-1, successfully created both XGP files, converted them to scan files with VERSA, and totally flubbed my FTP'ing to a linux host.  I overwrote one with the other.  In any case, I redid the experiment, and indeed, the font control is done with control-F.  circumflex-f just results in ^F1Foo^F0 being displayed in the resulting image. Sorry for the misinformation -- I guess my information regarding the need to use control-Q in emacs, however, was good, so only half disinformation!

larsbrinkhoff:
Thanks for checking, Eric.

Issue #874, opened 2018-05-09
make EMULATOR=sims fail


TheFausap:
Hello,
I'm trying to build the system with "make EMULATOR=sims", but after a while the build got a timeout with this error:

*:load sysbin;xxfile bin
*ttyop1$b   $g
$^K
*:PDUMP SYS2;TS XXFILE
DB     HACTRN SYS2   _DUMP_ OUTPUT WRITE  16:16:27
DB     HACTRN SYS2   TS     XXFILE DELRNM 16:16:27
*$P
$1B; TTYOP1>>SETZM TTYQ   kill
?U?

I'm using Ubuntu 18.04.

I did the same build under CentOS 6.9 and it worked.

regards,
Fausto

larsbrinkhoff:
XXFILE -> @atsampson.

Are you running Ubuntu and CentOS on different machines?

We have a long-standing bug where build stall for no reason.  But ?U? looks like a different problem.

atsampson:
Looks like the build script typed `:kill\r` after seeing the PDUMP line, but something ate the `:` before DDT got to see it. In my build log it says:

```
$1B; TTYOP1>>SETZM TTYQ   :kill
```

Maybe it should wait for `>>` - or we could invoke XXFILE in a way that'll let it run to completion - or add an explicit purification entrypoint to the code?

TheFausap:
Yes. I'm running Ubuntu and CentOS on different machines.

TheFausap:
An update: with EMULATOR=simh, under Ubuntu, I have no issue building the system.
So maybe this is a SIMS emulator issue.

larsbrinkhoff:
Thanks for the info.  Various combinations of computers and emulators can exhibit the problem.  We don't know what it is.  I have a vague sense that faster host computers are more likely to fail.  And klh10 is more prone to hang.

atsampson:
It builds OK for me with EMULATOR=sims. Looking at the XXFILE code, it hasn't opened the TTY yet by the time it hits the breakpoint, so I don't think this is a problem with XXFILE as such.

TheFausap:
Another update: I tested with another compiler version, gcc 5.5.0, but the result is always the same.

TheFausap:
Tried the new branch... I got always a timeout but in a different place

```
(ASUM SUPRV COMM DSKFN MLISP MEDIT DB DISPLA GRAM GRIND COMPAR NFORMA TRIGI CSI!
[...]

PR #875, opened 2018-05-09
Possible workaround for #874

atsampson:
At @larsbrinkhoff's suggestion - wait until we're definitely back in DDT before typing the :kill command. I don't know if this'll work around the problem, but it can't hurt!


larsbrinkhoff:
I believe @TheFausap confirmed this change was an improvement.

PR #817, opened 2018-05-10
Switch SIMH escape to ^\

larsbrinkhoff:
@rmaldersoniii recommends switching the SIMH console escape character to ^\

    d wru 034

Apart from being a great relief when working with EMACS, it will also make the escape character the same as the default in KLH10.

### DIFF
--- a/doc/DUMP-itstar.md
+++ b/doc/DUMP-itstar.md
@@ -1,31 +1,39 @@
 # DUMP and itstar
 
-### Saving ITS files and extract them using itstar
+### Saving ITS files and extracting them using itstar
 
-- Mount a tape image in the emulator.  Type Control-\\, and then for KLH10
+- Mount a tape image in the emulator.
 
-       devmount mta0 files.tape create
-       continue
+  - Type Control-\\.
 
-  For SIMH, type
+  - For KLH10, type
 
-       attach tu0 files.tape
-       continue
+         devmount mta0 files.tape create
+         continue
+
+  - For SIMH, type
+
+         attach tu0 files.tape
+         continue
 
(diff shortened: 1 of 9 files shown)


doc/DUMP-itstar.md
```diff
@@ -18,12 +17,12 @@
   wild card.  Finish with an empty line.  When DUMP has finished
   writing the files, type `quit` to exit.
 
-- Unmount the tape image.  For KLH10, type Control-\\, and
+- Unmount the tape image.  Type type Control-\\, and then for KLH10
```
atsampson: Type type.

doc/DUMP-itstar.md
```diff
@@ -2,13 +2,12 @@
 
 ### Saving ITS files and extract them using itstar
 
-- Mount a tape image in the emulator.  For KLH10, type Control-\\, and
-  then
+- Mount a tape image in the emulator.  Type Control-\\, and then for KLH10
```
atsampson: Make "Type Control-\" a separate bullet point? (Ditto below.)

doc/DUMP-itstar.md
```diff
@@ -2,13 +2,12 @@
 
 ### Saving ITS files and extract them using itstar
```
atsampson: Other tweaks to the file, while you're updating it:

- "and extracting" here
- Typo below: "directores"
- Is it worth mentioning doing "rewind" before "load"? (I usually need to do this when I'm reading in multiple files...)

larsbrinkhoff:
@eswenson1, I reworded DUMP-itstar.md to reflect that the character is now the same in both KLH10 and SIMH.  The escape character still needs be typed, same as before.

@atsampson, I'll make some fixes as per your suggestions.

Issue #862, opened 2018-05-07
Printing


larsbrinkhoff:
Add support for printing.  Over the years some of these devices came and went:

- XGP - Xerox Graphics Printer.  An early laser printer.
- Dover - A newer Xerox laser printer.  Attached to an Alto running Spruce.
- Calcomp plotter.
- Versatec 80.
- Gould electrostatic printer.  Attached to MC.  Later in Building 38.

larsbrinkhoff:
Of the three hardcopy devices, the XGP seems most promising.  Most typesetting applications support it, and the Dover is backwards compatible with SCAN files.

There are many options:

- Run the PDP-11 code in SYSEN2; XGP 31.
- Intercept files sent from SYSEN2; XGPSPL 197.  This is a PDP-10 program which sends files from the spooling directory to the XGP 11.
- Convert SCAN files to PDF without any PDP-10 or 11 involvement.

larsbrinkhoff:
Hardcopy file formats:

- Plain ASCII.
- SCAN.  XGP bitmap.
- PRESS.  Dover.
- PLT.  Written by SUDS.

larsbrinkhoff:
I see e.g. TJ6 writes output with text still in ASCII form, i.e. not SCAN files.  So font rendering happens in the PDP-11.  So the third option (converting SCAN to PDF) isn't useful in this case.

atsampson:
Yep, XGP sounds like a good bet. Going from XGP to PDF directly would probably produce the nicest output - some of the fonts used could be mapped to the standard PDF fonts, and others could be converted (which'd be a bit of work) or rendered as bitmaps.

A lower-tech approach would be to modify XD (#635) or VERSA (#856) to write their rasterised output as a bitmap file, then convert to PDF from there. It'd probably look OK but not be searchable.

It looks like SAIL used a similar format for their XGP - [this description](https://www.saildart.org/j5/J17_html/XGPSER.html#XGPSER:0023) is nearly the same as [INFO;XGP](https://github.com/PDP-10/its-vault/blob/master/files/info/xgp.24#L341).

larsbrinkhoff:
There's something to be said for pixel-perfect rendering.  Also, I'd like to see SUDS drawings.  Maybe XD would be the way to go for this.

But you have a good point about searchability.  I suppose (given enough effort) the fonts could be embedded into the PDFs.

atsampson:
Here's a SUDS drawing:

[...]

PR #877, opened 2018-05-10
Add VERSA

### LINKED ISSUE #856: Versatec Spooler
DCP; VERSA 210

Can read many file formats:
- ARDS
- Tektronix
- XGP scan
- Harvard scan
- Dover press

Can display bitmaps using SUPDUP graphics.

atsampson:
This is a rather flexible rasterising printer spooler, which works as a route for getting images out of applications like SUDS that print in PLOT or XGP format.

I've added a JCL parser so you can convert files (or show them with SUPDUP graphics, untested!) without needing to run the spooler demon. In spooler mode, it does seem to run OK, but I don't think we have the program for putting files into the spool directory; perhaps this was a modified version of XQUEUE, which uses the right spool directory but the wrong filename convention.

Fixes #856. Related to #862.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ SRC = system syseng sysen1 sysen2 sysen3 sysnet kshack dragon channa	\
       tensor transl wgd zz graphs lmlib pratt quux scheme gsb ejs mudsys \
       draw wl taa tj6
 DOC = info _info_ sysdoc sysnet syshst kshack _teco_ emacs emacs1 c kcc \
-      chprog sail draw wl pc tj6 share
+      chprog sail draw wl pc tj6 share _glpr_
 BIN = sys2 emacs _teco_ lisp liblsp alan inquir sail comlap c decsys moon \
       graphs draw datdrw fonts fonts1 fonts2
 
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -798,3 +798,9 @@ respond "*" "purify\033g"
 respond "DSK: SYS; TS NTJ6" "\r"
 respond "*" ":kill\r"
 respond "*" ":link sys; ts tj6, sys; ts ntj6\r"
+
+# Versatec spooler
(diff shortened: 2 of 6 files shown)


larsbrinkhoff:
I'm thinking the spooler would have been put in (or linked from) the DRAGON directory and doesn't need to be in SYS: or have a TS name.

larsbrinkhoff:
There was a CHANNA; V80SPL during the 1982-1985 period.  Also GLPT; TS SPOOLR in 1976-1978.

VERSA 210 has timestamp 1983-09-19, and version 4 which is the first I can see is dated 1981-06-22.

So I'm guessing the Gould printer was installed in the late 70s, and then maybe VERSA used some code from the Gould spooler.

larsbrinkhoff:
Also CHSGTV; TS SPOOLR and SPGLPT 19 both from 1978.

larsbrinkhoff:
With this information in mind, I think there could be an optional link from CHANNA; V80SPL to SYS3; TS VERSA.  Currently, we don't have a way to specify which system services should be enabled or not.  For now the link could be commented out.

larsbrinkhoff:
This from INFO; GLP 22 makes it sound very much like VERSA.

> The Gould spooler can currently print graphics files using ARDS, Tektronix, DPLT, and IMAGE formats.  ARDS files are most commonly produced by the MACSYMA PLOT2 package or Muddle graphics.  Tektronix files usually originate on some other machine and are copied to MC for printing.  DPLT files are produced by the Stanford Drawing System (SUDS) or GEOMED.  Image files are raw bit data.  In the future more types will be added (FR80, ITS, SCAN, etc).

atsampson:
Sounds good. I've removed TS SPOOLR and add a commented-out link to CHANNA; RAKASH V80SPL.

larsbrinkhoff:
Conflict in Makefile DOC, sorry!

PR #651, opened 2018-05-15
Lisp display slave

larsbrinkhoff:
SYS; ATSIGN 10SLAV  
SYS; ATSIGN 6SLAVE  
SYSEN2; LD10 61  
SYSENG; FDITS 43  
L; SLAVE 8 (in AR1)


larsbrinkhoff:
Some outstanding questions:

- Where to put the GT40 binary for LD10?
- What is FDITS?
- How does it differ fromthe GT40 LD10?
- Who uses it?
- Where should the binary be?

larsbrinkhoff:
"GT40 Utility Programs and the LISP Display Slave"  
https://github.com/larsbrinkhoff/its-archives/blob/master/ailab/AI_WP_059.pdf

larsbrinkhoff:
Ah, the working paper answers some questions: SYENG; FDITS > is .INSRTed when assembling the GT40 slave.

eswenson1:
@larsbrinkhoff  how would I identify lisp code that uses this stuff?  What would that code be doing that would identify itself?

larsbrinkhoff:
Basically anything that makes calls to `(DIS...`  I made the obvious checks already.  I just asked in case you'd know already.

larsbrinkhoff:
We don't have anything installed at the moment, but

- SHRDLU makes a call to DISINI
- LLOGO makes calls to DISINI, DISCOURSE, DISFLUSH, DISCREATE, etc...

PR #896, opened 2018-05-15
Macsyma build script problem

larsbrinkhoff:
I may have found a problem with the script used to build Macsyma.

When `(todo)` is sent, there will be a lot of output about not finding unfasl files.  When `(todoi)` is sent, the type procedure will ensure that characters are echoed.  But the individual characters will match the pending output from the compiler.  So the script thinks characters are echoed, even though they not yet have.

I'm not sure if this is actually a problem or not.  Maybe the characters are buffered and received later.


larsbrinkhoff:
I made a change so the characters aren't expected one by one, but rather in a complete string.

Also, it seems my version of expect doesn't handle `expect -timeout N {` well.  I changed it to an explicit assigment to `timeout` instead.

Issue #869, opened 2018-05-09
Update SUDS to latest Stanford version


larsbrinkhoff:
SUDS has now been verified to work with the TEN11 files.  But it's not happy about ADVENT DRW.

Investigate whether SUDS works better when updated to the latest Stanford version.

Issue #852, opened 2018-04-27
SUDS parser


larsbrinkhoff:
GLR; SUDS LSP58

From here: https://github.com/PDP-10/SUDS/blob/master/bits/suds.tar.gz

By Gerald Roylance.

larsbrinkhoff:
@lisper's C tools here:  
https://github.com/PDP-10/SUDS/tree/master/tools

larsbrinkhoff:
I tested SUDS LSP58 with the ADVENT DRW file.  It wasn't exactly a success.  After having added a definition for LOGCLR, I get this:

```
***SUDS-ORIENTATION***** 
-67777777 
; Loading FORMAT ([ML]DSK:FORMAT;FORMAT 827)
      -5000012       -6,,    -12        -1310730.  ..W.{ ggbgg^
 -354000000000  -354000,,      0    -31675383808.  E.... JH((((
             0        0,,      0               0.  ..... ((((((
             2        0,,      2               2.  ..... (((((*
      70000001       70,,      1        14680065.  ..@.. ((`(()
      -5000012       -6,,    -12        -1310730.  ..W.{ ggbgg^
 -264000000000  -264000,,      0    -24159191040.  S.... QH((((
             0        0,,      0               0.  ..... ((((((
             2        0,,      2               2.  ..... (((((*
        777711        0,,    -67          262089.  ....d (((gg1
      -5000012       -6,,    -12        -1310730.  ..W.{ ggbgg^
 -310000000000  -310000,,      0    -26843545600.  N.... O(((((
             0        0,,      0               0.  ..... ((((((
             2        0,,      2               2.  ..... (((((*
            71        0,,     71              57.  ..... (((((a
      -5000012       -6,,    -12        -1310730.  ..W.{ ggbgg^
;BKPT ORIENTATION

PR #905, opened 2018-05-19
Fix ADVENT DRW so the ITS version of SUDS can load it

atsampson:
Loading ADVENT DRW into D produced several error messages about empty sets and invalid font sizes, and didn't load a usable drawing. (See #852 and #869.)

SUDS files start with a version number. Comparing ADVENT DRW with the known-good files in TEN11 revealed that while they both claim to be version 25, ADVENT DRW has an extra zero word in its header. Stepping through D's loading process showed that this knocked the input file parser out of sync, so it was trying to parse the type definitions section as macros, etc., hence the errors.

D does expect to see this extra word -- actually an ASCIZ string -- in version 26 files (search for NOMVER in IN >), so marking this file as version 26 makes it load properly.

I guess the same problem is likely to come up in the future when loading SUDS files from SAIL or DEC. A quicker workaround is to set a breakpoint on RDNOM, load the file, and change RDVER from 25 to 26 when the breakpoint fires...


atsampson:
Here's what it looks like - you need to zoom out a bit (^F*) to see the whole thing.

harscn 4 0 pbm
harscn 6 0 pbm

larsbrinkhoff:
Great!

PR #907, opened 2018-05-20
Added NGAME and games invoked by it.

eswenson1:


src/games/lsrrtn.1
```diff
@@ -0,0 +1,140 @@
+title LLSRTN - LISP interface to -*-MIDAS-*- inquir lookups
```
larsbrinkhoff: There already is a INQUIR; LSRRTN 31.  Is this one different?

src/games/stone.extra
```diff
@@ -0,0 +1,105 @@
+
```
larsbrinkhoff: STONE EXTRA seems to be a personal resume rather than game documentation.

src/games/stone.next
```diff
@@ -0,0 +1,95 @@
+
```
larsbrinkhoff: Another version of Mr Cohen's resume.

eswenson1:
I wonder how I got those two stone.(extra next) in there?  I removed them. The version of LSRRTN is different in INQUIR and GAMES. I kept the one in GAMES local to GAMES for now. I'll enter an issue to compare the two and attempt to consolidate if appropriate.

larsbrinkhoff:
Thanks for fixing my LIMERI mistake!

PR #940, opened 2018-05-31
Get @ working with VERSA, and add a HOWTO

atsampson:
Several more VERSA improvements, to format @ output properly, and fix a couple of bugs in my JCL code. Import @'s manual, and add a document showing how to print from TJ6, @ and SUDS.

Rather than documenting how to create a PDF by hand, I've written a script and submitted [a pull request](https://github.com/larsbrinkhoff/pdp10-its-disassembler/pull/20) to the dis10 repo for it.

Do the instructions sound reasonable? Are there any other programs that I should include instructions for?


larsbrinkhoff:
Will backspace do overprinting?

larsbrinkhoff:
Assming you have seen the @ output, do you think the DAZDRT and CHESS 144 listings were done with @?

larsbrinkhoff:
I read the instructions, and I think they are reasonabe.  I can't think of any other programs off hand.

larsbrinkhoff:
Oh wait.  Please link to printing.md from somewhere.  Probably the Documentation section in the top-level README.md.

atsampson:
Backspace will indeed do overprinting, and you can backspace past the left margin. (I tested backspace and tab by hand-hacking an XGP file, since TJ6 doesn't use either and @ only uses tabs in the cross-reference list.)

The CHESS 144 and DAZDRT 110 listings are from a contemporary version of @ - the current version prints a key at the top of the cross-reference listing, but otherwise it looks pretty much the same if you give it the switches listed on the front page.

eswenson1:
I suspect so, that looks like @ output to me.  And that's what we all used to generate listings of MIDAS code.

PR #949, opened 2018-06-03
Fixes to the documentation from the GPL release

atsampson:
Three changes:

- Some files in SYSDOC are in formats where the license header should be a comment (as has been done for SYSTEM).

- We've picked up some more of these files since @36bit added the headers originally, so do the same for the new files. (The remaining files are mostly mail archives, which probably shouldn't be included in DB.)

- Our copy of .CALLS 127 was corrupt; fix it.


larsbrinkhoff:
@36bit,

Yes, that's what the guideline says.  But it's just a guide and we don't always follow it to the letter.

There's some amount of red tape involved in creating pull requests, so sometimes it makes sense to add a closely related fix.

PR #951, opened 2018-06-06
Added support for macsyma translate_file.  Also added ctensr fasl. 

eswenson1:
Now tsetup() should work.


eswenson1:
I appear to have no end of problems with sub modules. I had no intention of modifying (nor do I remember touching) that sub module.

larsbrinkhoff:
Ok, I'll fix it.

larsbrinkhoff:
In general, submodules doesn't work like regular files or directories in git.  But in this case, `git checkout origin/master -- tools/dasm` works fine even though the target directory is a submodule.

eswenson1:
I guess the obvious practice doesn't work.  Whenever I'm about to start on something new I do this:  "git checkout master", "git fetch", "git rebase origin/master", and then "git checkout -b ejs/foo".  I make my changes and then push my local branch.  Somewhere along there, my submodules sometimes get out-of-sync and are viewed as "different".  When I do my push, it seems like I've either changed the submodule (which I haven't) or have an older version of the submodule.

larsbrinkhoff:
Everything above looks ok.  I'll update the master branch with `git merge --ff-only origin/master`, but rebasing ought to work too.

I don't know exactly how or when submodules are included in a commit.  I think you'll have to `git add` them, or maybe if you touch the `.gitmodules` file.  I'll sometimes do an overreaching `git add` myself and get some files or directories included which I didn't intend.  I'll usually do a `git status -s` before making a commit to see that nothing unintended has snuck in.

eswenson1:
Doing the "git checkout origin/master -- tools/dasm" doesn't work for me.  After I do it, and when I do a "git status", it still shows me:

> Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)
>
>        modified:   tools/dasm

larsbrinkhoff:
Ok, then maybe you have to fix the staging area first.  With git reset as the help text suggests.

I'm sorry, it's hard to say exactly without access to the working area.

eswenson1:
I got past it with a git submodules sync and update.

PR #956, opened 2018-06-07
Use <kbd> to mark control keys in the examples.

### LINKED ISSUE #946: writing TJ6 file from EMACS
Hello,

I'm trying to write down the example in printing.md using emacs and supdup terminal.

All is fine, except for the font switching. According to the TJ6 manual, I should use CTRL-F (|F and font number), but this, in EMACS, move the cursor.

In the emacs.guide I find something related, I should use M-# with an argument (M- is ESC, right?), but if I type ESC then # (shift-3) I have no way to write the font number, the M-# (at bottom) disappear.

How can I do that ?

thanks,
Fausto

atsampson:
Fixes #946.

Is it worth doing this for the other .md files as well? I'm happy to do so, but it does make the markup more complicated (especially for something like the DDT guide).

### DIFF
--- a/doc/printing.md
+++ b/doc/printing.md
@@ -16,13 +16,13 @@ TJ6 is a typesetting program, in the same tradition as TJ-2, RUNOFF
 and roff. It justifies paragraphs of text, and produces output either
 in plain text or XGP format. Here is a simple example of an XGP file:
 
-```
+<pre>
 .xgp
 .font 0 times 12rom
 .font 1 times 12bold
 .twinch 4
 .center
-No. 1190. ^F1A Flan of Puff Paste.^F0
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I looked at the DDT and TECO files, and I think they look ok as is, for now.

36bit:
It may be useful to try this with other Markdown implementations other than GitHub's own version of Markdown. My concern would be that they may not all work well with this due to the lack of standardization across Markdown variations. Like using Pandoc to convert to PDF or whatever.

atsampson:
It works OK with markdown_py and pandoc provided you're converting to HTML. It doesn't work with pandoc to PDF, because pandoc throws away HTML entirely when converting to other formats.

PR #957, opened 2018-06-08
Build hangs on FreeBSD

larsbrinkhoff:
build log: https://thunix.org/~neo_chen/TEMP/its.makelog

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -82,15 +82,16 @@ proc build_macsyma_portion {} {
     respond "302615" "(load \"libmax;module\")"
     respond "303351" "(load \"libmax;maxmac\")"
     expect "307161"
-    send "(todo)"
-    expect "(todo)"
-    sleep 10
-    send "(todoi)"
-    sleep 10
-    expect "(todoi)"
-    sleep 10
-    send "(mapcan "
-    expect "(mapcan "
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
@NeoChen1024, if you could test [this](https://github.com/PDP-10/its/pull/957/commits/cef794b4714d737a8ab765b2ef065e682e6a2bc9), that would be great!

NeoChen1024:
But, it hangs at "(ASUM...", not "(todoi)"

larsbrinkhoff:
@NeoChen1024, yes the last output in the log is "(ASUM...", but the reason the build script fails (I'm hypothesising) is that it doesn't see "(todoi)" in the output.  It's there, but it's broken up into "(", "tod", "io", and ")".

My change in this pull request should fix that.  It works on my machine, and on Travis CI.  But they also worked without this change.  I don't know if it works on your machine, so if you can test it that would be great.

larsbrinkhoff:
@NeoChen1024, according to #963 I see you have built ITS.  Does that mean this issue is resolved?

NeoChen1024:
Yes, it's resolved.

larsbrinkhoff:
Thanks.

PR #542, opened 2018-06-11
Microcode assemblers, and microcode sources

larsbrinkhoff:
SYSENG; MICRO 52  

KSHACK; TS MICRO  
KSHACK; MICRO 116

The first is the one used for KL microcode.  The second is for KS microcode, and is derived from the first.  But it's unclear whether version 116 also can do KL.

There is also a microassembler in KSHACK; DUCK 1, which may be DEC's version.


larsbrinkhoff:
The KL10 microcode listing comes out the same (modulo timestamps) whether assembled with MICRO 52 or MICRO 115.  Still, I'm a little wary of just flushing the old version.

```
;Remaining to be done:
;       Test whether KL10 microcode still assembles

**Comment by larsbrinkhoff** [comment]
It looks like MICRO 116 doesn't work when purified, so that's s difference from version 52.

File: doc/kshack/ainote.8 [hunk 1]
```diff
@@ -0,0 +1,182 @@
+-*- Text -*-				This is the file AI:KSHACK;AINOTE >
```
atsampson: If you're adding this document, you could also add SYSDOC; KL HUNG, KL10 FLKLOR, KL10 INFO from the GPL release (with the copyright header), and MOON; KLPROC 11 from its-vault, which give equivalent advice for the KL?

larsbrinkhoff: Ok, good idea.

larsbrinkhoff:
I used the less verbose copyright notice introduced by @36bit.

larsbrinkhoff:
I wonder why the GitLab CI check shows up sometimes, and sometimes not.

PR #978, opened 2018-06-15
Documentation: KLDCP and "Spells and Incantations"

larsbrinkhoff:
- KLDCP.HLP from Saildart.  The older version which seems closer to the ITS program.
- Christopher Stacy's "Guide to ITS Operations - Spells and Incantations".


larsbrinkhoff:
Thanks, I added all missing directories to DOC.

PR #381, opened 2018-06-16
KERMIT - file transfer protocol

larsbrinkhoff:
SYS3; TS KERMIT  
MATH; KERMIT 170  
MATH; KERMIT DUMPER  
INFO; KERMIT 4  

Lisp program.


larsbrinkhoff:
@eswenson1, since KERMIT DUMPER wants to load LISP; MLSUB FASL, I added the source in this pull request.  I tried to compile it in various places, but I got various errors.  Finally I found a place which worked.  It's needed in COMPLR, so the binary FASL needs to stay.

jar398:
This is amazing.
I do remember making Kermit work, and later of porting it to Common Lisp, and I do know I got some real use out of it, but I have no memory of why I was using it, or between which computers. I expect I only made those features work that I was actually using.

PR #979, opened 2018-06-16
LEMOTE MIPS : build.tcl change

larsbrinkhoff:
Hello,

in order to build under Debian8 (using LEMOTE MIPS loongson2f), the timeout needs to be 1000 at line 110 in build.tcl

Do you think this it's possible to change that value?

thanks
Fausto

### DIFF
--- a/build/muddle.tcl
+++ b/build/muddle.tcl
@@ -16,7 +16,7 @@ respond "*" ":midas ts stinkm_stink\r"
 expect ":KILL"
 
 respond "*" ":xfile assem xfile\r"
-expect "Assembly done!"
+expect -timeout 300 "Assembly done!"
 
 respond "*" ":print mudsav; ..new. (udir)\r"
 type ":vk\r"


larsbrinkhoff:
Thanks, I submitted a change to increase the timeout.

larsbrinkhoff:
I see an unfortunate side effect of turning an issue into a pull request, is that the creation author and date changes.

ams:
There seem to be often issues with the timeouts.  Would it make sense to have a delta parameter that would double, triple, etc, the current timeout by way of a parameter? Or is that not possible?

larsbrinkhoff:
Almost all timeouts happen because the expected output doesn't appear.  In these cases, it doesn't help to increase the timeout.  Rather, you want the timeout to be a reasonably small amount so the build doesn't linger around.

If someone were to, say, build ITS on their C64, then sure a timeout factor would be good to have.  But I'd rather introduce that when there is an actual need.

PR #640, opened 2018-06-16
Build times out waiting for string

larsbrinkhoff:
It always seems to die for me in the same place, waiting on the string EINIT even though it's right there.

[build-log.txt](https://github.com/PDP-10/its/files/1668842/build-log.txt)

Steps to reproduce:

git clone https://github.com/PDP-10/its.git
cd its
git submodule init
git submodule update
make EMULATOR=klh10

I also tried building with simh. It still dies as the same place waiting for EINIT.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,9 @@ $(OUT)/ka-minsys.tape: $(ITSTAR)
 $(OUT)/sources.tape: $(ITSTAR) build/$(EMULATOR)/stamp $(OUT)/syshst/$(H3TEXT)
 	mkdir -p $(OUT)
 	rm -f src/*/*~
+	touch -d 1981-10-06T19:03:37 'bin/emacs/einit.:ej'
+	touch -d 1981-09-19T21:42:56 'bin/emacs/[pure].162'
+	touch -d 1981-03-31T20:41:45 'bin/emacs/[prfy].173'
 	cd src; $(ITSTAR) -cf ../$@ $(SRC)
 	cd doc; $(ITSTAR) -rf ../$@ $(DOC)
 	cd bin; $(ITSTAR) -rf ../$@ $(BIN)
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
This "24 hours ago" didn't work for me on a fresh repository.  So apparently it's something about checking out new files with git, and then starting a build.

I'll check in more detail what's going on with the timestamps.

- What are the Unix atime, ctime, mtime timestamps?
- Which one(s) is itstar picking up and putting in the tape image?
- What is DUMP reading off the tape?
- What are the ITS creation and reference timestamps?

larsbrinkhoff:
I was touching the filestamps of the wrong files.  It's not the source files in EMACS1, but the purified files in EMACS.

larsbrinkhoff:
I tested this a few times by checkout out a fresh repository and building.  It should work now.

Makefile
```diff
@@ -66,6 +66,8 @@ $(OUT)/ka-minsys.tape: $(ITSTAR)
 $(OUT)/sources.tape: $(ITSTAR) build/$(EMULATOR)/stamp $(OUT)/syshst/$(H3TEXT)
 	mkdir -p $(OUT)
 	rm -f src/*/*~
+	touch -d "24 hours ago" bin/emacs/einit.*
```
atsampson: That kind of -d is a GNU extension - it won't work with (e.g.) FreeBSD's touch.

> $ touch -d "24 hours ago" foo
> touch: out of range or illegal time specification: YYYY-MM-DDThh:mm:SS[.frac][tz]

larsbrinkhoff:
Ok, instead of a gross hack which doesn't work on FreeBSD, I'll use the actual original filestamps for those three files.

See also #275.

larsbrinkhoff:
I tried to use itstar's DIR.LIST feature, but it wouldn't work with just those three files.  Because then it only added those three files to the tape image, and ignored those files not mentioned in DIR.LIST.

larsbrinkhoff:
@atsampson, can you test the "tmp" commit on FreeBSD?  If it works, I'll squash it in.

atsampson:
Yep, that touch syntax works fine on FreeBSD:

> $ touch -d 1981-10-06T19:03:37 foo
> $ ls -l foo 
> -rw-r--r--  1 ats  users  0 Oct  6  1981 foo

PR #972, opened 2018-06-18
SHUTDN - Shut down ITS

larsbrinkhoff:
BAWDEN; SHUTDN 6


larsbrinkhoff:
There's an imposed time limit.  You must specify a time at least five minutes in the future.  And there must be a SYS: MAIL DOWN file.

Without JCL, prints information about when ITS is going to shut down.

larsbrinkhoff:
There was no system-wide binary installed.  I picked SYS3.

PR #987, opened 2018-06-18
Fix assorted typos and formatting variations in programs.md.

atsampson:
Some of the choices I've made here are a bit arbitrary - e.g. ARPAnet vs. Arpanet or nickname vs. nicname, where both were used - in these cases, I've tried to pick what seemed to be the most common spelling in the docs/mail archives we've got.

### DIFF
--- a/doc/programs.md
+++ b/doc/programs.md
@@ -11,7 +11,7 @@
 - ADV350, 350-point Adventure.
 - ADV448, 448-point Adventure.
 - ANIMAL, an animal guessing game.
-- ARCCPY, copies and old-format archive, converting to new format.
+- ARCCPY, copies an old-format archive, converting to new format.
 - ARCDEV, transparent file system access to archive files.
 - ARCSAL, archive salvager.
 - ARGUS, alerts you when specified users login or logout.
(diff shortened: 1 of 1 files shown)


doc/programs.md
```diff
@@ -73,9 +73,9 @@
 - FILE, Chaosnet file server.
 - FIND, search for files.
 - FRETTY, display list of free TTYs.
-- FTPS, FTP Server.
-- FTPU, FTP Client.
-- GCMAIL, delete old files from .MAIL.
+- FTPS, FTP server.
+- FTPU, FTP client.
+- GCMAIL, delete old files from .MAIL.;.
```
larsbrinkhoff: I don't think `;` is needed unless it's a full file name with FN1 etc.  But I haven't checked all files to see what the most common usage is.

doc/programs.md
```diff
@@ -164,45 +164,45 @@
...
+- SCANDL, TTY output spy.
 - SCHEME, Scheme interpreter.
 - SCNV, convert plotter files to XGP scan files.
-- SCRAM, encypt/decrypt file.
+- SCRAM, encrypt/decrypt file.
 - SCRMBL/UNSCR, scramble/unscramble file.
 - SEND, REPLY, replacements for DDT :SEND.
 - SENDS, Chaosnet SEND server.
-- SENSOR, sends censor.
+- SENSOR, SEND's censor.
```
larsbrinkhoff: I think "sends" is a plural form here.

doc/programs.md
atsampson: I was going for consistency with TMPKIL ("clean out old files in .TEMP.;."), but equally we could remove the semicolon from the latter...

doc/programs.md
atsampson: I was thinking of it as censoring SEND, but yes, I guess it works both ways!

How about "An alternate message receiver", which is how it's described in DIR?

doc/programs.md
```diff
@@ -164,45 +164,45 @@
...
+- STYLOG, convert PTY output file into ASCII file.
 - SUPDUP, Supdup client.
 - SYSCHK, check up on system job.
 - SYSLOD, system load histogram.
 - SYSMSG, displays system messages.
 - RABBIT, Scheme compiler.
 - TAGS, generate tags table for sources.
 - TALK/WHO/WHOJ/WHOM/USERS, list users.
-- TCTYP and CRTSTY, terminal handling.
+- TCTYP/CRTSTY, terminal handling.
```
larsbrinkhoff: TCTYP and CRTSTY should probably be separate lines.

doc/programs.md
larsbrinkhoff: Yes, that is more descriptive.

PR #991, opened 2018-06-19
Fixes emacs mode comment in ctensr funcs. Adds eigen source for macsyma. 

eswenson1:
Also changes way macsyma files are translated and compiled.

### DIFF
--- a/build/lisp.tcl
+++ b/build/lisp.tcl
@@ -471,6 +471,7 @@ respond ";BKPT" "(quit)"
 respond "*" ":midas maxtul;ts mcl_mcldmp midas\r"
 respond "*" ":link maxtul;.good. complr,sys;ts complr\r"
 respond "*" ":link liblsp;gcdemn fasl,lisp;\r"
+respond "*" ":link maxtul;ts utmcl,maxtul;ts mcl\r"
 
 respond "*" "complr\013"
 respond "_" "mrg;macros\r"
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
TCL hint: You can use `{` `}` to quote a string, then you don't have to use `\` to quote `"`s.

larsbrinkhoff:
The change to lisp.tcl in 54c2914 seems to be a fix for 7411614.  I'd like to move that change to the first commit.  I can do the git incantations.

larsbrinkhoff:
I get this now:

```
(C2) compile_lisp_file(translate_file("tensor\;ctensr funcs")[2]);

Translation begun on DSK:TENSOR;CTENSR FUNCS.

You have run out of LIST space.
Do you want more?
Type ALL; NONE; a level-no. or the name of the space.

all;

You have run out of SYMBOL space.
Do you want more?
Type ALL; NONE; a level-no. or the name of the space.

all;
:$ Calling the compiler $
*:MAXTUL;UTMCL DSK:TENSOR;CTENSR FASL_DSK:TENSOR;CTENSR TRLISP (T)
Macsyma Source Compiler

LISP COMPILER 2155 [in LISP 2154]

Compilation begun on ((DSK TENSOR) CTENSR TRLISP)
;Including file ((DSK LIBMAX) TPRELU /47)
;Loading MAXMAC 227
;Loading MOPERS 49
;Loading DEFOPT 8
;Loading TRANSQ 87
;Loading MDEFUN 57
^G^GTOO MANY NXM ERRORS.
PI LEVEL 7 BUGPAUSE.  FIND A WIZARD, CONSIDER TAKING A CRASH DUMP,
OR TYPE <ALTMODE>P TO ATTEMPT TO REVIVE THE SYSTEM.
YOU ARE NOW IN DDT.
BUGPC/   BLKI MEMHNG+4   $Q-2/   TDNE B,SUPCOR

larsbrinkhoff:
@rcornwell says:

> might run it under gdb and set a break point where it sets nxm then finish to main sim_intsr, "set reason=1" and continue then you can get an instruction trace right up to failing instruction

larsbrinkhoff:
I dropped commit 916cf99 and rebuilt.  I got past compiling CTENSR.

eswenson1:
@larsbrinkhoff I thought, when you said you dropped commit https://github.com/PDP-10/its/commit/916cf993cba7dba5dff2ec4e8867da3491a284c7 that you meant you updated the master branch to no longer include that commit.  It doesn't appear that that is what you mean -- I guess you did it locally?  In any case, do you want me to update the build so that it does the translate and compile as two separate steps (the first in Macsyma and the second in COMPLR)?  Or do you want to leave things the way they are so that we can debug the issue in KA10?

larsbrinkhoff:
Right, I only dropped it locally for testing.   I wouldn't do anything rash to the master branch on GitHub.

I think to have a working build for all three emulators, it would be good to do the workaround.  Translate and compile as two steps.

Of course we should make an issue and point out the commit that broke the KA10 build, so it's possible to debug in the future.

eswenson1:
I can do the workaround.  However, it would be great if you or @atsampson could debug this because the compile_lisp_file(translate_file(...)) approach is the recommended approach -- and what was used in ITS days to do this.  I just didn't know about it when I did CTENSR the first time and therefore did it the hacky way.  It would be therefore, to get this debugged. I take it you aren't up for doing this now -- rather than the workaround?  :-)

larsbrinkhoff:
We certainly should debug it.  I have just entered a period in which I will be focusing a lot on getting programs for the 340 display to work properly.  So I'm not sure how much time I will have left for other ITS issues at this point.  But having a stable operating system for the 340 programs requires fixing problems with KA10 ITS and the emulator too.

eswenson1:
Regarding:

> TCL hint: You can use { } to quote a string, then you don't have to use \ to quote "s.

That's really cool.  I see you used this in scheme.tcl.  I'll try to remember that in the future. Will it also work to not require \ in front of [ and ]?  I had to quote those inside a string too.

larsbrinkhoff:
Yes, I think that should work.  I believe `{` quotes all characters to a matching `}`.

This is the syntax normally used for code blocks, but it works just the same for quoting strings.

PR #126, opened 2018-06-19
TYPESQ - print out a squeezed file

larsbrinkhoff:
SYS2; TS TYPESQ
AR7: GZ; TYPESQ MID1 (not sure if source is available, will have to search for it)

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -748,6 +748,7 @@ expect ":KILL"
 # USQ
 respond "*" ":midas sys2;ts usq_sysen3;usq\r"
 expect ":KILL"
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
Verified with a squeezed file.

I made a Unix tool to convert an 8-bit file to the "TENEX" format.  (Four left-aligned octets and four zero bits to a word.)

larsbrinkhoff:
Thanks, I don't have that in my iptables script.  I was hoping conntrack would do the job.

We should probably have a ready-made script checked in.  Here's mine.  I based it on one by Björn.  
https://github.com/larsbrinkhoff/its.pdp10.se/blob/master/its/iptables.sh

PR #993, opened 2018-06-20
Resolves #724. Build MLMAC from source.

### LINKED ISSUE #724: LISP; MLMAC FASL needs to get built from source
Prior to compiling LSPSRC; MLMAC 92, you have to load NILCOM;SUBLOA > (LISP, not compiled).

eswenson1:
Also reverts translate-and-compile of macsyma programs in one step since
this causes an NXM error in KA10 emulator.  Will put back when we find
issue with KA10.

### DIFF
--- a/build/lisp.tcl
+++ b/build/lisp.tcl
@@ -535,18 +535,48 @@ respond "(C1)" "quit();"
 respond "*" ":link sys3;ts macsym,maxdmp;loser >\r"
 
 ### build ctensr for macsyma
+# ejs: commented out until we can fix NXM error in KA10 emulator
+#respond "*" "macsym\013"
+#respond "(C1)" "compile_lisp_file(translate_file(\"sharem\\;packg >\")\[2\]);"
+#respond "(C2)" "compile_lisp_file(translate_file(\"tensor\\;ctensr funcs\")\[2\]);"
+#respond "Type ALL;" "all;"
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I split this in two separate commits.  Started a full KA10 build to verify.

eswenson1:
I realized, after i’d made both changes and made a single commit, that this would have been better as two commits. But I didn’t know how to do that other than resetting and starting over. Is there a cool way to do that? (I’m sure there is, and that you used it, oh enlightened Git spirit!)

larsbrinkhoff:
Yes, there are many ways.

The big chainsaw of commit rewriting is `git rebase -i`.  Your editor will get a list of commits and operations to apply.  For example, replacing `pick` with `e` will let you edit commits any place back in history.  Note that when you're editing a commit during a rebase, you can also add new commits at that point.

Now, splitting one commit into several separate ones is tricker than the reverse operations.  Usually `git checkout -p <commit> -- <files>` is helpful.  It will let you checkout individual files from a commit, and also pick and choose hunks from the diffs (use `s` to split hunks even further).

So I'll usually use rebase to go back one commit before the one I want to split, and then pick out changes with commit -p and commit them separately.

I understand this isn't for the light of heart.  It's not something I was comfortable with initially.

PR #1017, opened 2018-06-25
PDP-10 maintenance

larsbrinkhoff:
SYS; TS 10RUN  
SYSEN2; 10RUN 11  
MAINT; PART *


build/misc.tcl
```diff
@@ -815,6 +815,11 @@ respond "*" ":midas sys3;ts scrmbl_ejs;scrmbl\r"
 expect ":KILL"
 respond "*" ":link sys3;ts unscr,sys3;ts scrmbl\r"
 
+# KA10 maintenance
+respond "*" ":midas \t sys;ts 10run_sysen2; 10run\r"
```
atsampson: /t rather than \t?

eswenson1: Same comment.

larsbrinkhoff:
The GitLab build should have caught the \t typo.  I wonder why it didn't.  It's supposed to be a full build using KLH10.

eswenson1:
Did the \t get turned into a TAB by expect and could MIDAS have assembled fine without /t?

larsbrinkhoff:
Yes, \t should have turned to an harmless whitespace.  **BUT** the TCL script should have timed out looking for "with ^C".

larsbrinkhoff:
No, because that's another strange thing with GitLab.  There is no build log.  I.e. when expect runs, it doesn't write anything to stdout like it's supposed to do.

larsbrinkhoff:
Right, I have seen it (or not seen it!) in several places.  I have tried to look into it, but I didn't get anywhere.

larsbrinkhoff:
I ran the diagnostics without 10RUN.  Just make a job, load PART A, and $G.  At the end, it will automatically continue with PART B etc.

It fails in PART K:

```
12276/   MOVE 5,13235   /   -100,,-100   $
12277/   AOBJN 5,12277   
12300/   CAME 5,77   /   0
```

This is strange, because it just counts up from -100,,-100 until AC 5 is no longer negative.  The test expects the accumulator to be 0 as this point, but it's 1,,0 which is the expected behaviour on a KA10.

eswenson1:
Does it do this on the other emulators? Is this expected behavior on all 10s?

larsbrinkhoff:
I only tested the ka10 simulator.  According to 10RUN, it seems the PART files are to be run from PDP-6 ITS, but loaded into a KA10.  But since the original timestamps are from 1975, maybe that theory is wrong, and they are for KL10 instead.

On a KA10 (or a Foonly), AOBJx is expected to overflow from the right half to the left half, so when -1,,-1 is input, the result is not 0,,0 but 1,,0.  But on newer processors, this quirk has been fixed.  They add 1 to both halves separately without any carry from the right half to the left.

larsbrinkhoff:
MAINT; PART K.OLD (not in this repository yet) has the AOBJN test for the KA10.  So PART K was patched, probably for the KL10.

larsbrinkhoff:
CC @rcornwell, here's another set of diagnostics for you.  Or maybe the PART A through M fles are closely related to DEC's DAKxx files.

larsbrinkhoff:
PART F has a MUL overflow patched out.  And some IMUL check too.

PART G has a floating point value adjusted, and some parts jumped over.

PART K has some changes other than the AOBJN result.

larsbrinkhoff:
Continuation: #1622

PR #493, opened 2018-06-25
CRC - compute file CRC

larsbrinkhoff:
SYS3; TS CRC  
AR3: GREN; CRC 70


larsbrinkhoff:
Yes, there were three trailing NULs which I have now removed.  Thanks, Eric.

PR #1031, opened 2018-06-27
Workaround for 340 hangage

larsbrinkhoff:
ITS uses a CONO instruction in the second interrup locationt used by the 340 special interrupt.

That's explicitly forbidden by the PDP-10 Processor Referece Manual.  My (unproved) theory is that there was a modification to allow this and also dismiss the current interrupt.

The workaround would be to replace the CONO with a JSR to a CONO.


larsbrinkhoff:
I pushed a commit.  The workaround should look like this, but I haven't tested it yet.

Let's mull over this for a while.

larsbrinkhoff:
Now tested and works fine.

larsbrinkhoff:
Improves FED.  The font is displayed, but there is an "APR ERROR IN NULL JOB" every now and then.  Restartable BUGPAUSE, but still not good.

larsbrinkhoff:
Also improves Spacewar which runs for a second and then crashes in the same way as FED.

larsbrinkhoff:
It turned out this isn't a complete fix.

larsbrinkhoff:
This should now be fixed in the latest KA10 simulator version.

Here's the explanation:  
http://pdp-6.net/asmhacks.html#pisystem

Issue #684, opened 2018-02-26
340 display without deselection device


larsbrinkhoff:
ITS uses the DSDEV to coordinate use of the 340 display.  Those two places could probably put inside DSDP.

larsbrinkhoff:
Or maybe don't bother.  DATAO to an unknown device is ignored anyway.

larsbrinkhoff:
Reopening to fix this.  The assembler errors look ugly.

larsbrinkhoff:
Fixed by #1033.

PR #1033, opened 2018-06-27
Fix BBLK and DSDEV in ITS

larsbrinkhoff:
When assembling KA ITS, I get this error:

    DRECYC+1        70756    2.    36-010   LOSS AT BBLK

It's because there's an unmatched BBLK due to conditional assembly.


larsbrinkhoff:
I'm expanding the scope to also include #684.

larsbrinkhoff:
There are some more uses of DSDEV outside DSDP, but inside PDP6P.  I think this is ok, because PDP6P should imply DSDP.

larsbrinkhoff:
Also default KLH10P to 0.  Fixes undefined symbol error.

PR #1034, opened 2018-06-27
Update FAIL manual

larsbrinkhoff:
This is a version from December 1975 sent in by @rmaldersoniii.

From the DECUS library of TOPS-20 programs.

### DIFF
--- a/doc/sail/fail.manual
+++ b/doc/sail/fail.manual
@@ -36,13 +36,10 @@
 
 
 
-
-
-
                             Sponsored by
                   Advanced Research Projects Agency
                          ARPA Order No. 2494
-Stanford Artificial Intelligence Laboratory                  May 1974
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I don't know which manual version describes our FAIL best.

I originally added a FAIL binary from ITS backups.  There was no source, but the binary had information it was assembled from a file called FAIL 143.

Later I got the lastest FAIL from Saildart and dubbed that FAIL 200.

Now we do have the original FAIL 143 source, so it's possible to analyze the differences between that, the latest SAIL version and the TOPS-20 version.

larsbrinkhoff:
KA10 ITS fails with a disk full error.  Time to add more RP03 packs?

larsbrinkhoff:
The push build failed, but the pull request build was ok.  Must be on the fence regarding disk space.

Better wait with this until the new disk packs have been added.

PR #1035, opened 2018-06-28
KA ITS disk full

larsbrinkhoff:
The KA10 ITS build fails due to full disk packs.


larsbrinkhoff:
@rmaldersoniii and @matthewt, you have done this kind of thing before.  Can you please check these changes and the added documentation?

eswenson1:
I’m a bit surprised by the directories out of phase and need to run UCOP. When I ran out of disk space on ES, I made a new NSALV, booted with it (and the orig ITS). I formatted the new pack and then booted ITS. Then I built a new ITS with an additional pack. Then I shut down and booted with the new ITS. I never had to run UCOP.

larsbrinkhoff:
I was thinking maybe it was a KA10 thing, since SALV is a little bit different from NSALV.  But NSALV also has a "Directories out of phase" message in the corresponding location.

That's not to say that location is ever called in NSALV.  Maybe, maybe not.

larsbrinkhoff:
Eric, that sounds great.  It'll be a good addition to the ADDING PACKS file I included here.

No rush of course.

rmaldersoniii:
@larsbrinkhoff I had a look at the document you've created. It looks good to me.
@eswenson1 It seems to me that I had to run UCOP if the disks were added later, rather than all created at the same time. A copy of the MFD is present on every system disk; there seem to be traces of private disks not part of the system disk group in a few places, but I've never tracked all that down.

eswenson1:
@rmaldersoniii You may well be right -- however, I never ran UCOP on ES ITS when I added a second RP06 disk pack.  And I don't get errors on boot that tell me that I need to.  I added the second drive after the fact (a couple years after I created the system).

TheFausap:
There's only a little typo in the doc:

FIRSTPLK and LASTPK

should be

FIRSPK and LASTPK

eswenson1:
Just wanted to confirm something.  I just added a second disk to a KLH10 DB machine built from a very recent HEAD of the master branch.  I never ran UCOP, although I did see the initial warning from NSALV CHKR that "you will have to UCOP to it".  I ignored that message.  I had no issues and ITS came up fine.  Then, I shut down ITS, booted DSKDMP, loaded NSALV and ran CHKR again.  and didn't get any errors/warnings.  I believe ITS wrote out the directories on the packs automatically.  Here is what CHKR said to me after shutting down ITS (after booting with new second pack):

>  DSKDMP
> l$ddt
> t$nsalv bin
> CHKR$G'
> 
> Salvager 262
> 
> Active unit numbers? 01
> Unit #1. ID is DISK2, Pack #1 (hardware says 0.)
> 
> Unit #0. ID is FOOBAR, Pack #0 (hardware says 58562.)
> 
> Use MFD from unit: 0
> Read all blocks of all files? (Y or N) y
> Get user dirs from unit: 0
> 
> CRASH has no files, User File Directory DELETED
> .TEMP. has no files, User File Directory DELETED
> Write out changes in MFD (Y or N) y
> 
> DDT

eswenson1:
I tried another experiment, which, I believe proves me wrong in the NEED to run UCOP.  I ran DSKDMP, loaded NSALV, ran CHKR, and told it to use the MFD from unit 0 but the user dirs from unit 1.  I got this error:

> UFD on unit 1 block 373 is .TAPE0, but expected .Correct it? (taking MFD entry as good) (Y or N)

Now, the fact that there are *any* UFDs on unit 1 is perhaps not surprising, because I expected ITS to write them there.  But the fact that one was WRONG suggests something isn't quite right.

So I might be living dangerously on ES ITS now with only one copy of the UDIRs on unit 0.  I'm going to try another experiment -- on my dual-disk DB system, I'm going to NOW run UCOP and see what happens.

larsbrinkhoff:
@eswenson1, we don't have any explicit protocol.  I think we can be informal.

I fixed the topos @TheFausap pointed out.  Thanks!

Eric, you can amend the latest commit with your information.  And then this can go in.

eswenson1:
How 'bout if you commit and when I get a chance, I'll update adding.packs in another commit.  Did you fix your reference to UCOP?  (You mention you fixed typos @TheFausap pointed out, but you didn't mention my UCOP comment).

rmaldersoniii:
Regarding adding.packs, the correct wording is "+Use UCOP in SALV to update the MFD on the fresh pack with UDIR information from the MFD of a pack already in use."

(IMAO)

larsbrinkhoff:
Sorry, I missed the comment about UDIRS.  I don't know what UDIRS is really.  I thought "COPY UDIRS" would be the same as copying the MFD, but I see now it could just as well be copy all UFDs.  Or something else.  I don't see that "UDIRS" is used anywhere else.

I'll use @rmaldersoniii's suggested wording, and then I'll check this in.

eswenson1:
The MFD is the master file directory (directory of directories).  The UFDs or UDIRs are the actual directories, like EJS; and SYSENG;)  UCOP (see the doc in NSALV) copies these directories from one pack to another. Normally ITS pushes all writes to any directory (MFD or UFD) to all packs.  But it only does this on writes.  For all the directories already on a pack, the only way to get them to the other packs (backups) is to write to them -- or use UCOP to manually copy them all.

larsbrinkhoff:
I did read the N/SALV doc, but since UDIR isn't defined anywhere but in N/SALV code I did't understand exactly what it does.

I suppose UDIR is exactly the same as UFD?  In that case, I think ADDING PACKS should say UFD, not UDIR.

rmaldersoniii:
@larsbrinkhoff For more details on the MFD and UDIRs, see pp. 22-23 and 42-44 of AI memo 161A.
Come to think of it, I have a corrected PDF of that manual--the one at CSAIL's archive is missing page 19, which I scanned in from MRC's copy (now at LCM+L) and added. I'll send it to you for the documentation repository.

larsbrinkhoff:
Ok, I looked briefly at what UCOP in SALV does.  I didn't go into details, but I think it looks like it first copies the MFD, and then all UFDs.

larsbrinkhoff:
Thanks for the AI memo, Richard!  I think the MIT library would be interested too.  I have talked to them, so I can pass it on.

larsbrinkhoff:
My suggestion would be to use this phrasing:

Use UCOP in SALV to update the MFD and UFDs on the fresh pack with information from a pack already in use.

eswenson1:
I'm sure that is fine. The wording in NSALV is a bit different, but I don't think it is worth fretting over.

eswenson1:
Sorry, I read your messages in reverse. My reading was also that it copies UFDs, not just the MFD so I think that "Use UCOP in SALV to update the MFD and UFDs on the fresh pack with information from a pack already in use." is great.

larsbrinkhoff:
@rmaldersoniii, I pointed out the missing page 19 to MIT in May.

It looks like they added it since:
https://dspace.mit.edu/bitstream/handle/1721.1/6165/AIM-161A.pdf

rmaldersoniii:
Cool. I downloaded my copy in February 2005, and checked for updates for years, until I finally updated my copy from Mark's in February 2017. In the mean time, I had a note pointing to the relevant pages of the ITS 1.4 manual (Memo 161) for the information missing with page 19.

larsbrinkhoff:
It turns out there's a problem with this change.  A full KA ITS build now times out when STINK is run.  See #1082.

PR #236, opened 2018-06-30
MAZE - game

larsbrinkhoff:
I see references to a game called MAZE, or [Mazewars](https://en.wikipedia.org/wiki/Maze_War).

There's an EMACS library for editing mazes, and a sample MAZE MAZE.

The file COMMON; MAZE DOC contains messages about the origin of MAZE, and also the Imlac PDS-1 source code!


larsbrinkhoff:
I mentioned COMMON; MAZE DOC earlier.  I'm not entirely comfortable including that file here, because it contains email correspondence between individuals.

larsbrinkhoff:
I came across this interesting message on how MazeWar ran at the Architecture Machine group:

> I remember playing MAzeWars between multiple IMLAC computers at both the MIT Architecture Machine Group and at Project MAC around 1974.  At that time it was not used much since it was old hat and evrybody at been hacked out on it.  The IMLACs were vector graphics single user computers several of which would use some common machine to act as control (it just passed messages, didn't
actually participate).  In the case of ArchMach the host was an Interdata machine running the MAGIC operating system, at Project MAC (now LCS) the host was a PDP-10 (MIT-DM) running ITS.  It was said that the version on the PDP-10 could use the network to allow multiple net hosts to act together as one control and that Stanford (or someone out west anyway) also had an implementation and that cross-country games had been played.  The rumor was that this was stopped when ARPA noticed that more than 50% (the number may be apocryphal) of the traffic during some period was cross-country MazeWar packets.

larsbrinkhoff:
Project MAC Progress Report XI says this about the Maze game, I mean teleconferencing appliation:

teleconferencing

jonahharris:
Does anyone happen to know if the Oracle SQL*Net version is still around? It likely used the old TNSAPI library which, while the headers are no longer included with Oracle Database, I can probably still get to work with the latest versions using my implementation of the library.

larsbrinkhoff:
Hello @jonahharris,

I have passsed on your question to some ex-Oracle people.  Let's hope it does exist, and if so possible to make public.

larsbrinkhoff:
I snarfed this article off imgur.  "The First First-Person Shooter", Computer Games Magazine, July 2005.
FPS1
FPS2
FPS3

larsbrinkhoff:
Chris Kantarijev's X10/X11/Sunview Mazewar is here: https://github.com/larsbrinkhoff/sunview-x-mazewar

The UCB version by Chris Guthrie, Nick Lai, and John Coke is here: https://github.com/hackerb9/ucbmaze1984

larsbrinkhoff:
Looking for:
- [ ] Alto Mazewar
- [ ] MIDI Maze
- [ ] "Oracle" Mazewar

jdersch:
I have binaries for a couple of versions of Alto Mazewar (you can find one of them here: http://xeroxalto.computerhistory.org/Io/Murray/.index.html -- see MazeWar.boot).  I may also have Mesa sources for same.

larsbrinkhoff:
Thanks @jdersch!  I head elsewhere that Mesa sources are not yet available to the public, so I guess I'll have to be patient.

PR #1046, opened 2018-07-02
COOKIE BEAR - prank

larsbrinkhoff:
Prank.

I didn't examine this in detail, but the code seems to make use of some tricks maybe for hiding its identify.

I'm not sure how to use the CHECK FOO file.  Looks like it should set up a translation for HACTRN and then reload it from TS CHECK.


larsbrinkhoff:
Hold on a bit.  I don't see how someone could send a message back to the cookie bear.  Maybe there's some special magic I haven't done.

doc/programs.md
```diff
@@ -39,6 +39,7 @@
 - COMIFY, convert HEX to COM format.
 - COMPLR, Lisp compiler.
 - COMSAT, mail server.
+- COOKIE BEAR, prank.
```
atsampson: Something like "- CHECK, Cookie Bear prank." might be clearer, although I appreciate clarity may be missing the point...

larsbrinkhoff:
That could be part of the puzzle.  Maybe you should login as BEAR and then :XFILE CHECK FOO.  But then I don't see how to pass the victim user name on the JCL to TS CHECK.

larsbrinkhoff:
Yeah ok, maybe that's just as well.

@atsampson, yes I'd like to preserve a little bit of the mystery.

Issue #1049, opened 2018-07-07
Update macsym;macsym tags after Macsyma build


eswenson1:
The macsyma build creates macsym;macsym ntags, but this must be moved/renamed to macsym;macsyn tags once it is verified to be correct.

Issue #1006, opened 2018-06-21
MACSYMA : ARRAY SHARE package missing 


TheFausap:
Hello,

it seems the file "array fasl" (in the share directory) is missing.
The source code should be called like "marray >"

```
DB ITS.1648. DDT.1546.
TTY 11
You're all alone, Fair share = 5%
:login fausap 
To see system messages, do ":MSGS<CR>"
macsym↑K!

This is MACSYMA 999

(C1) array(zz,8);
(D1)                                 ZZ

(C2) fillarray(zz,2.3);
Could not find file which matches "DSK:SHARE;ARRAY FASL"

(C3) 
```

regards,
Fausto

eswenson1:
Thanks! I’ll fix.

eswenson1:
I don’t think the source is marray >.  I think it is:

MAXSRC; ARRAY 5

I also know that there exists version 1, 2, 3, and 4.  I have MAXSRC; ARRAY 1, but believe the SHARE; ARRAY FASL that we have was compiled from MAXSRC; ARRAY 3.  So once I get versions 3, 4, and 5, I’ll try them all and update the build with one that works best.  I’ll compare the three sources to see what the differences are and whether I think it is safe to upgrade from 3 to 5.

eswenson1:
I got MAXSRC; ARRAY 5 and when I compiled it to SHARE; ARRAY FASL and retried array(zz,8); this is what I got:

> macsym^K!
> 
This is MACSYMA 999
> 
> (C1) array(zz,8);
> Call to an undefined function 'ARRAYP' at Lisp level.
> 
> Error during display
> 
> (C2)

So there is more to this.  Gotta find ARRAYP now.

eswenson1:
Interesting, ARRAYP is defined in LSPSRC; MLSUB 17.  It is compiled to LISP; MLSUB FASL.  If you breakpoint Macsyma and load LISP; MLSUB FASL, and then retry the array(zz,8); you get a result that appears like it might be correct:

> Back to MACSYMA, at LISP level.  LISP-BREAK.
> (load "lisp;mlsub fasl")
> 139217
> $p
> (C2) array(zz,8);
> (D2)                                                           ZZ

But if you then call fillarray on zz, you get this:

> (C3) fillarray(zz,2,3);
> 
> ARRAY FASL DSK SHARE being loaded
> DEFMAC FASL DSK LISP being loaded
> Loading done
> Loading done
[...]

Issue #976, opened 2018-06-15
Compile LISP; DEFSET FASL from source


larsbrinkhoff:
Probably NILCOM; DEFSET 107.

eswenson1:
Yes, this does appear to work and be the right source.  The current lisp; defset fasl is version 106 but nilcom; defset 107 appears to be very close and probably fixes a bug.  I'll include in a commit soon.

Issue #967, opened 2018-06-11
MACSYMA : example command missing function


TheFausap:
Hello,

I got the same problem as DECLARE command, with "example" command:

![screenshot from 2018-06-11 14-59-37](https://user-images.githubusercontent.com/1846443/41232961-3c1be52e-6d88-11e8-99df-a7ec95d9e505.png)

In this case I found a definition of FEXPRCHK in maxdoc/mdoc.47, at least an example, also in intro.errors.
I do not understand if this a user-defined function I have to load into macsyma by hand, or can be defined during the build process.

thanks,
Fausto

eswenson1:
Without having (yet) looked at this, I'd say it is a problem with the build process. There was no documentation on the correct way to build Macsyma and we don't know whether we have a consistent set of sources. It is quite possible that there is a missing dependency here. In cases like this one, where you get an undefined function at the Lisp level, it is either that a required macro was not available at compile time -- manifesting as an undefined function at run time, or that a required external function is not loaded (or autoloaded) in the environment.  The documentation in mdoc.47 does suggest that this function should be available.  I'll look around to see if I can find it defined anywhere, and if so, try to understand why it is not in the dumped Macsyma image (or an autoload).

eswenson1:
FEXPRCHK is supposed to be defined in MACSYM; SUPRV FASL.  MAXDOC; DCL AUTOLD defines an autoload property for 'FEXPRCHK to load it from SUPRV FASL.  Interestingly, JPG; SUPRV FASL, the source for MACSYM; SUPRV FASL, defines a Macsyma macro FEXPRCHECK (not FEXPRCHK).  More investigation is necessary -- it is clear that we have some version mismatches between sources used to build Macsyma.  It isn't (yet) clear to me which reference or definition is old.

eswenson1:
It is interesting, though, that there is this comment in MAXDOC; MDOC.47:  

> Demo:		;; One might implement FEXPRCHK as follows
>		(DEFUN FEXPRCHK (L FN)
>		  (IF (OR (NULL L) (CDR L))
>		      (MERROR "Wrong number of args to ~:@M~%~M" FN L)))

[...]

PR #1050, opened 2018-07-07
Resolves several macsyma issues

eswenson1:
#1049 fix macsyma tags
#1006 support macsyma arrays
#976 compile lisp;defset from source
#967 fix example function in macsyma
#960 fix declare and polarform functions in macsyma


larsbrinkhoff:
Sorry, I'm traveling now and can't do a review now.

PR #1030, opened 2018-07-14
SYSTEM GEN - low level DECtape formatting

larsbrinkhoff:
SYSENG; SYSTEM GEN

Used to format DECtapes.  Uses the old Type 551 PDP-6 device.

Ancient file?  DECtape garbage at the end.


larsbrinkhoff:
This hasn't been tested due to a lack of Type 551s.  Also, a particularly low-level simulation would be needed.

Thanks for @aap for pushing this forward and finding a fix in MACDMP.

larsbrinkhoff:
Even MIDAS 108 from 1973 won't handle the single !.  Works fine with two.

PR #1002, opened 2018-07-16
Too many NXM errors

larsbrinkhoff:
This is a problem in KA10 ITS only.  When Macsyma translates a file to Lisp and calls COMPR, there's a BUGHLT (actually PAUSE) "TOO MANY NXM ERRORS".


larsbrinkhoff:
This also reverts the reverted change to use compile_lisp_file from Macsyma.

Let me do a full build here on my own computer to verify.

larsbrinkhoff:
There was a conflict when I reverted the commit, and I remembered you removing those lines from reviewing an earlier pull requests.  See, code review is good!

larsbrinkhoff:
Works.

(Build hangs in sail.tcl, but I'm investigating that separately.)

PR #1071, opened 2018-07-17
Update ITS documentation

larsbrinkhoff:
SYSDOC; .CALLS > says the Rubin 10-11 address range is 2,,0-2,,777777.  The code says it's 3,,0-3,,777777.

### DIFF
--- a/doc/sysdoc/_calls.127
+++ b/doc/sysdoc/_calls.127
@@ -4255,13 +4255,13 @@ T11MP:	PDP-10 to PDP-11 map
 	PDP-10 pages are mapped into segments of variable size in the
 	PDP-11's memory.  The PDP-10 to PDP-11 page map is also directly
 	addressable in memory (it currently lives in absolute locations
-	2,,776000 - 2,,776377).  There are 256. such entries, one for each
-	page in the range 2,,0 - 2,,777777, which are the addresses fielded
+	3,,776000 - 3,,776377).  There are 256. such entries, one for each
+	page in the range 3,,0 - 3,,777777, which are the addresses fielded
 	by the PDP-10 to PDP-11 interface.  ITS allocates these pages
 	and their corresponding map entries randomly as needed.
 
 	When the PDP-10 attempts to access a physical address in the range
-	2,,0 to 2,,777777, the interface uses the 776000 bits to index
+	3,,0 to 3,,777777, the interface uses the 776000 bits to index
 	the map words.  If the appropriate map word does not have its
 	valid bit set, the memory is considered non-existent.  The same
 	is true if the reference would violate the protection bounds.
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
This can hang around for a while in case we need more updates.

larsbrinkhoff:
I think maybe the PDP-6 16K, the Rubin 10-11 control page, and the mapped pages were moved around in the address space.  There are commented out sections in ITS source code that hints at this.

larsbrinkhoff:
I think we should update SYSDOC; TV 100 too.  It could mention job blinkers.

@Devon7, I'm not sure what those blinking bars do, and I don't see it describe anywhere.  From looking at the ITS source code, it seems like there's ever only one location on the display that blinks.  So when the current job runs, it lights up, and when it sleeps the bar fades away.  Does that sound right?

(It would be cool if there could more of them.  Like one for each logged in user or something.)

atsampson:
I'm sure I've seen a description of this somewhere - the best I can find right now is <news:uwu307hx1.fsf@news.dtpq.com> from Christopher Stacy:

> On the Lisp Machine, the "lights" were drawn on the bottom of the screen, in the form of small horizontal bars that would blink on when active.  There was one for CPU activity, one for disk activity,            
one for GC activity, and I think another one.   Also down there was a higher-level status message that told you the state of the current process (as determined by the scheduling software or alternatively              
by the application) - the default run state message was "Run". There were also progress bars and messages down there that showed all network activity, file IO progress, so you could tell what the              
machine was doing.  And of course, a clock, and one or two other useful status thingies related to Lisp.   The low-level run bars were operated by the microcode - they really were like machine lights.          
>
> This was actually just a fancier version of what we had on ITS on the PDP-10 (with the displays bitmapped into the 10's address space).

larsbrinkhoff:
So far I only see one kind of blinker in ITS, and it's for the running job.  There's also another kind of blinker in ITS, but it's just one cursor for each frame buffer.

ams:
To expand a bit on Christopher Stacy comment, during early boot of the Lisp machine MCR we have three types of lights (this an XBUS address): 

```
A-DISK-RUN-LIGHT ((BYTE-VALUE Q-DATA-TYPE DTP-FIX) 77051763)
				;LOCATION IN TV BUFFER ILLUMINATED WHEN DISK TRANSFERRING
				;THAT + 2 IS THE COMPLEMENT (FOR EASY COMPARISON)
				;THAT - 2 IS THE SCAVENGER RUN-LIGHT
```
Later though, when everything is loaded we have more of a mode-line style thing, called the `who line'.   This shows date/time, user ID or process, current package, process state, and some other stuff.  This is handled in `SYS:WINDOW;WHOLIN LISP`.  This is not the same though as the mode-line in Zmacs, that is another thing all together.

larsbrinkhoff:
Thanks @ams.  FYI, ITS also has wholines.

larsbrinkhoff:
About the control page for the 10-11 interface.  STUFF 92 has what looks to be an older copy of the text in .CALLS.  It says the control page is 776000-776777.  So when that was written, there was a conflict between that and the 256K moby memory.

larsbrinkhoff:
@aap, the various lists of TV displays and keyboards can't possibly apply any more, right?

All, what to do?  Keep the old information for the historical flavour, or throw it out and rewrite?

aap:
I'd leave 'em in. I have a similar list in my TV code: https://github.com/aap/tv11/blob/master/tv.c#L79

larsbrinkhoff:
I found another error related to inter-processor interrupts.

PR #1076, opened 2018-07-18
CCONS - CONS console, Datapoint emulator, and graphics display

larsbrinkhoff:
CONS; CCONS BIN  
LMCONS; CCONS 478

"STUFF LISP" writes the BIN file to the GT40 attached to 10-11 Unibus 2.


larsbrinkhoff:
Supposedly "the same" as CC; CC LISP, except the former is written in Lisp.  So wouldn't that be running on the LispM processor rather than CCONS which runs on a PDP-11?  Maybe the 11 console came first, and it migrated onto the main CPU later.

larsbrinkhoff:
@ams or anyone else.  I see this PDP-11 code will accept commands from the PDP-10.  Something about doing disk I/O.  But I have no idea which PDP-10 program would have sent those commands.  Any clue?

larsbrinkhoff:
This program evolved from VT07 (#942) described in AI WP 59.

ams:
> @ams or anyone else. I see this PDP-11 code will accept commands from the PDP-10. Something about doing disk I/O. But I have no idea which PDP-10 program would have sent those commands. Any clue?

Sorry, no idea.

larsbrinkhoff:
Setting KBDP assembles code for a "fancy keyboard".  The translation table XTAB is virtually identical to the one in TV-11.

larsbrinkhoff:
About disk I/O.  I found this in LMMAN; FILES TEXT:

```
        Eventually, Lisp Machines will be able to support
their own file sytems, or use a special purpose "File Computer"
over the Chaosnet.  At the moment, the prototype Lisp Machine
uses the A.I. PDP-10 file system.  To allow it
to access the PDP-10 (which is not yet attached to the Chaosnet),
a special program must be run on the PDP-10, which is invoked
by typing :lmio;file to DDT.
```

The latest version of LMIO; FILE MIDAS is for Chaosnet.  Maybe LMI is the earlier version?

larsbrinkhoff:
LMDOC; CONS TEXT

```
        The CONS machine has a special diagnostic interface
by means of which a PDP-11 (or a machine with a compatible I/O bus
such as another CONS machine or a pdp10 equipped with a Ten-Eleven interface.)
can start and stop the processor,
examine various internal registers, and forcibly execute
specified instructions.

        To the PDP-11, the interface looks like
a set of 8 readable registers and 4 writable registers.
It should be noted that although the writable registers
are at the same UNIBUS addresses as four of the read-only registers,
there is no logical correspondence between the two sets,
and so it does not make sense to use PDP-11 instructions
which perform a read-modify-write on these registers.
The base UNIBUS address for the diagnostic interface registers
is currently 164000, and the standard symbolic name for the
base address is CONS.

**Comment by ams** [comment]
FWIW, this would be the SPY interface — the CADR also has it.

**Comment by larsbrinkhoff** [comment]
Important information for loading CCONS:

```
                        LOADING CCONS

        CCONS IS USUALLY LOADED VIA THE REGULAR GT40 ROM BOOTSTRAP
(STARTING AT 166000) AND USING THE REGULAR GTLOAD PROGRAM.
THE BINARY IS NORMALLY STORED AS CONS;CCONS BIN.  IMPORTANT NOTE!!
CCONS IS MORE THAT 4K OF CODE, AND THE GT40 ROM BOOTSTRAP USES
AN AREA JUST BELOW 4K IN MEMORY FOR TEMPORARY STORAGE.  THIS WILL CAUSE
LOSSAGE UNLESS THE COMMAND TO GTLOAD IS TERMINATED WITH AN <ALT-MODE>!!.
WHEN THE COMMAND IS TERMINATED WITH AN ALT-MODE, GTLOAD "CYCLES THRU TWICE",
THE FIRST TIME LOADING GT40;GT40LL BIN AND THE SECOND TIME THE DESIRED
FILE.  GT40LL IS A LOADER JUST LIKE THE ONE IN ROM, EXCEPT IT USES
AN AREA JUST BELOW 37000 FOR THE TEMPORARY STORAGE.
        CCONS SHOULD NORMALLY START AUTOMATICALLY UPON SUCCESSFUL LOADING,
IT CAN BE RESTARTED AT 1000 AT ANY TIME IF NECESSARY.
```

From LMCONS; CCONSD 14.

Issue #899, opened 2018-05-17
Add floating point instructions to 11SIM


larsbrinkhoff:
The Dazzle Dart game (#868) uses floating-point instructions.  11SIM doesn't support those, so running the game results in a crash.  Implement at least those instructions needed to run the game.

larsbrinkhoff:
Maybe it's in 11SIM 434.

The version we currently have is 11SIM 426, timestamped 1984-07-08.  Later versions (430-434) have earlier timestamps (1979-06-13 - 1981-10-24), so maybe the 426 is a copy with a new timestamp.

larsbrinkhoff:
I added disassembly of FP11 instructions on the lars/11sim branch.

larsbrinkhoff:
It wasn't in version 434.   See #928.

larsbrinkhoff:
FP11 instructions found in Dazzle Dart:  
LDF, SUBF, ADDF, STF, LDCIF, TSTF, CMPF, MULF, NEGF, CLRF, ABSF, STCFI, DIVF.

larsbrinkhoff:
The floating point number format is: one sign bit, eight exponent bits (bias 128), and then 23 or 55 bits of magnitude.

larsbrinkhoff:
#1085 added disassembly, and #1085 added typein of FP11 instructions.

larsbrinkhoff:
Load/store/clr/neg/abs are easy.  Convert/add/sub/mul/div are a little more work.

I'm planning to convert FP11 numbers to PDP-10 floating point, do the arithmetic, and then convert back.  Inaccurate, but probably good enough to run Dazzle Dart.

PR #1085, opened 2018-07-21
 Disassemble FP11 floating point instructions in 11SIM. 

larsbrinkhoff:
This decodes and prints FP11 instructions in the 11SIM debugger.

It's part of #899.  The two other parts would be entering instructions, and interpreting them.


larsbrinkhoff:
What about updating %S, I hear you say?  Ok fine, I'll look into that.

larsbrinkhoff:
There.  How about this?

PR #1092, opened 2018-07-22
Updated build/lisp.tcl to not specify FN2 for those lisp compilations

### LINKED ISSUE #1091: lisp.tcl builds wrong version of LIBDOC; TIME
lisp.tcl compiles LIBDOC; TIME KMP8 rather than LIBDOC; TIME KMP9.

eswenson1:
where the default of > would be more appropriate. This fixes, in
particular the incorrect compilation of LIBDOC; TIME KMP8 instead of
LIBDOC; TIME KMP9. Resolves #1091.


eswenson1:
Yes. There have been times when the next version of a libdoc source changed authors and versions, and where > would do the wrong thing in the face of both versions. But that doesn’t apply here.

On the other hand, it isn’t clear to me that this version of LIBDOC; TIME KMP8 should be called that since KMP didn’t make the change. We probably should have used yours or my initials. Had we used mine, that issue would have arisen since KMP8 is > EJS9!

PR #437, opened 2018-07-25
CARPET - remote PDP-11 debugger

larsbrinkhoff:
SYSENG; CARPET 107  
.INFO.; CARPET RECENT

CARPET is a PDP-11 debugger written in MIDAS.  It can access PDP-11 memory directly. or...  
HALI is a remote debug stub which runs in the target machine.  It talks to CARPET over a serial line.

The name is probably a play on RUG, a debugger written in PDP-11 assembly language.


larsbrinkhoff:
I have tested this, and at least some things work.  I could load a PALX program and write it to PDP-11 memory.  I can inspect memory symbolically.  I could not start the program.

I tried to assemble the HALI version, but two symbols are missing.

larsbrinkhoff:
I have tested this, and at least some things work.  I could load a PALX program and write it to PDP-11 memory.  I can inspect memory symbolically.  I could not start the program.

I tried to assemble the HALI version, but two symbols are missing.

larsbrinkhoff:
I don't have CARPET RECENT yet.  There's no other documentation that I can see.

I think I should attempt to write a brief CARPET ORDER and add it to this pull request.

larsbrinkhoff:
Apparently typing `:BOOT` does something.  After that, I can run the program with `;G`.

larsbrinkhoff:
Single stepping with ^N didn't work at the first attempt.  But maybe I'm not doing it right.  Somene will figure it out later, I guess.

PR #385, opened 2018-07-26
NAMDRG - name dragon

larsbrinkhoff:
SYSEN2; NAMDRG 148  
SYSEN2; TVKBD ROOMS  
SYS; DOWN MAIL  
CHANNA; NAMPIC  
CHANNA; LOGOUT TIMES  
CHANNA; \_DRGN_ TIMES  
SYSBIN; TTLOC DATA


larsbrinkhoff:
I just want this pull request posted here, so it's ready to go when @aap's Knight TV is up and running.

larsbrinkhoff:
I just want this pull request posted here, so it's ready to go when @aap's Knight TV is up and running.

I tested it, but it dies early due to lack of an NF11TY symbol in ITS.  Maybe NAMPICs need to be added too.

larsbrinkhoff:
@eswenson1, I don't think there's any conflict between NAMDRG and PFT.  Maybe there is between NAMDRG and NAME.

larsbrinkhoff:
I see in MC: MAPS; AI7512 1172 there was a link from CHANNA; RAKASH NAMDRG to SYSBIN; NAMDRG BIN.  So that's how it was started.

This link doesn't appear in other records, but maybe they don't show links!

larsbrinkhoff:
I have now tested this with the TV-11.  It turns out it refuses to run because the TTY # of the free display has changed.  NAMDRG expects 42. but now it's 9.  That's because there's no TTY scanner.

larsbrinkhoff:
NAMDRG checks the value of the ITS symbol NF11TY.  That's the first Knight TV TTY.  The value is computed as the sum of all other kinds of TTYs: consoles, TK10 TTYs, Datapont, Morton, etc.

Later it .OPENs TTY T52 to get the free display buffer.  The name is a hardcoded SIXBIT value.  That must be why NAMDRG checks the value of NF11TY.

It's an easy fix to patch in T09 instead.  But the long-term solution would be the make the .OPEN calls dynamic with respect to the TTY number.

larsbrinkhoff:
There was a problem with drawing a NAMPIC, but that seems to be fixed now.
namdrg

larsbrinkhoff:
I can't see any trace of who did it.  It seems ITS programmers are usually very selfless.

It would be an interesting and creepy exercise to feed programming styles to an algorithm to see who wrote what.  But then, I suppose many people touched the files over time, so it's probably futile.

eswenson1:
There are several things I really like about this code. First, the way the ITS squoze symbols are evaluated is elegant and parameterized. Second, the way parameters are passed on the stack for this logic.  Third, the fact that rather than having constants for the demon message entry sizes, stored in the source (or an include file), the size is retrieved from a dynamic variable in ITS.  Fourth, the code is generally better commented that a lot of ITS code, and uses lower-case assembly (for the most part).  I find uppercase ugly and hard to read.  I could go on, but generally, some of the most readable ITS code I've ever seen.

eswenson1:
I think this is rather interesting self-modifying code:

```
        rotc a,-3	;TTY # of console free display
	lsh a,3
	rotc a,3
	iorm a,@name1
	iorm a,@name1+2
```

where name1 is:

```
name1:	.open tyoc,[%tjmor+%tjctn+%tjdis+.uao,,'t00]	;char unit out, disp. mode, no continuation lns
	 .value
	.open tybc,[%tjmor+%tjctn+%tjdis+.bao,,'t00]
	 .value
```

I think it is doing this:   It is replacing the 't00 in the words referenced by the effective address of both the instructions at name1 and name1+2 with the tNN TTY number that it got from ITS at NF11TY.  And I guess what it is doing is looking at name1 and using the effective address in that .open tyoc, xxxx instruction (xxxx) and then updating the location at xxxx by OR'ing in the 'tNN.  

In other words, this is self-modifying code.  While this is kind of cool/elegant/efficient, I don't like the fact that is is modifying pure pages like this.  I suppose it doesn't matter, because there is only one NAMDRG daemon running on the system. Does anyone else find this interesting?

eswenson1:
Can someone explain this code to me:

```
.suset [.roption,,a]
	skipe debug
	 setz a,
	tlne a,optddt
	 .value [asciz/5/]	;manual start.
```

eswenson1:
Sorry 'bout that fumble, above.  I was going to post a comment asking to explain the code:

```
.suset [.roption,,a]
	skipe debug
	 setz a,
	tlne a,optddt
	 .value [asciz/5/]	;manual start.
```

and when I copied/pasted the code from the Github-displayed code, which only showed the ASCIZ string as "5" and not the $$5^K (which I'm assuming it really is), I noticed that the ASCIZ string was $$5^K, which I understand.  I had thought it was simply 5 and couldn't figure out what it was doing.  

In any case, when I pasted the code fragment, noticed what was really there, I attempted to abort posting my comment.  Unfortunately, I clicked on "Close and comment", which closed the PR.  So I opened it up again.  Again, I apologize for the stupidity.

larsbrinkhoff:
About the self-modifying code.  I added that.  You're right it's not good to write a pure page.  I can come up with something else.

(Related.  I tried to call INIT and then pdump the binary, but that didn't work so well.)

larsbrinkhoff:
By the way, we discussed how to disown demons, and this one clearly prefers $$5^K.

eswenson1:
Yes, I've been using $$5^K to disown daemons ever since our discussion.  In this case, Github just showed asciz/5/ and I was wondering what a .value [asciz/5/] did.  I didn't realize until I pasted the code that Github was just hiding the control characters.

larsbrinkhoff:
I reported the hidden control characters to GitHub support.

larsbrinkhoff:
I updated my change to use the OPEN .CALL instead of modifying the code: https://github.com/PDP-10/its/compare/c4dc5391ca67bfe000621a49dd44a4cb76e743d7..9f41b01597bccbecb75a483409532392d06aebf9  
Thanks Eric, I think this is better.

eswenson1:
Much better! Thanks.

-- Eric

PR #1097, opened 2018-07-26
ITSTER - Donald Fisk's block-stacking puzzle game

### LINKED ISSUE #424: TETRIS - game
[Donald Fisk]() has written a Tetris in Maclisp.

http://web.onetel.net.uk/~hibou/ITSter.txt  
SV: HIBOU; TETRIS 185

atsampson:
Imported from SV.

I've set up the ITSTER (INIT) file to match the other Lisp games Eric's recovered, but please let me know if there's a different way I should do this.

Fixes #424.


larsbrinkhoff:
I hope Eric can tell us more about this, but is it the case that the game menu runs the (INIT) files?  In that case, I think it would be nice if ITSTER was added to that menu.

If not, I think it would be better to dump out a TS program.  LIBDOC; OD (DUMP) shows how that can be done.

eswenson1:
Many of the lisp games run by NGAME are stored as FASL files with an (INIT) file to start the game up. Look in the NGAME source to see how to add a new entry to the main menu. You’ll see how the invocation of a menu item causes lisp to be loaded and passed the JCL to run the init file.

eswenson1:
And if you don't want to do the NGAME part, let me know -- assign a ticket to me -- and I'll do it.  I don't see why this PR shouldn't get merged as is.  It should work fine as @atsampson has done it. We can add to NGAME as a separate step and updated games.md when that is done.

atsampson:
OK, I'll merge as is (once the CI finishes on the rebased version). Cheers!

PR #1109, opened 2018-07-29
Tweak Makefile to allow overriding of shell commands

b4:
BSD's `touch` on OS X doesn't handle -d for dates. I needed to do `gtouch` from coreutils via Homebrew. Might as well make it possible to override everything else, too. Sometimes modifying PATH is an extreme chore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,16 @@
+# Some important environment variables
 EMULATOR ?= simh
+
+# Sometimes you _really_ need to use a different `touch` or `rm`.
+TOUCH ?= touch
+MKDIR ?= mkdir -p
+EXPECT ?= expect
+RM ?= rm
+LN ?= ln
+SED ?= sed
+TEST ?= test
+GIT ?= git
+CAT ?= cat
 
 include conf/network
 
@@ -27,12 +39,12 @@ RAM = bin/ks10/boot/ram.262
 NSALV = bin/ks10/boot/salv.rp06
 DSKDMP = bin/ks10/boot/dskdmp.rp06
 
-KLH10=${PWD}/tools/klh10/tmp/bld-ks-its/kn10-ks-its 
-SIMH=${PWD}/tools/simh/BIN/pdp10
-KA10=${PWD}/tools/sims/BIN/ka10
-ITSTAR=${PWD}/tools/itstar/itstar
-WRITETAPE=${PWD}/tools/tapeutils/tapewrite
-MAGFRM=${PWD}/tools/dasm/magfrm
+KLH10=$(CURDIR)/tools/klh10/tmp/bld-ks-its/kn10-ks-its 
+SIMH=$(CURDIR)/tools/simh/BIN/pdp10
+KA10=$(CURDIR)/tools/sims/BIN/ka10
+ITSTAR=$(CURDIR)/tools/itstar/itstar
+WRITETAPE=$(CURDIR)/tools/tapeutils/tapewrite
+MAGFRM=$(CURDIR)/tools/dasm/magfrm
 
 H3TEXT=$(shell cd build; ls h3text.*)
 SMF:=$(addprefix tools/,$(addsuffix /.gitignore,$(SUBMODULES)))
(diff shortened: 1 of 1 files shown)


ams:
But `-d` is a standard POSIX option to `touch`? And is supported on the BSDs.

b4:
Precisely why I am baffled by OS X's `touch` vomiting on it.

ams:
Should also use $() variable expansion...  

Trying to set`MAKE` is kinda redundant, it will be set by the called `make` program. This was required back in the day, but not anymore.

I think most of type of overrides are kinda ugly, and not really needed if one sticks to standard behavior, which so far all of this is.

b4:
Can you explain the differences between $() and ${}? I was having a bit of difficulty finding the differences.

Agreed they can definitely be cleaner, but does its intended job.

Also: https://opensource.apple.com/source/file_cmds/file_cmds-272/touch/touch.c.auto.html there's genuinely no `-d` in the `touch` bundled with OS X.

larsbrinkhoff:
@b4, I forcibly merged your commits.

larsbrinkhoff:
@ams, my reaction too was that $() is to be preferred, but then I looked it up and I didn't see why.

larsbrinkhoff:
Awating input from @ams.

TheFausap:
I thinks also overriding make could be useful, because sometimes in openbsd (if I remember well) I found a BSD make, not compatible with GNU make, but needed to compile the BSD kernel (for example).

ams:
$() is to be preferred since some old makes handled ${} in wonky ways (if I recall, ${} and $() could eval to different values depending on the shell environment).  

For `MAKE` this does not need to be set, since it is set implicitly by BSD make and GNU make, and as long as one always calls $(MAKE) things will be fine.  Then if you call BSD make it would use BSD make in all subsequent instances.

Makefile
```diff
@@ -41,120 +54,120 @@ OUT=out/$(EMULATOR)
...
-	../configure --bindir=${PWD}/build/klh10; \
-	make base-ks-its; \
-	make -C bld-ks-its install
+	../configure --bindir=$(PWD)/build/klh10; \
+	$(MAKE) base-ks-its; \
+	$(MAKE) -C bld-ks-its install
 
 $(SIMH):
-	cd tools/simh; make pdp10
+	cd tools/simh; $(MAKE) pdp10
```
atsampson: You could use `$(MAKE) -C tools/simh pdp10` here, as in the previous rule.

atsampson: You could use `$(MAKE) -C tools/simh pdp10` here, as above.

Makefile
```diff
@@ -1,5 +1,18 @@
...
+RM ?= rm
+LN ?= ln
+SED ?= sed
+TEST ?= test
+GIT ?= git
+CAT ?= cat
+
 include conf/network
 
 # The directores listed in SRC, DOC, and BIN are put on the sources tape.
```
atsampson: Use `$()` rather than `${}` in the definitions of KLH10, etc.?

Makefile
```diff
@@ -1,5 +1,18 @@
+# Some important environment variables
 EMULATOR ?= simh
 
+# Sometimes you _really_ need to use a different `touch` or `rm`.
+TOUCH ?= touch
+MKDIR ?= mkdir
```
atsampson: This is always used with `-p`. autoconf usually does this as `MKDIR_P ?= mkdir -p`, since then you can replace it with a script that does the same thing (like `mkinstalldirs`) if you have a non-POSIX `mkdir`.

Makefile
larsbrinkhoff: I wrote ${PWD} a long time ago since I thought that would use the environment variable PWD rather than a makefile variable.  @ams alluded to such a difference.  Maybe this is outside the scope of this pull request.

Makefile
larsbrinkhoff: Updated to include -p.  Removed -p from all users.

Makefile
larsbrinkhoff: Fixed all applicable places.

Makefile
atsampson: Probably; it was changed to `$(PWD)` in the rp0.dsk rule, though. In GNU make they mean the same thing - you could also use `$(CURDIR)` instead of the environment variable.

larsbrinkhoff: @aap also wanted $(CURDIR) for some reason, so I changed to that througout.

PR #1123, opened 2018-07-31
Update the markdown files in doc.

larsbrinkhoff:
Use \<kbd> and ◊.

### DIFF
--- a/doc/DDT.md
+++ b/doc/DDT.md
@@ -1,36 +1,45 @@
 ## DDT cheat sheet for Unix users
 
-To get access to a terminal, first type ^Z (Control-Z).  After that,
-you can log in.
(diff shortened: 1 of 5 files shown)


atsampson:
A few more for the Unix cheatsheet...

Unix command | DDT command | Colon command
--- | --- | ---
clear | <kbd>^L</kbd> | :clear
kill -9 123 | | :lock then 123gun
man 2 open | | :call open
dmesg | | :sysmsg
uptime | | :time

p<kbd>^K</kbd> is the stylish way to invoke PEEK.

Is it worth having a hint that ◊X can usually be typed as M-X?

larsbrinkhoff:
Thanks, I incorporated the suggestions.

PR #1110, opened 2018-08-01
GT40 bootstrap, and Lunar Lander

larsbrinkhoff:
I think it would be in the spirit of ITS to take the GT40 Lunar Lander game and convert it to PALX syntax.

@philbudne put files here:  
https://github.com/simh/simh/tree/master/PDP11/lunar11

I'm not sure if the provided .mac file or the binary is the best version.  The former could be edited to match the latter.


larsbrinkhoff:
Very small updates to assemble with PALX.

Minor observations.  What to do?

- Lunar Lander is actually titled Moonlander in the source code.
- The filename is GTLEM.MAC.
- The source code says LEM.090, so I called my version 91.
- ITS backups show there was a file GT40; LANDER BIN.

larsbrinkhoff:
Lunar Lander works when stuffed to SIMH.  STUFF rightly complained the unmodified file was trying to write to 0, which is how I knew I had to move it up.

I haven't tested the bootloader.

larsbrinkhoff:
All files come from @brouhaha, I believe.

PR #1132, opened 2018-08-02
Current makefile changes broke build on my machine

larsbrinkhoff:
The recent changes to the Makefile have broken the build on my Ubuntu laptop. My source tree is located at /home/eswenson/ITS/ws/its.  /home/eswenson/ITS is a symlink to /media/eswenson/ITS Volume/ITS, which is a mounted disk drive.  Note the space in the volume name "ITS Volume".  The build no longer works due to:

```
mkdir -p out/klh10
cd bin/ks10; /media/eswenson/ITS Volume/ITS/ws/its/tools/itstar/itstar -cf ../../out/klh10/minsys.tape _ sys
/bin/sh: 1: /media/eswenson/ITS: not found
Makefile:71: recipe for target 'out/klh10/minsys.tape' failed
make: *** [out/klh10/minsys.tape] Error 127
```

Looks like we need quotes around these pathnames in the Makefile.  And please don't tell me that spaces in filenames are evil. I know that. This is a volume that I created a long time ago.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,12 +39,12 @@ RAM = bin/ks10/boot/ram.262
 NSALV = bin/ks10/boot/salv.rp06
 DSKDMP = bin/ks10/boot/dskdmp.rp06
 
-KLH10=$(CURDIR)/tools/klh10/tmp/bld-ks-its/kn10-ks-its 
-SIMH=$(CURDIR)/tools/simh/BIN/pdp10
-KA10=$(CURDIR)/tools/sims/BIN/ka10
-ITSTAR=$(CURDIR)/tools/itstar/itstar
-WRITETAPE=$(CURDIR)/tools/tapeutils/tapewrite
-MAGFRM=$(CURDIR)/tools/dasm/magfrm
+KLH10=tools/klh10/tmp/bld-ks-its/kn10-ks-its
+SIMH=tools/simh/BIN/pdp10
+KA10=tools/sims/BIN/ka10
+ITSTAR=tools/itstar/itstar
+WRITETAPE=tools/tapeutils/tapewrite
+MAGFRM=tools/dasm/magfrm
 
 H3TEXT=$(shell cd build; ls h3text.*)
 SMF:=$(addprefix tools/,$(addsuffix /.gitignore,$(SUBMODULES)))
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
What I usually do in shell scripts: throw "s at it.  Better?

larsbrinkhoff:
Sorry, I shouldn't have pullrequesterized this one.

larsbrinkhoff:
Because now it looks like I wrote the orginal issue.  It's a GitHub bug or quirk.

atsampson:
That almost works, but it doesn't do what you want for the targets and dependencies that use $(KLH10) etc. - you need to quote spaces in target and dependency names, else it sees `name with spaces in` as a list of four separate targets...

```
Makefile:151: warning: overriding recipe for target '"/home/ats/src/name'
Makefile:141: warning: ignoring old recipe for target '"/home/ats/src/name'
Makefile:151: warning: overriding recipe for target 'with'
Makefile:141: warning: ignoring old recipe for target 'with'
Makefile:151: warning: overriding recipe for target 'spaces'
Makefile:141: warning: ignoring old recipe for target 'spaces'
```

Could it be written with all of the tools being relative paths? If we added a -C option to itstar (like tar and make have), the macdmp invocation is the only other thing that'd need changing...

atsampson:
Seems to be doable and it's maybe a bit simpler overall. Here's [itstar with -C support](https://gitlab.com/atsampson/itstar/tree/master) and [its with relative paths for the tools](https://github.com/PDP-10/its/tree/ats/spaces), which works at least with simh in a path with spaces in...

larsbrinkhoff:
I was going to suggest reverting back to ${PWD}, but the solution with -C seems nice.

I unarchived the GitHub itstar so you can update it.

atsampson:
OK. I've merged that change into itstar, and I've updated [ats/spaces](https://github.com/PDP-10/its/tree/ats/spaces) - I've confirmed that that version completes a full build on all three emulators in a path with spaces in.

larsbrinkhoff:
@atsampson, you can forcibly push that commit onto this branch if you like.  Drop mine first.

eswenson1:
Hmm.... after I merged this and pulled on my machine, build failed because GT40 directory wasn't there when we tried to run PALX on some source in that directory. I noticed that Makefile didn't include GT40 in src list.

Issue #1137, opened 2018-08-03
Jenkins KA10 build fails


larsbrinkhoff:
There's something strange about Jenkins' jobs and/or expect.

Output from other programs shows up in the log, but no output from expect except this error:

```
expect -f build/sims/build.tcl 192.168.1.100 192.168.0.45
spawn ./tools/sims/BIN/ka10 build/sims/init
error writing "stdout": bad file number
    while executing
"puts [llength $args]"
    (procedure "start_dskdmp" line 2)
    invoked from within
"start_dskdmp build/sims/boot"
    (procedure "start_dskdmp_its" line 2)
    invoked from within
"start_dskdmp_its"
    (file "/var/lib/jenkins/workspace/pdp-10-its-sims/build/mark.tcl" line 12)
    invoked from within
"source $build/mark.tcl"
    (file "build/build.tcl" line 111)
    invoked from within
"source build/build.tcl"
    (file "build/sims/build.tcl" line 58)
Makefile:65: recipe for target 'out/sims/rp03.2' failed
make: *** [out/sims/rp03.2] Error 1

atsampson:
"error writing 'stdout': bad file number" sounds like Expect's accidental close(1) problem.

PR #1138, opened 2018-08-03
Expect improvements

larsbrinkhoff:
- First and foremost, get 'expect' input from /dev/zero to fix missing build log on some CI services.
- Second, make build script progress more clear.

### DIFF
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -1,3 +1,5 @@
+log_progress "ENTERING BUILD SCRIPT: BASICS"
+
 expect "\n"; type "\033g"
 pdset
 
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -1,3 +1,12 @@
+proc log_progress {x} {
+    puts ""
+    puts "$x"
+    puts [exec date]
+    puts ""
+}
+
+log_progress "ENTERING MAIN BUILD SCRIPT"
+
 # If the environment variable BASICS is set to "yes", only build
 # the basics; ITS, tools, infastructure.
 if {![info exists env(BASICS)]} {
@@ -140,3 +149,7 @@ type "quit\r"
 
 shutdown
 quit_emulator
+
+puts ""
+puts "MAIN BUILD SCRIPT DONE"
+puts [exec date]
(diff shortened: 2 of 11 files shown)


larsbrinkhoff:
Pun intended.

larsbrinkhoff:
Fixes #1137.

larsbrinkhoff:
Seems like expect is running at 100% so maybe it's busy trying to read all that input from /dev/zero!

larsbrinkhoff:
/dev/null doesn't work.  Probably reads as EOF to expect.  Same result as before this pull request.

/dev/tty doesn't work, at least not at GitLab.  It said "no such device" or something.

mkfifo doesn't work.  Expect seems to hang indefinitely.

Now I'm trying to pipe in the output from a long 'sleep'.

atsampson:
Having just spent a while staring at strace output and the Expect code, it really does want stdin to be a tty - if it ever tries to read it and gets EOF, it'll close both stdin **and stdout**, meaning you won't see any log output. In fact, with the current upstream Expect source, it's [even worse than that](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=440818) - it does `close(0); close(1);` and then continues writing log messages to fd 1, so on my system it was writing the log output to `/dev/ptmx`!

On the other hand, it only seems to try reading from stdin when we do `expect_after timeout abort` -- presumably because at that point, we haven't spawned any child processes yet, so the spawn_id is pointing at stdin/stdout. Doing that after the spawn seems to work around the problem. [ats/expecthack](https://github.com/PDP-10/its/tree/ats/expecthack) has a couple of commits trying this out (for simh only at the moment)...

larsbrinkhoff:
Piping `sleep` to expect seems to almost work, except expect (hah) doesn't quit when it's done, even if I call `exit`!  So at this point I'm happy for any hacky workaround.

I saw something about using `fconfigure` to set stdin buffering or something.

larsbrinkhoff:
@atsampson, I folded your changes into this pull request and added sims and klh10 correspondingly.

larsbrinkhoff:
KLH10 didn't like the eof thing.

larsbrinkhoff:
Travis is ok.  GitLab is ok.

PR #1140, opened 2018-08-03
Check that ITS memory management handles TEN11 and 768K

larsbrinkhoff:
- Update the KA10 simulator with the 10-11 interface stub.
- Enable TEN11P in ITS.
- Increase core memory to the 786K maximum.


larsbrinkhoff:
Verified with a full KA10 build.

PR #1147, opened 2018-08-04
Revert to TRANSS 79 and build from source.

### LINKED ISSUE #1146: Build macsyma TRANSS from source
Currently, TRANSS is not being built from source, but a FASL is used.  Recently, TRANSS was built from source, but the resulting Macyma translator didn't work properly, so this was reverted.  This ticket is to once again build TRANSS from source. 

TRANSL; TRANSS 79 appears to work and result in translate_file correctly behaving.  Eventually, we need to figure out why versions 90 and 91 don't work.  MC backup tapes had version 80 as the one the last version of macsyma was built with, so we should attempt to find that version and use it, if not later versions.

eswenson1:
No longer include TRANSS FASL or later versions of TRANSS, since these break Macyma's
translate_file. Resolves #1146.


larsbrinkhoff:
The CircleCI test will pass once this is rebased on top of the lars/cicleci branch, or master once it has been merged.

larsbrinkhoff:
Rebased.

Issue #1134, opened 2018-08-03
master branch won't build for me


eswenson1:
I get the following when trying to build the master branch:

```*:cwd games
*:midas jotto
JOTTO HACK
JOTTO HACK
Constants area inclusive
From    To
54362   54633
Run time = 0.03
1621 Symbols including initial ones (59% used)

:KILL
*:job jotto
!
*:load jotto bin
*erase0$b   beg$g
$1B; ERASE0>>MOVE R,54456   utopen/   656470   $1'   dsk$
b/   0   $1'jotto$
c/   0   $1'dict$
beg7$g

6627.
IT IS NOW  5:53:18 PM EDT, THURSDAY, AUG 2,2018
words
$1B; ERASE0>>MOVE R,54456
The last command timed out.
Makefile:50: recipe for target 'out/klh10/rp0.dsk' failed
make: *** [out/klh10/rp0.dsk] Error 1
eswenson@eswenson-ThinkPad-W520:~/ITS/ws/its$```

It seems that darned "IT IS NOW..." message breaks the expect (ations).

eswenson1:
This has happened three times in a row!  Not entirely sure why this is reproduceable for me.  In any case, I'm trying with the ". words" changed to "words" misc.tcl -- looking for that "6627. words".  Seems for me, the "6627." is always on a line by itself, followed by the "IT IS NOW..." message, followed by the "words" output.

This expect matching stuff is SO BRITTLE.  Can't think of how many times the build has failed for us all because of this.

atsampson:
SYSJOB messages appearing in random places is kind of hard to work around. I guess we could add a flag to ITS to disable them, or we could do the build process against a TTY that isn't the console...

Issue #1100, opened 2018-07-27
DUMP speed differences between emulators


atsampson:
When working with tapes, I'd noticed that klh10 seems to DUMP much faster than it LOADs...

For a [current basics-only build on Travis](https://travis-ci.org/PDP-10/its/builds/408718282), here's how the time taken by the three emulators breaks down:

Emulator | Total | LOAD sources.tape | Build | DUMP output.tape
--- | --- | --- | --- | ---
simh KS | 9m34 | 0m22 | 5m45 | 1m02
klh10 KS | 21m01 | 5m46 | 8m21 | 1m41
sims KA | 31m40 | 7m31 | 7m03 | 7m44

So the (mostly CPU- or expect-input-bound) build stage takes about the same time on all the emulators - but why are the LOAD and DUMP times so different?

larsbrinkhoff:
I don't know about KLH10 LOAD times.

Does DUMP include the ICHECK pass?  If so, that may in part explain why SIMH takes five times longer to do that.

As for KA, it simulates writing individual words with a reasonably realistic delay between each.  I tried to add the option to skip this delay, but it turned out to be harder than I thought and I decided to leave it alone.

rcornwell:
For KA the speed of I/O operates much like the real KA10 in that data is read off the tape one character at a time. For SimH Bob reads/writes the whole block as a unit. I am not familiar with how KLH10 works. 

For the KA10 the TM10A was a polled type data transfer which means that bulk I/O is not possible. Also since the TM10A and TM10B are virtually identical other then one uses a DF10 and the other uses Polled mode. I did not feel it was good to separate the devices into two separate units. Other then Build/Dump time tape I/O is not used much so I did not see it as a problem.

Also as Lars said decreasing the delay time for TM10A caused problems with ITS messing itself up for some reason.

larsbrinkhoff:
I tried to speed up the TM10B to kind of do block transfers.

The issue with speeding up TM10A is that it's interrupt driven and ITS doesn't like those interrupts at a high rate.

atsampson:
Would there be any mileage in designing a "virtual tape" device that emulators and ITS could easily implement? (Or even a "host filesystem access" device...)

larsbrinkhoff:
I think MLDEV should mostly solve that, but only for ITS of course.  It goes in the other direction, "guest filesystem access".

What's missing for the Linux MLDEV to be really useful is to present files using the evacuate format and to mangle file names exactly like itstar does.

[...]

PR #1149, opened 2018-08-04
Silence the system job by patching ITS on boot

atsampson:
Avoid problems like #1134, where the system job has printed a message at an unhelpful time, by disabling its print routine with a boot-time patch. This also makes build logs shorter and more consistent, so they're easier to diff.

This needs some discussion before merging; in particular:

* Patching a random instruction into STYO+2 seems a bit ugly; we could add a label to patch, or even have a flag bit somewhere. But it has the advantage of working with unmodified original ITS when we're using it for bootstrapping...

* The sources.tape LOAD phase now doesn't produce any output until it completes. This may cause problems with tests - e.g. I think Travis gives up if it hasn't seen any new output for ten minutes, and at present LOAD is taking about 7 minutes on some builds (#1100).

### DIFF
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -1,6 +1,6 @@
 log_progress "ENTERING BUILD SCRIPT: BASICS"
 
-expect "\n"; type "\033g"
+patch_its_and_go
 pdset
 
 respond "*" ":login db\r"
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -37,8 +37,19 @@ proc respond { w r } {
     type $r
 }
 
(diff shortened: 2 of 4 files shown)


larsbrinkhoff:
This seems nice and unobtrusive!

1. The expect script could try the new label first, and fall back to STY+2.
2. LOAD could use the LIST option.

larsbrinkhoff:
I took the liberty of forcibly rebasing this to get a thumbs up from Circle CI.  Caveat programmator.

atsampson:
OK, let's go with this as is for now, then.

PR #1041, opened 2018-08-10
MacHack VI with 340 chess board

larsbrinkhoff:
RG; C 524  
MS; CHEG 104

Will not assemble as is.


larsbrinkhoff:
Outstanding problems:

```
Unterminated successful bracketed conditionals
The first was at 40-041 of file C
CHESS C
NF2LA2+1        74077    0.   280-005   NF2LA1  Undefined
NF2LA2+5        74103    0.   280-009   NF2LA4  Undefined
NF2LA2+10       74106    0.   280-012   N2FLA5  Undefined

**Comment by larsbrinkhoff** [comment]
The undefined symbols seem to be in a section of code that's not completed and not used elsewhere in the program.

**Comment by larsbrinkhoff** [comment]
I have been hunting for the cause of the `Unterminated successful bracketed conditionals` `The first was at 40-041 of file C` without success so far.  The indicated file position is a closing bracket.

MIDAS 324 says the exact same thing, but 135 says this:

```
AHPPV8+4        41406    0.   134-123   ) SEEN WHEN ] EXPECTED
CMPPNT+6        41635    0.   135-040   > SEEN WHEN ] EXPECTED
CMPP2+14        41677    0.   135-075   > SEEN WHEN ] EXPECTED

larsbrinkhoff:
MIDAS 135 and EMACS don't agree on the page number.  MIDAS page 134 is EMACS page 131.

I think MIDAS 135 doesn't handle .ASCII properly, so that's a dead end.

larsbrinkhoff:
Fixing all uses of .ASCII to add a space after the ! expression (as documented in the MIDAS manual) makes MIDAS 135 assemble C cleanly.  However, MIDAS 458 still complains about the unterminated conditional, so that's not it.

larsbrinkhoff:
MIDAS 289 also complains about the unterminated conditional.

larsbrinkhoff:
Ok, I officially give up!  There may be an unterminated conditional but if so, I don't think it's anywhere near page 40 as I commented out every conditional there and in the neighbouring pages.

Anyone else want to give it a try?

larsbrinkhoff:
I did not succeed.  There's still a... I'd say warning.  The message points out a problem, but it's benign.  Maybe the MIDAS manual says it best:

> If  the "]" of a successful bracketed conditional is missing, no harm results unless a literal was in progress.  However, it means that if in another assembly the same conditional failed (perhaps due to different parameter settings), lossage would occur.  Therefore, MIDAS prints a message at the end of the assembly stating that there was an unterminated successful conditional, and says where the first one was.

So we should be OK.  I'd fix it, but I can't for the life of me find which conditional it is.  It doesn't seem to be where MIDAS says it should be.

larsbrinkhoff:
I guess we'll make do with this.

larsbrinkhoff:
@IanOsgood, FYI MacHack VI sources here.

PR #1161, opened 2018-08-14
Some memos

larsbrinkhoff:
TJ6 sources for some memos.

- WP 78: FED
- AIM 240: 11SIM
- AIM 260A: LOCK
- AIM 261A: PEEK


larsbrinkhoff:
The increased influx of typeset documents raises the question: what should we do with them?  Just let them sit, or print PDFs or something?  I suggest maybe XGP-11 should be used when it comes online, unless that complicates the build process too much.

larsbrinkhoff:
In #1422, I did run BOLIO to get a typeset XPG file.  I added a new build script typeset.tcl for this.

larsbrinkhoff:
@atsampson, I checked the LOCK and PEEK files.  They don't seem to be made with BOLIO or R.  Searching some random commands online, e.g. `.every`, turns up PUB.

PR #1141, opened 2018-08-15
PDP-6 auxiliary processor

larsbrinkhoff:
Notes.  Eventually, a pull request.


larsbrinkhoff:
Branch is work in progress.  I made a pull request so people can see it more easily.

This seems to work now with a PDP6 job:

- Loading a program.
- Reading and writing individual words in DDT.

Things that seem intended not to work:

- Setting breakpoints.
- Starting.

larsbrinkhoff:
@cbf9427, did you ever use the AI lab PDP-6 from the PDP-10?  As in, for example, making a PDP6 job in HACTRN.

If so, do you remember which operations were available?  So far we have tested clearing PDP-6 core through the RESET system call, and loading a program.  DDT can read and write individual words.  Is there anything else that can be done?

larsbrinkhoff:
KLH confirms there is only memory access to the PDP-6 from ITS.  So no starting, stopping, breakpoints, stepping.

> Yes, the 10 only had memory access as far as I remember.  The most common use was loading Spacewar or one of the other fun programs to do stuff with the 340 display or the music speakers.  This was far, far easier and faster than using either paper tape or DECtape.

larsbrinkhoff:
KA10 simulator pull request in progress: https://github.com/rcornwell/sims/pull/93

larsbrinkhoff:
In particular, this conditional expression is hairy:

    IFE <1-PDP6P>\<<T11CPA_-18.>-<PDP6BM_-18.>>

What I mean to say is that PDP6P must be true, *and* also T11CPA and PDP6BM must be in the same moby.  Inverting this logic gives us: skip the following code if PDP6 is 0, *or* the addresses are in separate mobies.

larsbrinkhoff:
Lightly tested with @aap's pdp6 simulator.  Made a PDP6 job and read and wrote some words.

larsbrinkhoff:
```
TO START THE PDP-6:
    PRESS THE STOP SWITCH
    PRESS THE I/O RESET SWITCH
    SET THE CONTENTS SWITCHES TO ZEROES
    SET THE ADDRESS SWITCHES TO 40(OCTAL)
    DEPRESS DEPOSIT
    PRESS I/O RESET
    SET ADDRESS SWITCHES TO 41(OCTAL)
    DEPRESS DEPOSIT
    PRESS I/O RESET
    PRESS START
```
https://www.saildart.org/INFO.ITS%5BUP,DOC%5D

PR #1167, opened 2018-08-16
LOCK aliases: VTFIX, and start TV-11 & PDP-6

larsbrinkhoff:
TS VTFIX, unknown directory.

LOCK checks its JNAME when it starts.  If it's VTFIX, it'll call a special routine and exit.

(I'd like to add something like PDP6UP and TV11UP to make it easier to start those facilities.)


larsbrinkhoff:
I made it SYS3; TS VTFIX.

Added TVFIX and 6FIX.  They are not linked, but could be from CHANNA; RAKASH TVFIX and 6FIX.  This way, TARAKA will start them when ITS has booted.

Not tested yet, so not ready for merging.  But ready for comments!  To me this feels like an ITS-y way to do it.

larsbrinkhoff:
I made TVFIX call STUFF first, then set TT11P to 1.  Maybe that's overdoing it.

PR #1171, opened 2018-08-16
Assemble Tech II for TV display.

larsbrinkhoff:
Add AGB; C DIS which looks similar to CHPROG; OCDAGB.

This endowes Tech II with a chess board very similar to MacHack, but about 60% larger.


build/misc.tcl
```diff
@@ -390,8 +390,7 @@ respond "*" ":pdump games; ts trek\r"
 respond "*" ":kill\r"
 
 # Chess: timesharing, no TV display
```
atsampson: Update the comment?

larsbrinkhoff: Of course, thanks!

PR #1176, opened 2018-08-22
Port Tenth to MIDAS

larsbrinkhoff:
Tenth is @aap's toy Forth for KS10.  I ported it to MIDAS syntax.  Verified with SIMH.


larsbrinkhoff:
On second thought, I'll make a version 3 to also run under ITS.

larsbrinkhoff:
Left as an exercise: porting to KA10.

larsbrinkhoff:
This isn't a significant change, so maybe another round of reviews aren't needed.

PR #1178, opened 2018-08-23
Check Makefile SRC, DOC, and BIN.

larsbrinkhoff:
Run this with `make check-dirs`.  This is added to all CI builds.


larsbrinkhoff:
First commit is just the check and it fails as expected.

The second commit adds missing directories.

PR #971, opened 2018-08-23
Fix KS10 microcode

larsbrinkhoff:
Should be assembled from these files:

```
; Version 262:
; DSK: KSHACK; ITS 15           15:29:13 12-JUNE-86
; DSK: KSHACK; KS10 47          21:53:04 16-JAN-87
; DSK: KSHACK; SIMPLE 41        22:17:15 16-JAN-87
; DSK: KSHACK; FLT 5            21:53:03 16-JAN-87
; DSK: KSHACK; EXTEND 4         20:01:38 16-JAN-87
; DSK: KSHACK; INOUT 50         20:00:51 16-JAN-87
; DSK: KSHACK; ITSPAG 98        20:01:24 16-JAN-87
; DSK: KSHACK; PAGEF 10         02:03:53 8-OCT-84

### CODE REVIEW

**Comment by larsbrinkhoff** [comment]
I have been confused.  The file name in the 262 listing is SIMPLE 41.  But my reconstruction has a typo in it.

The backed up SIMPLE 42 is actually identical to the text from the listing.  And generates a binary identical to the one from MIT backups.

**Comment by larsbrinkhoff** [comment]
I did find a difference.  Version 42 adds a comment.

**Comment by larsbrinkhoff** [comment]
The file bin/ks10/_/ram.ram ends up in .; RAM RAM and is written to the front end file system.  It has now been tested on a real KS10, and it doesn't work.  It turns out that file isn't identical to the MIT backed up AI: KSHACK; RAM 262.

**Comment by larsbrinkhoff** [comment]
With this new RAM file, we get this from KSFEDR:

```
!write
Are you sure you want to scribble in the FE filesystem? yes
Which file? ram
Input from (Default DSK: DB; RAM RAM): .;ram ram
ERROR:  File too small.  Nothing was written.
```

I don't know what's wrong.  LF and $$^F say the file is exactly 6 blocks as it should be.

Being able to write the front end RAM file is crucial for installing on a real KS10.

larsbrinkhoff:
Having issues booting a real KS10.  It could be the microcode.  I'll update this pull request when it's working.

larsbrinkhoff:
The fresh microcode RAM file we build from source code is identical to AI: KSHACK; RAM 262.  Yet somehow KSFEDR think it's too small to go into the FE file system.

larsbrinkhoff:
My only idea right now is that it could be something about padding with ^C or NULs.

larsbrinkhoff:
The `ERROR:  File too small.  Nothing was written.` has disappeared.

On second thought, we might as well keep SIMPLE 42 since it only added a comment.

I updated the bin/ks10/_/ram.ram file and verified it's identical to the RAM 262 file in the 1990 "its-snapshot".

eswenson1:
I have no way to review this. You changed the ram file, but I have no clue what the changes are. Why are we doing this? What are we fixing? How can this be tested? Is there an associated ticket?

eswenson1:
Is this for ticket #1182? Is there a KS10 emulator that accepts and uses microcode? Or a real, working KS10? 

Please provide some background for this ticket.

larsbrinkhoff:
The changed ram file can't be reviewed, because it's a binary file.  There is no separate ticket, this pull request has the background and (rambling) discussion.  I tried to summarize in the commit comment:

> Update bootstrap version of KS10 microcode.
> 
> This file was assembled from sources, and is identical to the KSHACK;
> RAM 262 file.  It has been tested on a real KS10.  The older file
> doesn't work on a real KS10.

So the reason for the change is that the microcode ram file didn't work.  The new one works.

I considered rolling back the source file SIMPLE 42 to version 41, but since the only difference is a comment I changed my mind.

eswenson1:
So the RAM file now matches the source for it and before didn't? And the binary has been tested on a real KS10. That's good enough for me to approve. But what are the differences between the old binary RAM file and the new one? Did the old one correspond to an old source or was it corrupted? If it came from
An old source, do we have that source?

eswenson1 (approved):

larsbrinkhoff:
Sorry, I no longer remember where the previous ram file came from.  It's a very old commit; in fact the second one in this repository.  All I can say is I probably got it from one of the "1990" files.  At that point there was no KS10 to test it on.  Maybe I picked one at random and got a bad file, or maybe the file was good but mangled in some transfer.

We probably do have most versions for the microcode sources, but since it's a dozen files or so it could be slightly tricky to find the right versions for some particular microcode binary.  If we're lucky there may be a listing file that should have the versions.

PR #627, opened 2018-08-24
TIME, DATE, DATSRV fixed when .RDATE returns -1

larsbrinkhoff:
When time and date hasn't been set, DATE displays something like

    MONDAY, _______!MR3(          !)     /     __, 20__  __:__:__ AM


larsbrinkhoff:
Fixed by checking .RDATE return value for -1.

The ^V ^W and Altmodes doesn't show in GitHub.

larsbrinkhoff:
Now I think I should increase the scope to include all MIDAS programs calling .RDATE.

larsbrinkhoff:
Maybe not *all*, but a few more.

Issue #641, opened 2018-01-26
Add VT100 support to ITS


larsbrinkhoff:
If VT100 support were added, there would be no need to run CRTSTY.

It has been suggested that AAA (which is supported) uses almost the same escape sequenses as a VT100.

larsbrinkhoff:
I tried AAA with xterm, but it wasn't a good experience.

atsampson:
I had a quick look at putting together a pull request based on @mattwyrm's patch. I'm thinking that it'd make sense to add two new terminal types:

- VT100, changing the cursor location sequences as above, and also adjusting the other AAA-ish control sequences for compatibility with a real VT100 (including padding);

- ANSI, which is like VT100 but with no padding, for use with terminal emulators.

larsbrinkhoff:
Sounds good to me!

larsbrinkhoff:
@mattwyrm and @atsampson,

I have been using TCTYP AAA for a while now, and it seems to work fine.  I don't know what problems I thought I had when I wrote the above comment.

Similar to CRTSTY, maybe VT100 should be a fixed screen size, and ANSI should request to set width and height.

larsbrinkhoff:
Also, change from AAA 48 lines to ... 24 or 25 lines on a VT100?  This is actually quite nice when the character rate is slow and EMACS decides to redraw the full screen.

larsbrinkhoff:
Unpatched AAA works quite well with xterm, but not cool-retro-term (uses QTerminal emulator) or PuTTY.

larsbrinkhoff:
@mattwyrm @atsampson, in order to have a good user experience with a wide range of terminal emulators I suggest adding a VT100 terminal type using a safe subset of control codes.  This would then be the default type for most lines.

larsbrinkhoff:
I expect the @PiDP -10 to draw more attention to ITS, so I want to smooth out most potential speed bumps.  A garbled terminal display will not impress users.

mattwyrm:
Sadly I do not have the time currently to really work on this, but...
Don't expect unpatched AAA to work. If it does, it's luck.
The bummer with my patch is that you always use ESC [y;xH to position, which is slightly longer than the relative code - but that should not matter on lines above 4800 baud.
What I can do is going over the AAA codes in TS3TTY and tell you if they're save to use. That is, if they're reasonably common. I have an idea about doing relative cursor motion, but don't count me in on that, it really depends how much time I can fight free.

[...]

PR #1190, opened 2018-08-30
Use TK10, Knight kludge TTY scanner

larsbrinkhoff:
Pull request preview.

Adds to KA ITS 16 terminal ports on TELNET port 10000.

They are set to terminal type AAA since this is close to VT100 (see #641).  The very first port is set to be "close to 340 display".


larsbrinkhoff:
You can set `speed=X` in the SIMH `attach` command.  It's not set in the default configuration, which means I/O will not be throttled.  There's a polling interval set to 1 ms so there's still an upper speed limit.

larsbrinkhoff:
I just ran PEEK with TK10P==1, and it crashed.  Need to check this before merging the pull request!

larsbrinkhoff:
Waiting for @rcornwell to merge the TK10 device into his KA10 simulator.

larsbrinkhoff:
It seems this device may have only supported speeds up to 300 baud.

larsbrinkhoff:
Added new ports to the comment above: DZ11 and GT40.

larsbrinkhoff:
Rebased, and brought in the latest version of the KA10 simulator.  Now ready for merging.

larsbrinkhoff:
Over five years later, allocating port 10001 for GE consoles now turns out to be useful.

PR #1189, opened 2018-08-31
Login messages

larsbrinkhoff:
TELSER prints out SYS: NET MAIL when a user connects.  This only shows briefly.  As soon as DDT starts, it will clear the screen and print SYS: SYSTEM MAIL.  (There's also SYS: NETLSR MAIL.  I don't know when it's used.)

Helpful messages should be entered in these files.

### DIFF
--- a/src/sys/net.mail
+++ b/src/sys/net.mail
@@ -1,0 +1,1 @@
+It's a lovely day to be a turist!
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
I added some useful information to SYS; SYSTEM MAIL.  Added a less useful message to SYS; NET MAIL.

Suggestions welcome!

PR #1121, opened 2018-09-01
Update DDT to use TV meta bucky bit

larsbrinkhoff:
When logged in from a TV, typing to e.g. DDT will ignore the Meta bucky bit.  Typing <kbd>Meta</kbd>+<kbd>U</kbd> prints just U, which is a disappointment.

Also, ^Z is no longer a system-level attention characters, but a plain control character.  That's jarring, but maybe we can get used to that.

### DIFF
--- a/src/sysen1/ddt.1547
+++ b/src/sysen1/ddt.1547
@@ -6966,9 +6966,20 @@ TYI:	CALL TYOFRC	;.LISTEN IS SUPPOSED TO FORCE OUTPUT.
 TYI3:	CALL HAKKAM	;DO HAKKAH RQ'S, SAY DO ANY MORE WHEN QUEUED.
 	SETZM CTLZFL
 	SKIPGE D,TTYUNR
-	  .IOT TYIC,D
+	  .IOT TYIFC,D	;READ FULL CHAR SET INPUT CHANNEL
+	SKIPE METAP
+	 JRST [TRZ D,%TXTOP+%TXSUP	;CLEAR TOP AND SUPER
+	       TRZE D,%TXCTL	;IF CONTROL BIT SET,
+	        TRZ D,140	;CONVERT TO ASCII.
+	       TRZN D,%TXMTA
+		JRST .+1
+	       MOVEM D,TTYUNR	;META BIT SET, SAVE CHAR FOR LATER.
+	       MOVEI D,33	;RETURN ALTMODE.
+	       JRST TYI4]
 	SETOM TTYUNR
-	setzm hakok	;defer HAKKAH rq's again.
+TYI4:	setzm hakok	;defer HAKKAH rq's again.
+	skipe echop
+	 pushj p,echo	;Echo character.
 	skipn tthelp		   ;If [HELP] isn't desired
 	  caie d,%txtop+"H	   ;  help key becomes question mark
 	    caia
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
This is now working for input.

However, typing Meta-X doesn't echo as ◊X.  This is because echoing is done by ITS, not DDT, and apparently ITS echoes characters stripped off bucky bits.

larsbrinkhoff:
This raises the question, how *should* ITS echo meta characters?  I believe it %TOFCO isn't set, it'll just strip off everything but %TXASC and print that.  So that's what we get now.

But if %TOFCO is set, ITS uses the SAIL alpha/beta/epsilon for control/meta/super.

... though I didn't consider control characters.  I do get ↑ for control characters.  So it would be nice to get ◊ for meta.  Or maybe that's only for Altmode...   EMACS handles Altmode as a Meta prefix, but that's not how it works in general.

larsbrinkhoff:
Setting %TSFCO works, but then control characters are output with an alpha prefix!  I'm not sure I'm willing to go that route.

Sample session with what would render as ^R, $J, and $$V on a plain ASCII terminal:

```
*αr sys; net mail
Welcome turist!
*fooβj
*◊βv
* FOO - 11
*

larsbrinkhoff:
I don't think it would be OK to change this in ITS.  But it might be OK to change DDT to do its own echoing and use ◊ to indicate Meta as well as Altmode.  Since my change is to make them work the same for input.

larsbrinkhoff:
As you can see, I opted to make DDT do it's own echoing, though only when %TOFCI is in effect.  So using ordinary 7-bit ASCII terminals should work the same as before.

Now, when DDT reads a character with the meta bit set, it will echo an Altmode first.  This displays as ◊ on a TV.

larsbrinkhoff:
@eswenson1, @atsampson, can you please check this?

eswenson1:
I was waiting for atsampson to review first. I’m not confident that my doubts are legitimate, but it seems wrong to do these fixes in DDT, rather than ITS. Can you rationalize why you did these here?

larsbrinkhoff:
Most programs probably ignore the meta (etc) bit.  I haven't checked, but from testing with DDT I believe this is the case.  The primary meta-using programs that spring to mind are EMACS and SUDS.

When I considered this I didn't see any pressing use case for making everything work with a meta key.  It's not clear if programs should inspect the meta bit, or convert it to an Altmode prefix.

It would be possible for applications to tell ITS they want the meta to Atlmode  prefix conversion.  If we see many programs would benefit from this, sure.  In that case I can update DDT to use that.  For now, I think DDT is the only program I want to fix use the meta key.  I don't want to add system-wide changes just in case.

I haven't considered Lisp (or, say, Muddle).  I imagine sophisticated applications would want to see the meta bit untouched.  Others can be fine ignoring meta (like before), and some could use an Altmode prefix.  I suppose Macsyma is *the* Lisp application; what would you have it do?

larsbrinkhoff:
Though of another program: SUDDUP.  It should probably pass through the bucky bits untouched.  I haven't checked what it does.

eswenson1:
Yes, but if the "right" way to make it possible for arbitrary programs (including DDT) to see/display and process characters with these bits enabled is to change ITS (so that all programs benefit), then I don't think changing DDT is the right thing to do.  It seems a bit of a kludge.   Do you disagree with that?  And if so, please explain your thinking. I'm happy to be convinced that these changes to DDT are "good" -- I just don't feel comfortable (yet) with this.

eswenson1:
Rereading this thread, it seems like you are saying that ITS already does the "right" thing with these bucky (etc) bits.  That they get passed through to programs (including DDT) and it is up to the program that gets these characters whether they can display, act on, understand, etc. these characters.  If that is the case, and you are saying that DDT currently fumbles these characters, then I would agree that "fixing" DDT is the right approach.  Please confirm this is your position.

larsbrinkhoff:
Yes, that's what I'm saying.  Applications request the bucky bits by opening a TTY channel with the %TIFUL control bit set.  DDT already does this, but doesn't use that channel for general input.

larsbrinkhoff:
> That they get passed through to programs (including DDT) and it is up to the program that gets these characters whether they can display, act on, understand, etc.

Well to be more precise, programs must opt in to this.  Not using %TIFUL amounts to saying, "I ignore those bits, please throw them away for me".  This is kind of the default since a vast majority of programs leave %TIFUL off.

PR #18, opened 2018-09-02
Need INFO program

larsbrinkhoff:
Now that we have  INFO and .INFO. directories, we should include the INFO program, which knows how to locate and process the .info files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ SRC = system syseng sysen1 sysen2 sysen3 sysnet kshack dragon channa	\
       tensor transl wgd zz graphs lmlib pratt quux scheme gsb ejs mudsys \
       draw wl taa tj6 budd sharem ucode rvb kldcp math as imsrc gls demo \
       macsym lmcons dmcg hack hibou agb gt40 rug maeda ms kle aap common \
-      fonts zork 11logo kmp
+      fonts zork 11logo kmp info
 DOC = info _info_ sysdoc sysnet syshst kshack _teco_ emacs emacs1 c kcc \
       chprog sail draw wl pc tj6 share _glpr_ _xgpr_ inquir mudman system \
       xfont maxout ucode moon acount alan channa fonts games graphs humor \
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -18,6 +18,11 @@ type ":vk\r"
 respond "*" ":print device;..new. (udir)\r"
 type ":vk\r"
 
+# TCTYP
+respond "*" ":midas sysbin;_syseng;tctyp\r"
+expect ":KILL"
+respond "*" ":link sys;ts tctyp,sysbin;tctyp bin\r"
+
 source $build/emacs.tcl
 
 # TAGS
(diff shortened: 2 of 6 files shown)


larsbrinkhoff:
Checks out on my KA ITS too.  Doing a full build locally to verify.

The TSINFO version on backups is 62.  Should we write version 63?

eswenson1:
I was just about to commit.  I have it rename to EMACS; TSINFO 96 since the version of EMACS1; INFO 96 is 96.  Maybe that's wrong. I'll wait to hear from you.  I went to assign this issue to myself and see it is already assigned to you.  I'll hold off.

larsbrinkhoff:
Did you do a full build from scratch?  I did, and INFO doesn't quite work.  When started, it doesn't update display until you press a key.

I used SIMH.  I'm testing KLH10 now.

larsbrinkhoff:
I've had this assigned for 1.5 years or so!

There is one thing I'm wondering about.  The older file had a macro for ..L.  Maybe it's something we want.

eswenson1:
I'm doing a full build now. But when I just did the INFO build steps above, it displayed the intro screen without my having to type anything. Will let you know after my klh10 full build. As to ..L., I don't know. I'll have a look.

larsbrinkhoff:
Right, I did the steps on both a finished KA10 and KS10 ITS, and INFO worked fine.

But I have now checked all three emulators, and they all behave the same: starting INFO displays nothing until the first key press.

eswenson1:
Yes, I can confirm that a full build results in an INFO that doesn't display anything until a key is pressed. I wonder why?

larsbrinkhoff:
Thanks.  I too have noticed that about the EMACS mode line.  It never bothered me so I didn't look into it, but now that you mention it, it could be related to how the fresh INFO doesn't display anything. 

I tested adding the ..L macro.  Will test it today.

Also, building INFO on a finished ITS build works.  So maybe build it late in the build script?  Ugly fix, I know!

eswenson1:
Yes, it is really strange that it doesn't work when built in emacs.tcl (even at the end).  We should try to put it near the end and then use a binary search to figure out what dependency makes it work.

larsbrinkhoff:
The ..L macro makes no difference with regards to the display.

larsbrinkhoff:
Very strange.  I put building INFO almost last in build.tcl.  Still same thing.

But if I rebuild it, it works.

larsbrinkhoff:
SRCCOM /$ show some differences.  If I rebuild yet again, the two binaries generated after rebooting are identical.

larsbrinkhoff:
Now testing a reboot in the built script right before generating TSINFO.

larsbrinkhoff:
I wonder if something could be cached in memory which needs to be flushed?

larsbrinkhoff:
Not even putting it last and a reboot right before seems to work?!?

eswenson1:
What about building it twice at the end of emacs.tcl?

larsbrinkhoff:
Hmm!

If I start ITS and generate a new TSINFO right after ^Z, like the expect script, INFO doesn't work.

However, if I first :TCTYP AAA and then generate a new TSINFO, it does work.

larsbrinkhoff:
Success!  If terminal type is set to AAA, the generated INFO works.  Looks like some of the terminal settings spills over to the dumped file.

larsbrinkhoff:
Yes, it's a hack.  It's not very surprising though.  I suppose Dump Environment uses a TECO function to write out its core image in a purified form.  So probably some internal state comes along for the ride.  I guess if TECO is started from a printing terminal, some things will behave differently.

Waiting for favourable reviews to merge this.

eswenson1:
Well, it was worth a try. Could be this was just how it worked.

larsbrinkhoff:
That's a fair point.  I moved TCTYP.

PR #629, opened 2018-07-29
URUG - GT40 Microrug

larsbrinkhoff:
SYSEN2; URUG 27  

PDP-11 debugger for GT40, written in PALX assembly language.

Part of the rug/carpet family of debuggers.

PR #1206, opened 2018-09-07
Attach GT40

larsbrinkhoff:
Set up SIMH to emulate a GT40.

- PDP-11/05 processor.
- VT11 display processor.
- Keyboard through DL11.
- Host interface through one more DL11.
- Boot ROM?

Attach it to a PDP-10 serial line.

- Traditionally T40 on AI.  Some software assumes this.

Software:

- Boot ROM: GT40; BOOTVT S09 (#1110).
- URUG debugger (#629).
- Lisp display slave and Datapoint emulator: VT07 (#942).
- CONS console: CCONS (#1076).

For CCONS, it was attached to 10-11 Unibus number 2.


larsbrinkhoff:
This should now work.  Running the `start` script automatically starts a GT40 simulation in the background and connects it to a PDP-10 serial line.  Since the KLH10 doesn't have a working DZ11, it's left in the cold.

This isn't ready to go in, because I'd like the [PDP-11 ROM device](https://github.com/simh/simh/pull/614) to be finished first.  The reason for that is that the ITS build assembles the VT11 boot ROM, and we use use the fresh binary in the GT40.

@markpizz, this it a turn-key solution to run a GT40 attached to a PDP-10 host.

larsbrinkhoff:
CircleCI gets the wrong commit from tools/dasm for some reason.  Travis CI gets it right.

larsbrinkhoff:
Thanks @atsampson, good points.

I changed the `start` script to it accepts arguments.  Any argument passed in is just executed as a shell function.  The intent is that there should be functions for all additional machines, but for now there'll just be `gt40`.

larsbrinkhoff:
Waiting for SIMH to have a bootable ROM device: simh/simh#614.

larsbrinkhoff:
Merging this, and fingers crossed, CircleCI will get the submodule commit ID right.  Or else I'll back out the commits.

larsbrinkhoff:
Looks like it worked.

However, another problem popped up: #1219.

PR #1218, opened 2018-09-10
Fix KA10 build script MARK responses.

larsbrinkhoff:
SALV prints `PACK # =` for information, not request for input.


larsbrinkhoff:
I added some more build script tweaks.

PR #1187, opened 2018-09-10
KA ITS pack #2 almost full

larsbrinkhoff:
I'm now getting occasional KA ITS build errors due to the disk being full.  The cause seems to be that RP03 pack # 2 has very little free space.  ITS is distributing files over the other packs, but apparently not staying away from # 2.

Maybe it's because # 2 has all source files on it.


larsbrinkhoff:
Now I got the disk full error building this branch, on the very first invokation of MIDAS.  So my ugly eyesore fix still doesn't work 100%.

```
*:login db
ML: DB; DB MAIL - NON-EXISTENT DIRECTORY
:dump
DUMP  .444
_reload THIS WILL SMASH YOUR DIRECTORY.
 ARE YOU SURE ? y
links crdir sorry
FILE=*;* *
TAPE NO      1 CREATION DATE  180911
 E-O-T
_quit
:KILL
*:print sysbin;..new. (udir)
DSK: SYSBIN; ..NEW. (UDIR) - FILE NOT FOUND
:vk
*:midas
sysbin;_midas;midas
MIDAS
Pure pages = 13
ST = 12537
1755 words initialization coding.
Wasted gap pages (MINPUR-MAXMAC) = 7
MIDAS
IOCERROR: DSK: SYSBIN; _MIDAS OUTPUT - DEVICE FULL
420613>>XCT 1

larsbrinkhoff:
It certainly is, but it was the best I could do on short notice.  And it was getting urgent.

eswenson1:
Why not add another pack? Why are files not more evenly distributed as they are loaded from tape? Is pack 1 also almost full?

larsbrinkhoff:
I did add two more packs, but that didn't help much.

When the files are loaded from tape, there are only two packs: number 2 and 3.  This is when we boot from the binary ITS.  Still, the files would comfortably fit if they were evenly distributed.  But apparently pack 2 is preferred.  I don't know why.  Fixing that may be the best long-term solution.

When the new ITS is assembled from source, it reboots using four packs.

rcornwell:
How about splitting the load into two source tapes. First tape has basic core ITS system no docs. Second set has all the docs and extra things like Lisp, Macsyma etc. First load the very basic system, then load base source rebuild base tools, then reboot and load rest of system. This might also solve the load time problem.

eswenson1:
I think we should solve the real problem -- when the tape is loaded, the files should get evenly distributed across all available packs.  We should look at DUMP to see what algorithm it is using -- if it is doing anything, and if not, look at ITS to see how it picks a pack.

eswenson1:
Why are there only 2 packs when you load the files from source?  Could we not use an ITS configured for 4 packs for the initial boot (as a workaround until we understand and fix DUMP/ITS to allocate files on packs better)?

larsbrinkhoff:
The sources tape split is implemented in #1238.

larsbrinkhoff:
Apparently assembling ITS with QRSRVP enabled is the cause of the disk pack being filled up.  See #2049 

I'm considering disabling QRSRVP unless we can come up with a good reason to keep it.  We still don't do the GFR thing.

PR #1221, opened 2018-09-13
Newest MIDAS with device codes

larsbrinkhoff:
Install the newest version of MIDAS that has device codes built in.

Use it to assemble old programs, without the /T trick.


larsbrinkhoff:
SYSENG; PT 2 assembles differently.  MIDAS 458 complains about MAP.  Maybe actually MIDAS 324 does the right thing here.

All other programs assemble identically according to SRCCOM /$.

larsbrinkhoff:
SYSTEM GEN low level formats a DECtape and writes a boot block (?) suitable for the MIT shadow loader.  See #1030.

PR #1226, opened 2018-09-16
KLH10 console lights

larsbrinkhoff:
This enables use of the Panda Display console lights with the KLH10 emulator.

- Compiling KLH10 with `--enable-lights` uses libusb to talk to the display.
- In KLH10, the command `lights usb` looks for the device on the USB bus.

If the device isn't there, it's just ignored.  However, this will force everyone to install the libusb library to build KLH10, and that's not good.  Any suggestions how to avoid that?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,7 +171,7 @@ $(KLH10):
 	$(MKDIR) tmp; \
 	cd tmp; \
 	export CONFFLAGS_USR=-DKLH10_DEV_DPTM03=0; \
-	../configure --bindir="$(CURDIR)/build/klh10"; \
+	../configure --enable-lights --bindir="$(CURDIR)/build/klh10"; \
 	$(MAKE) -C bld-ks-its; \
 	$(MAKE) -C bld-ks-its install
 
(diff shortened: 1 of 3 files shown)


eswenson1:
Thanks. I don't suppose you can buy one pre-made anywhere?  It looks complicated enough that I'm sure I'd have a miserable time building one.

larsbrinkhoff:
@DavidGriffith said he maybe will make more.

eswenson1:
@DavidGriffith If you can and will, I'd like one!  Will you? :-)

larsbrinkhoff:
I suppose the Panda Display is much like the KL-DUGE board on MC!

DavidGriffith:
Indeed I will make more.  One thing that bugged me about the original Spare Time Gizmos design is that to neatly mount it, one is expected to carve up a blank 5.25" drive bay filler panel, then print and apply the overlay.  I want to come up with a 3D-printable bracket that will accept a Panda Display panel from the rear, accept a nicely produced hard vinyl overlay on the front, and is able to be screwed into a standard drive bay.  So far, my tinkerings with FreeCAD suggest that I will need to move some things around on the Panda Display board to accomplish these design requirements.  I don't know when I'll be done with the bracket design.

larsbrinkhoff:
Ok, I'm merging this as is.  If someone complains about having to install libusb, I can make an update.

Issue #1223, opened 2018-09-14
CROSS - microcomputer cross assembler


larsbrinkhoff:
SYS1; TS CROSS

This is a pdumped DECUUO binary.   Not sure where the source code is, but it's not SYSENG; CROSS 38.

larsbrinkhoff:
I want this for assembling KNS10, the KS10 8080 front end code.

larsbrinkhoff:
Probably the one in DECUS:  
http://pdp-10.trailing-edge.com/cgi-bin/searchbyname?name=cross.mac

It's also in the Panda distribution.

larsbrinkhoff:
The Panda and SRI-NIC version is a little newer than the DECUS 10/20 version.

PR #1228, opened 2018-09-17
KNS10 - KS10 console software

larsbrinkhoff:
KSHACK; KNS10 4.2

This is an assembler listing.  But it includes the file names, so it an be used to restruct the original files.

We need CROSS (#1223) to assemble it.


larsbrinkhoff:
Notes for the future?

- I also got a binary image of version 5.2.
- And there's a listing for 6.2 on Bitsavers.

PR #1234, opened 2018-09-20
Assemble TARAKA (ATSIGN DRAGON) from original source code.

larsbrinkhoff:
CHANNA; TARAKA 14 is the original source code for CHANNA; ATSIGN TARAKA, which is linked from SYS; ATSIGN DRAGON.  Verified the binary to be identical.


larsbrinkhoff:
@eswenson1, do you want to add in any of your comments from your reconstructed source?

eswenson1:
@larsbrinkhoff I don't think it is necessary, do you?

larsbrinkhoff:
Not really.  The original has three comments, that's plenty. ;-)

PR #1235, opened 2018-09-20
MACRO-10

larsbrinkhoff:
- Adds binary files for LINK-10 (no attempt to build from sources).
- Adds historical binary file DECSYS; MACRO SHR from 1976.
- Adds DEC version of MACRO-10 from 1979.  This is a sweet spot: it can be assembled with the 1976 version, and it can assemble CROSS.
- Assembles this version using the old MACRO SHR.
- Deletes MACRO SHR, because it's no longer needed except for bootstrapping.
- Assembles CROSS from source using the new MACRO-10.
- Removes the binary TS CROSS from the repository.


larsbrinkhoff:
@rcornwell or other TOPS-10 user, I get an error "Non-existent device REL:" first time I try to run LINK, but the second time I try it works.  Am I doing something wrong?  I don't know why it thinks I want the REL: device.

```
*:dec sys:link
*dsk:macro/go
%LNKNED Non-existent device REL:
[       Please retype the incorrect parts of the file specification]
*macro/go
```

I tried adding DSK: but it didn't help:

```
*:dec sys:link
*dsk:macro/go
%LNKNED Non-existent device REL:

rcornwell:
REL: is typically defined as [2,5] I believe. It is where the .rel files live. You should be able to remove this check by which options you use to build link.

PR #632, opened 2018-09-21
RUG - PDP-11 debugger

larsbrinkhoff:
AR7: PDP11; RUG 526  
PDP11; RUG 526, original location  

Part of the rug/carpet family of PDP-11 debuggers.


larsbrinkhoff:
Now all .INSRT files are in place.  There's a complaint if you select machine "1", but other than that all assembly-time switches seem to work.

larsbrinkhoff:
Sorry, I forgot.  There should be some documentation files too.

larsbrinkhoff:
Hello @cbf9427,

I see you worked on the PDP-11 debugger called RUG.  Do you remember in which scenarios it was used?  Which computers were debugged with RUG?  How was it moved from ITS onto the target computer?  How was LODER used?

PR #1238, opened 2018-09-22
Split sources tape.

larsbrinkhoff:
KA10 ITS keeps running out of space.

I'll split off the essential sources for bootstrapping a new ITS:

- ITS
- DDT
- MIDAS
- SALV, NSALV
- KSFEDR
- DUMP

I was hoping these could go onto the minsys tape.  That works fine with SIMH (KS and KA), but not KLH10.  It seems some interaction between KLH10 and NSALV doesn't work right.


larsbrinkhoff:
The CI services seem to think so!

larsbrinkhoff:
After this, there's a lot of free space on the previously almost full # 2.  But now # 0 is somewhat short on space, and # 3 is almost unused.

```
FREE BLOCKS #2=2324 #3=8395 #0=714 #1=4343

PR #1240, opened 2018-09-23
DSKDMP 215 versus 216

larsbrinkhoff:
I just noticed that the DSKDMP I have on ES was built from SYSTEM; DSKDMP 216 and the one in DB was built from SYSTEM; DSKDMP 215. I did a SRCCOM on the two versions and there are several differences -- including some comments by KLH regarding running under KLH10.  Are we using version 215 for a specific reason?

### DIFF
--- a/src/system/dskdmp.216
+++ b/src/system/dskdmp.216
@@ -1276,14 +1276,11 @@ LOADG:	JSP TT,WRCB		;MAKE SURE ALL LOADED CRUFT IN PSEUDO-CORE IS OUT
 	TRNE CMD,10
 	 JRST READ		;NON-GOING COMMAND
 GO:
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
The KLH10 build fails intermittently.  Maybe this is why.

eswenson1:
Yeah I noticed that. Must be a bug in the mobile interface. I’ve seen this before. I select the approve radio button and press submit and it tells me a comment is required. It MAY be that I pressed submit when the default radio button (comment) is selected and then getting the error message, change radio buttons to Approve, but cookies prevent this change from ever “taking”.

larsbrinkhoff:
It's a good thing we can side-step the rules occasionally.

PR #1083, opened 2018-09-27
PCLSRing

larsbrinkhoff:
BAWDEN; PCLSR 201  *"PCLSRing: Keeping Process State Modular"*  
SYSDOC; PCLSR TEST


larsbrinkhoff:
It's hard for me to tell apart documents written in TJ6, Scribe, PUB, Runoff, R, etc.

If this is TJ6, we could.  There are also some other documents, e.g. in WP, that could be built.  It would be more interesting to build these when there is a simulation of the XGP printer.

larsbrinkhoff:
Let's get all the TJ6 documents built when the XPG simulation is ready.

eswenson1:
Can we produce .txt files (with pagination) with TJ6 now (prior to XGP)?  Or does it really require XGP?

PR #393, opened 2018-09-27
WHOIML - Imlac wholine

larsbrinkhoff:
SYS2; TS WHOIML  
SYSEN1; WHOIML  151


larsbrinkhoff:
Typical output when not run on an Imlac terminal:

```
:$Use :WHOLIN$KILL

larsbrinkhoff:
It is self-purifying like NAME.  The breakpoint is just to halt the program right after it has PDUMPed itself.

eswenson1:
Gotcha. Thanks.

PR #567, opened 2018-09-27
CTN - network CRTSTY

larsbrinkhoff:
By enabling NET, you get a networking version of CRTSTY, which is called SYS3; TS CTN.


eswenson1:
What does it do when NET is enabled? Allow you to have CRSTTY on a remote system that you are telnetted/supduped to?

larsbrinkhoff:
Actually I, don't know what it does.  A comment near the top of the source file makes it quite clear CRTSTY was intended to be assembled in this alternative fashion.  I didn't read in detail what it does, or why.

PR #548, opened 2018-10-01
JEDGAR - counterspy program

larsbrinkhoff:
Binary missing, source missing.


larsbrinkhoff:
Note that the source for JEDGAR shows scar tissues.  There's a jump to the PATCH area, and that area itself looks like it's binary patched.

larsbrinkhoff:
I left out comments on purpuse, to retain some air of mystery.

src/moon/jedgar.1
```diff
@@ -0,0 +1,234 @@
...
+	.SLEEP T,
+	MOVEI C,'T00
+	JUMPL A,JUST00
+	LSHC A,-3
+	LSH A,3
+	LSHC A,3
+	IOR C,A 
+	HRLI C,3
+	.OPEN CH2,C
+	JRST JUST00
```
atsampson: Indent instruction after .OPEN?

src/moon/jedgar.1
```diff
@@ -0,0 +1,234 @@
...
+
+MRLMES:	ASCII /Crime does not pay
+/
+MRLPTR:	MRLMES-.,,MRLMES
+
+SPCTTY:	MOVE A,@TTYTBL
+	JUMPGE A,SPCGOT
+SPCUP:	MOVE B,@SUPPRO
+	CAME B,[-1]
+	JRST [HRR U,B ? JRST SPCUP]
```
atsampson: Indent after CAME?

src/moon/jedgar.1
```diff
@@ -0,0 +1,234 @@
...
+SPCDWN:	MOVE A,@TTYTBL
+	JUMPGE A,SPCGOT
+	HRR U,A
+	TLNN A,200000
+	JRST SPCDWN
```
atsampson: Indent after TLNN?

larsbrinkhoff:
> Nice work, Lars!

Thanks!

Actually, this is quite painless now.  I start with your -Sall option to dis10.  Then I run :srccom /$ to compare the new binary with the original.

> I wonder if the 6 vs. 10 was a more subtle attempt at defanging it...

Yes, that thought ocurred to me too.

larsbrinkhoff:
Updated indentation as per @atsampson's suggestions.

larsbrinkhoff:
Yes, I tested it.  JEDGAR check every minute, and when it detects spying it will print a message and exit.  If it has license to kill, it will also detach the toplevel job of the spying user.

PR #1279, opened 2018-10-01
Fix wrong filename for timesharing NSALV.

### LINKED ISSUE #1263: SYS1; NSALV BIN ?
Why do we have this in DB?  I don't think there are supposed to be any * BIN files in SYS:.

larsbrinkhoff:
Resolves #1263.


larsbrinkhoff:
GitLab fails all the time now.

Issue #1277, opened 2018-09-30
LIBDOC; LOOP 818 is broken with respect to ALWAYS condition


eswenson1:
The version of LOOP we have in DB (LIBDOC; LOOP 818) is broken with respect to using the ALWAYS condition. Using it results in an UNBOUND variable failure. The function LOOP-DO-ALWAYS has conditionalizations for For-NIL and not For-NIL.  The #-For-NIL case makes reference to an unbound "pred" symbol.  Using the #+For-NIL logic in both cases appears to work fine.  This function should be updated to use the #+For-NIL logic (actually, the #-For-NIL logic can be commented out and the conditionalization dropped.  

Fixing this will allow MACAID to be compiled.  See related issue #1276.

eswenson1:
Fixed in #1282

Issue #1276, opened 2018-09-30
NILCOM; MACAID > doesn't compile


eswenson1:
It doesn't compile because it gets errors compiling those functions that use LOOP with the ALWAYS keyword.  Turns out that LIBDOC; LOOP 818 has a bug -- it doesn't work with ALWAYS. The NIL conditionalization works fine in MacLisp, but the non-NIL doesn't.  Using the NIL version of the code makes ALWAYS work and allows MACAID to be compiled.  

The following patch to LIBDOC; LOOP 818 fixes ALWAYS:

```
:srccom libdoc;loop 818,ejs;

;COMPARISON OF DSK:LIBDOC;LOOP 818 AND DSK:EJS;LOOP 819
;OPTIONS ARE    /3

**** FILE DSK:LIBDOC;LOOP 818, 20-58 (60205)
                    #-For-NIL `(,pred ,form ,(loop-construct-return nil)))
**** FILE DSK:EJS;LOOP 819, 20-58 (60205)
;                   #-For-NIL `(,pred ,form ,(loop-construct-return nil))
                    #-For-NIL `(,(if negate?
                                     loop-when-function
                                     loop-unless-function)
                                ,form
                                ,(loop-construct-return nil))
)
***************
```

Once LOOP is fixed, MACAID will compile.

eswenson1:
Fixed in #1282

Issue #1275, opened 2018-09-29
LISP; SETF FASL


eswenson1:
Currently, the DB built doesn't build SETF FASL from source.  The reason, up to now, has been that it didn't compile.  The latest source we have for this is NILCOM; SETF 293.  I don't believe there ever was a later version of MacLISP.  Attempting to compile it gets the error (there actually is a second, similar error):

```
(COMMENT **ERROR**  (EVAL-ORDERED* (LIST* (QUOTE A) (APPEND GENSYMS (QUOTE (V)))) (LIST* ARRAY (APPEND INDICES (LIST VAL))) (LIST !
(QUOTE LIST) (LIST* (QUOTE LIST) (QUOTE (QUOTE STORE)) (LIST* (QUOTE LIST) (QUOTE (QUOTE ARRAYCALL)) (LIST (QUOTE QUOTE) (SETF-USE!
R-SLOT EXPF)) (QUOTE A) GENSYMS) (QUOTE (V)))))
                Wrong number of args in function |ARRAYCALL SETF-X-INVERT|)
; DATA ERROR - TO PROCEED TYPE $P
;BKPT DATA
```

What the compiler is saying is that the call to eval-ordered* is supplying an incorrect number of paramers when called from |ARRAYCALL SETF-X-INVERT|.  Now, this frequently occurs because the function in question takes a variable number of arguments, but this fact isn't known to the compiler. What typically happens is that the compiler examines the first call to an function it doesn't know anything about and remembers the number of arguments it was passed.  As long as all other calls in the compilation unit pass the same number of arguments to that function, all is ok.  However, if the compiler sees another call with a different number of parameters, the above "Wrong number of arguments in function..." error is raised.

However, in this case, there are two calls to eval-ordered* in NILCOM; SETF 293; and both calls pass 3 arguments.  So that is not the issue here.

It turns out the issue was caused by the fact that SETF, itself, declares eval-ordered* (which is actually defined as an EXPR/SUBR in LISP; EVONCE FASL) as an LEXPR.  I have no idea why it does that.  And, since the compiler loads SETF and load/dump time, the compiler has recorded that eval-ordered* is an LEXPR.  To fix everything, you have to do a few things:

1) update SETF to not do a (*LEXPR EVAL-ORDERED*).
2) bring up COMPLR, break to a REPL, and remove the property with (REMPROP 'EVAL-ORDERED* '*LEXPR)
3) then compile the updated SETF, installing the resulting FASL as LISP; SETF FASL
[...]

Issue #1272, opened 2018-09-28
Updated lisp compiler


eswenson1:
DB has COMLAP; COMAUX 25. There is at least COMAUX 27 to be had. DB has COMLAP; COMPLR 936 and there is a COMPLR 937.  DB has PHAS1 86, but there is a PHAS1 87.  DB has SRCTRN 20, but there is a SRCTRN 21.

Might want to get COMPLR COMP, COMPLR LOADER.  DEFMAC 77, DEFUN& 13, ELAP 106, FASLAP SETUP, FDRV 32, FORMAT 73, GENFNS 71, GFILE 422, GFN 456, GGLOAD 26, GRILAP 16, GRIND 368, INDEX 150, LAP 109, LEDIT 21, LET 15, LOADER 1, LOSER 898, MACROS 5, MAKE@ LIST, MAKIDX NCOMPLR, NFASL 17, NFASL ARITH, NFASL EVAL, NFASL LAP, NFASL MAKWRD, NFASL RANDOM, NFASL TROUBL, NGING LISP, NINDEX 115, STRACE 21, SUBFOR 29, TNXLED 11, TNXLED EMACS, TRACE 66, TRACEA 65, TRACEB 54, TRACEC 54, TROUBL LAP, ZTYI 1.

eswenson1:
Hmm.... I'm not positive that COMPLR 937 is "newer" than COMPLR 936.  And COMPLR 937 sets its version number to 935.  Also, there is, what I think is file damage in COMPLR 936 and 937.  In the function COMPLRVERNO, there is a symbol '|/iLISP COMPILER |, that has a funky character in the second position (here depicted with "i").  I believe this is supposed to be an "n" to cause a newline.  If you compile COMPLR 936 or COMPLR 937 and dump the resulting compiler, the compiler advertises itself as nLISP COMPILER rather than LISP COMPILER and if you type ^G when in the compiler, it does not emit a new line before LISP COMPILER, but rather displays QUITnLISP COMPILER.  The original FASL file (that we use in our current compiler) emits:

```
*
QUIT
LISP COMPILER 2155 [in LISP 2154]
```
The newly compiled version emits:
```
QUITnLISP COMPILER 2156 [in LISP 2154]
```

eswenson1:
[...]

PR #1282, opened 2018-10-01
Builds all LISP; * FASL files that are on autoload properties 

### LINKED ISSUE #1271: LISP; STRING FASL
This, and possibly other FASL files in LISP; are not built from source. These should be all built from source.  source for LISP; STRING FASL is NILCOM; STRING 186.

eswenson1:
Builds all LISP; * FASL files that are on autoload properties when the lisp interpreter is first booted.

Redumps lisp compiler with updated FASL files built from source.

Resolves #1271, #1272, #1275, #1276, and #1277.


eswenson1:
@larsbrinkhoff I have one commit here and multiple tickets are being resolved. It is too difficult to extricate all these into separate commits. There are a lot of interdependencies.  For example, some sources can't be compiled until issues are fixed.  So I had to fix all the issues in order to finish the task of compiling all the fasl files used by initial autoload properties.  

And while I could have separated out the compiler redump from this in a separate commit, I didn't because I didn't want to leave a possible build (of the first commit) with a compiler that had inconsistencies with respect to what other programs would see if they used LISP; fasl files.  So now we have a consistent library and compiler.  Well almost.

I'm going to do more reorganization in another commit/PR and will likely want to recompile the compiler before redumping it out.  (First compile uses the runtime from the binaries on the sources tape, second compile uses the recompiled runtime).  To do this *really* write, there should be a third compile that uses the previously-compiled compiler, but I don't really think that is necessary.

eswenson1:
@larsbrinkhoff  what do we do about this? the klh10 build failed to produce any output within 10m during the dump.   This builds fine and runs fine for me.

larsbrinkhoff:
I restarted the KLH10 build.  I think it's an intermittent failure.

larsbrinkhoff:
Good to go now.

PR #1214, opened 2018-10-02
ULISP - Lisp in Lisp, for teaching

larsbrinkhoff:
TEACH; ULISP 459


larsbrinkhoff:
Two outstanding items.

- Dumping asks `Editor Communication Filename:`, my script currently answers `FOO`.
- Next there's the question `Script Filename`.  I answer `BAR`.

eswenson1:
I don't see where you are responding "FOO" and "BAR" to prompts....

larsbrinkhoff:
I changed them after I wrote the comment.  The DUMP function only allows for one word without any whitespace.  I changed FOO and BAR to EDIT and SCRIPT.

Typing ^E inside ULISP jumps to a previously started job, supposedly an editor.  I picked EMACS.

larsbrinkhoff:
From https://archive.is/uVmiM

> I designed and implemented a dialect of Lisp, Ulisp, that was used for one semester teaching during a transition period before Scheme was available for students. (For those who were at MIT, Ulisp was vaguely based on Delphi Lisp, but was implemented in Maclisp, and basically hid most of Maclisp complexity and weirdness in favor of a simpler model designed for teaching.) 

From Usenet comp.lang.lisp:

> I wrote a dialect of Lisp called ULISP that was used briefly [one semester] at MIT for teaching before Sussman ran screaming from it (because it didn't have tail call elimination) and decided to use Scheme.  ULISP had first-class macros, but no compiler.

larsbrinkhoff:
I'm copying @netsettler's comment from Reddit:

> The context, as best I can recall, was that Lisp was being taught at MIT on a PDP-11/45 (originally running the Delphi operating system, if I remember, and later Unix). MIT's EE&CS folks were wanting to move to something more powerful. I believe the course number was 6.031, as it was not yet 6.001. I had only recently finished the course myself, but I used to do my homework on a PDP-10, where I'd written an emulator of their Lisp in the local Lisp in order to do my problem sets without leaving my office across the street. Someone heard there was a possibly-suitable Lisp on the 10 and asked me about using it.
> 
> The Lisp on the PDP-11 was not the sort of Lisp you'd want to perpetuate though, and I volunteered to make something clean and pretty for class use. Not for pay even, I don't think. Just because it seemed fun. The "U" in "ULisp" refers to "unix", since it had originally been inspired by a lisp that ran there. And this is how ULisp (no relation to EuLisp) arose.
> 
> The result was used for one semester for teaching at MIT. But Sussman kept pestering me about tail calling and I kept saying "but won't that make stack debugging hard?" He and I were normally on friendly terms, I think, but he was not at all happy about this detail. In light of history of SICP, it's easy to see why. But at the time, in context, I was being dense and didn't quite get the focus of what he wanted to do with his curriculum and why he was so obsessed with tail calling, which seemed to me such a detail. :) The next semester, as I recall, they switched to Scheme for the class, and ULisp was quickly forgotten. I wasn't bothered, but at a friendly intellectual level I perceived it as a furious rejection of my Lisp. More likely it was a planned move all along. I certainly did not expect this Lisp to function as anything more than a stopgap.
> 
> But I did design this Lisp dialect to have some interesting characteristics. The way macros and special forms are integrated are different than other Lisps, as I recall, and probably rely critically on the way this Lisp is an interpreted-only Lisp. I just wanted people to be able to get the idea of what macros do.
> 
> And, of course, this requires Maclisp to run. So you'd need a PDP-10 emulator in order to even try this.
> 
> There is a manual for this dialect that I have at least in hardcopy. I'll see if I can find it. It was written in the same style as my revised maclisp manual, in a plaintext notation that compiled down to TeX as a backend.

PR #1296, opened 2018-10-03
Fix GTLOAD to allow lower case input.

larsbrinkhoff:
Issue reported by @TheFausap.


larsbrinkhoff:
It's not an unreasonable thing for the TTY device to do, but if they took it out I'd hesitate to put it back in.

Maybe have ITS print a warning to the console?  So we can find these programs more easily.

PR #1299, opened 2018-10-03
Delete UNFASL files in lisp; and liblsp; directories.

### LINKED ISSUE #1289: Cleanup UNFASL files in LISP; and LIBLSP; directories
In my latest commits to build all LISP; fasl files from source, I didn't clean up all the UNFASL files created by the build. There is some value to UNFASL files (for debugging and investigatory purposes) but in some case, they cause too many entries in a directory and therefore have to get pruned. I have a bunch of DELETE commands in lisp.tcl to clean up UNFASL files that prevented the build from succeeding (due to directory full conditions) and may wish to clean these up so that they don't accumulate and prevent future futures builds.  But I consider this a lower priority than actually getting things to work so am not updating my previous commit, but rather entering another ticket.

eswenson1:
Resolves #1289.

### DIFF
--- a/build/lisp.tcl
+++ b/build/lisp.tcl
@@ -938,6 +938,13 @@ respond "_" "\032"
 type ":kill\r"
 respond "*" ":link liblsp;vsaid fasl,lisp;\r"
 
+# cleanup unfasl files in lisp;
+respond "*" ":dired lisp;\r"
+respond "@" "delete * unfasl\r"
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Isn't there a long list of `respond "*" ":delete foo; bar unfasl\r"` somewhere?

eswenson1:
Yes, i’ll Get rid of that soon.

PR #1305, opened 2018-10-04
Update location of TCTYP; no link.

### LINKED ISSUE #1261: TCTYP location
I noticed that on DB, we have SYS; TS TCTYP as a link to SYSBIN; TCTYP BIN.  However, on ES/PI, we have SYS1;TS TCTYP as the binary file.  May way to see which was more historically correct.  I don't think TCTYP is PDUMP'ed, is it?  If not, then we don't really need the SYSBIN; TCTYP BIN.

larsbrinkhoff:
SYS1; TS TCTYPE should link to TS TCTYP.

Resolves #1261.

### DIFF
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -34,9 +34,8 @@ respond "*" ":print device;..new. (udir)\r"
 type ":vk\r"
 
 # TCTYP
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
What's wrong with it?

PR #1313, opened 2018-10-06
Updating mismatched apostrophes and single quotes

dend:
In the `zork` file, the apostrophes and single quotes are currently used to start and open a sentence. This PR standardizes on one character.

### DIFF
--- a/src/sysen3/zork.21
+++ b/src/sysen3/zork.21
@@ -31,7 +31,7 @@ zork:	syscal OPEN,[ %clbit,,.uao\%tjdis ? %climm,,ttyo  ;open TTY for output
 There appears before you a huge figure clothed in a dark
 robe. As you shrink back in awe, he speaks:
 
-   ``This is not the machine of the Zork Implementors!''
+   ''This is not the machine of the Zork Implementors!''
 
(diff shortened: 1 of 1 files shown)


eswenson1:
Isn't that the way some folks get the two different directioned quotation marks?

dend:
@eswenson1 probably. This is definitely one of those "_Nit, I just like consistency_" PRs - no hard feelings if it gets rejected.

larsbrinkhoff:
There's a tension between software preservation on one hand, and making updates on the other.  (Disregard for a moment there's no consensus on whether to begin a quote with <code>``</code> or `''` or `"`.)  We want to stay true to the original form and the spirit of ITS, whilst still allowing the code to live, evolve, and interoperate with current technologies.

I don't have a quick and easy answer, and of course it's highly subjective what the right thing to do is in each case.

larsbrinkhoff:
In this case, I don't see a huge benefit over keeping the text as the original authors intended it.

Of course, if the Zork imps were to sign off on this change, it would be a different matter. ;-)

larsbrinkhoff:
No one has made a strong argument either way, so I'll close this.

PR #1314, opened 2018-10-06
Add assembler and disassembler for Intel 8748

eswenson1:
Updated both assembler and disassembler to work on Maclisp and NIL (as well as LISPM, for which they were originally written).

Note, while the assembler appears to work, I haven't been able to figure out how to test the disassembler.


larsbrinkhoff:
Did you test it with NIL?!?

eswenson1:
No, I didn’t test with NIL. But I used macro expansions that should work for NIL.

eswenson1:
What do I do about the failed circleci job. I don’t think I have access to restart it.

eswenson1:
I don’t think I can merge with the circleci failure.

larsbrinkhoff:
I only see a GitLab failure, and it's not required to succeed for merging.  I can click the merge button if you like.

eswenson1:
Hmmph. Silly mobile interface: “Pull requests that have a failing status can’t be merged on a phone.”

eswenson1:
I can’t merge. Please do.

larsbrinkhoff:
Done.

src/lspsrc/umlmac.42
```diff
@@ -5,7 +5,7 @@
 ;;;  ** (c) Copyright 1981 Massachusetts Institute of Technology *************
 ;;;  *************************************************************************
 
-(herald UMLMAC /40)
+(herald UMLMAC /42)
```
atsampson: Could just say `(herald UMLMAC)`, and it'll use the FN2 as the version (e.g. like MAXTUL does).

src/lmio1/ukbd.prom
```diff
@@ -0,0 +1,917 @@
+PROM LOCATION UKBD SUM-CHECK NIL
```
atsampson: What is this file - a dump of a real PROM made with PROMP (since that seems to be where the header comes from)? Or a PROMP conversion of AS8748's output?

src/lspsrc/umlmac.42
eswenson1: You're probably right.  I was just following the pattern used for this file over the years. The version was always explicit.  And in this case, wrong.

src/lmio1/ukbd.prom
larsbrinkhoff: I don't know.  I never heard about PROMP but if it writes that header, then maybe its a dump.

larsbrinkhoff:
Better port LMIO1; PROMP > so we can programs those keyboards.... will this ever end?

ams:
The Lisp Machine is taking over the world! 


Question, how do we handle the LM changes, back porting things to LM-3 or whatever?

larsbrinkhoff:
Right, the files now exist in two copies.  I don't expect these programs will see frantic development ahead, so just copy the files over.

ams:
https://github.com/LM-3/sys/issues/11

PR #1316, opened 2018-10-06
Updated to build remaining lisp; fasl files from source.

### LINKED ISSUE #1286: LISP; CNVD , EXTHUK, GCDEMN, GFILE, GFN, LISPT, QUERIO, SENDI, STRAUX, VECTOR FASL files
Need to get compiled from source. 

Source for CNVD should be NILCOM; CNVD 2.  
Source for EXTHUK should be LSPSRC; EXTHUK 34
Not sure where source to GCDEMN is.
Source for GFILE may be LSPSRC; GFILE 422
Source for GFN may be LSPSRC;  GFN 456
Source for LISPT may be LIBDOC; LISPT JLK708
Not sure where the source to QUERIO is.
Source for SENDI should be  LSPSRC; SENDI 107
Source for STRAUX should be LSPSRC; STRAUX 74
Source for VECTOR should be LSPSRC; VECTOR 74 (had issues compiling this one earlier)

Once we get these compiled, we'll have all the FASL files in LISP; compiled from source.

eswenson1:
Resolves #1286.


build/lisp.tcl
```diff
@@ -1005,6 +1008,33 @@ respond "_" "lisp;_nilcom;defset\r"
...
+expect ":KILL"
+respond "*" ":midas lisp;_lspsrc;straux\r"
+expect ":KILL"
+
+# lispt source is in libdoc, therefore fasl should be in liblsp
+# version in lisp; should be a link to liblsp;lispf fasl
+respond "*" ":delete lisp;lispt fasl\r"
+respond "*" ":link lisp;lispt fasl,liblsp;\r"
+respond "*" ":link sys2;ts lispt,sys2;ts edit\r"
+respond "*" ":link sys2;ts edit,sys2;ts emacs\r"
```
larsbrinkhoff: The build fails due to SYS2; TS EDIT already existing.  It's created in emacs.tcl.

eswenson1:
Yeah, i’ll fix. Build takes so damn long that I didn’t test locally all the way through.

eswenson1:
What is wrong this circleci?  I don't have permissions to restart the build.  Can you give me permission?

src/lspsrc/exthuk.34
```diff
@@ -0,0 +1,121 @@
+;;;   EXTHUK			-*-Mode:Lisp;Package:SI;Lowercase:T-*-
+;;;   ****************************************************************
+;;;   *** MACLISP **** EXTended datatype scheme, compiler helper *****
+;;;   ****************************************************************
+;;;   ** (c) Copyright 1981 Massachusetts Institute of Technology ****
+;;;   ****************************************************************
+
+(herald EXTHUK /33)
```
atsampson: /34 or just (herald EXTHUK)?

src/lspsrc/vector.75
```diff
@@ -0,0 +1,324 @@
+;;;  VECTOR    				-*-Mode:Lisp;Package:SI;Lowercase:T-*-
+;;;  **************************************************************************
+;;;  *** MacLISP ******** VECTOR support **************************************
+;;;  **************************************************************************
+;;;  ******** (c) Copyright 1982 Massachusetts Institute of Technology ********
+;;;  **************************************************************************
+
+(herald VECTOR /74)
```
atsampson: 75

src/lspsrc/exthuk.34
eswenson1: Good catch.  I didn't change the source -- I just included the one found on backups.  There is an issue of faithfulness here, but in this case, we can either a) update the herald in this commit, or b) add an issue to fix the heralds in various files where they are wrong -- introducing a new version (with corrected herald).  Otherwise, we're changing an original source without updating the FN2 (version) and can set ourselves up for confusion when "our" version N is not the same as the historical version N.  Thoughts?

src/lspsrc/vector.75
eswenson1: Now here, I actually changed VECTOR 74 to get it to compile.  So here, I'd be comfortable either a) removing the version number from the HERALD or b) updating it to 75.  Recommendations?

PR #1320, opened 2018-10-07
Separate macsyma build from lisp build. 

### LINKED ISSUE #1283: Split macsyma build out from lisp build
Lisp is pretty fundamental and lots of things depend on lisp.  Macysma is a very big app that takes a long time to build and has a limited audience. We shouldn't combine the lisp and macsyma builds and should make it possible to build ITS without macsyma (but with LISP).  

This ticket is to create a task of moving the macsyma-specific build steps out of lisp.tcl and into its own macsyma.tcl.

eswenson1:
Allow building without macsyma.  Resolves #1283.


eswenson1:
Lars, are you good with this change too? (Before I merge.)

PR #1191, opened 2018-10-08
Use Morton box, high-speed terminal multiplexor

larsbrinkhoff:
Set MTYP==1 and NMTYS up to 32.  Speed settings include the usual suspects, plus 25, 40, 50, and 80 k!  In particular, the Imlacs were used with 50k speed.


larsbrinkhoff:
This doesn't quite work.  Have to check why.

larsbrinkhoff:
Apparently the images I build still had MTYP=0, or maybe I copied the wrong files to the wrong place.  But assembling a new ITS did the trick, so it should be OK.

larsbrinkhoff:
Used on both DM and ML.

Issue #1325, opened 2018-10-08
OSX building


TheFausap:
Hello,

I found a couple of issues during a build under OSX (High Sierra).

1) touch under OSX is different: -d option is not present, so I modified the makefile with the correct syntax, using a check for the OSTYPE

2) ka10 simulator seems is not able to find the correct X11 libraries. After installed XQuartz, I modified the makefile, when it searches for libXt, adding to OS_CCDEFS the X11 include path, and everything went fine.

Now I'm building, and it seems everything ok.

If you want I can create a new branch with these changes.

TheFausap:
Created a new branch : osx-build
I also modified the makefile under tools/sims, but I'm not authorized to push to that repo.

Just few notes about running under OSX / XQuartz: to be able to create the X11 window for 340 display, the sim needs to be start from an XTERM terminal (not the standard OSX terminal app).
The GT40 terminal start without any problem, because it's an SDL application.

Could be possible, modify the 340 display to use SDL too ? If so, we could skip the XQuartz part.

larsbrinkhoff:
Thanks, I'll take a look some time soon...ish.

I'm afraid the 340 display is outside our jurisdiction.  As far as I know, direct use of SDL is not allowed.  You will have to go to http://github.com/simh/siimh and check with the people there.

larsbrinkhoff:
I pushed another commit onto your branch to build it with Travis CI:  
https://travis-ci.org/PDP-10/its/branches

larsbrinkhoff:
I need more changes to appease Travis CI, so I made a new branch off yours: lars/osx.

- KLH10 builds, but the expect script doesn't.
- KA10 sims complains about not finding X11/X.h.  Maybe your OS_CCDEFS fixes that?
- KS10 simh completed a full build.

larsbrinkhoff:
Logs here:  
https://travis-ci.org/PDP-10/its/builds/439271789

TheFausap:
Yes Lars, OS_CCDEFS is needed to find X11/X.h
In the DARWIN section of makefile, I added `OS_CCDEFS += -I/usr/X11/include`

rmaldersoniii:
Back in the mists of time, I had to make some changes to KLH10 to build on OSX. I don't know whether these are helpful, but I put them on my personal web site in 2011:
http://www.panix.com/~alderson/klh10-2.0h_for_OSX.6.8/

larsbrinkhoff:
[...]

PR #1332, opened 2018-10-10
CI build on Mac

larsbrinkhoff:
Building on what @TheFausap did in #1325.  Builds using SIMH KS10 on Travis CI.  KLH10 doesn't work well with expect for unknown reasons.  The KA10 simulator complains about not finding X11/X.h.

Tested `touch -t` on Linux and NetBSD, so I suppose it's not a problem to change from `touch -d`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os:
+  - linux
+  - osx
 language: c
 dist: trusty
 sudo: required
@@ -5,6 +8,12 @@ env:
   - EMULATOR=simh BASICS=yes
   - EMULATOR=klh10 BASICS=yes
   - EMULATOR=sims BASICS=yes
+matrix:
+  exclude:
+    - os: osx
+      env: EMULATOR=klh10 BASICS=yes
+    - os: osx
+      env: EMULATOR=sims BASICS=yes
 install: sh -ex build/dependencies.sh install_${TRAVIS_OS_NAME:-linux}
 script: make check-dirs all
 deploy:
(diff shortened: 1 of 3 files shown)


TheFausap:
We should add the `OS_CCDEFS += -I/usr/X11/include`, in the sims Makefile. The problem arise cause XQuartz doesn't create a sim link in /usr/include pointing to X11 directory.

larsbrinkhoff:
Please suggest the `OS_CCDEFS` change in the SIMH repository: http://github.com/simh/simh

TheFausap:
I could do that, but PDP10 in simh, doesn't have any issue because it doesn't use X11 at all. PDP11 has no issue because SDL2, under OSX, because doesn't call X11.
So I don't know how to motivate my request.

larsbrinkhoff:
The SIMH PDP-1 simulator should have the same issue.  It includes a Type 30 display using the SDL library.

TheFausap:
But SDL is ok, the type 340 display unfortunately uses an X11 interface, not SDL, and that's the problem.
We could fix the issue if we could rewrite the 340 display code using SDL.

TheFausap:
AFAIK, the problem is in the SIMS repo, not SIMH.

larsbrinkhoff:
I fixed some things even though the SIMH KS10 doesn't use the 340:  
https://github.com/simh/simh/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+340+author%3Alarsbrinkhoff

TheFausap:
Ok, I didn't find anything because this wasn't in the master branch.

PR #1348, opened 2018-10-15
DOWNLD - print file in super-image mode to TTY.

larsbrinkhoff:


larsbrinkhoff:
Travis is done, but apparently the signal to GitHub was lost.

PR #1352, opened 2018-10-16
OINIT - init program reporting last login, email.

larsbrinkhoff:
This is a C application.

Thanks to @mattwyrm for the YouTube instructions on how to build C programs on ITS.

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -1163,6 +1163,17 @@ respond "DSK: SYS; TS NTJ6" "\r"
 respond "*" ":kill\r"
 respond "*" ":link sys; ts tj6, sys; ts ntj6\r"
 
+# OINIT
+respond "*" ":cwd c\r"
+respond "*" ":c;cc sysen2/oinit\r"
+expect ":KILL"
+respond "*" ":stinkr\r"
+respond "=" "x c/clib\r"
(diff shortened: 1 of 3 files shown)


larsbrinkhoff:
Yes, me too!  I don't know any other real system program.  For some value of "real".

There's the C compiler itself, and its library, and the linker STINKR.  And there's R, the text formatter.

larsbrinkhoff:
Oh, I remember now TS SHELL which is a system program.  But we don't have the source code.

larsbrinkhoff:
This runs into the problem that SIMH can't run the C compiler: https://github.com/simh/simh/issues/403.

larsbrinkhoff:
Agreed.  Can't merge this as long as SIMH breaks.

larsbrinkhoff:
The SIMH issue now has an explanation why it can't run the C compiler.

larsbrinkhoff:
I checked all TS files to see if they link against [CLIB].  I found many C applications:

- R; TS R42
- SYS2; TS CP
- SYS2; TS IPAK
- SYS2; TS RALP
- SYS2; TS SEARCH
- SYS3; TS BROKEN
- SYS3; TS DWIM

larsbrinkhoff:
TS SHELL seems to be statically linked with the C library.

PR #1356, opened 2018-10-19
Build C library from source

larsbrinkhoff:
Instructions from CLIB; MAKLIB C

1. Edit the file MKCLIB STINKR to contain all of the files you want to be in the library.  See that file for further instructions.
2. Create a TS CLIB in the C directory by running Stinkr on MKCLIB STINKR.
3. Create a TS MAKLIB using MAKLIB STINKR.
4. Run TS MAKLIB.

PR #1359, opened 2018-10-21
Build CC from source

larsbrinkhoff:
Build C; TS CC from source code files.


larsbrinkhoff:
This depends on the C library (#1356) being in place.

larsbrinkhoff:
The old TS CC is linked against C; [CLIB] 14.  However, it can't be removed until C; _M BIN is also rebuilt.

larsbrinkhoff:
I added a fix for the CLIB build script.  When MAKLIB is run, it valrets a string to kill itself and run MIDAS.  Since my script only looked for "*", it tried to inject input when MIDAS was running.

larsbrinkhoff:
Tested with OINIT (#1352).

PR #1365, opened 2018-10-26
C compiler LP, C, and M phases

larsbrinkhoff:
Saildart has the missing source code for the C compiler: https://www.saildart.org/[C,SYS]

- LP is lexing and parsing.
- C is code generation.
- M is macro expansion (to target assembly language).


larsbrinkhoff:
Getting into shape.  This could probably go in as is, but I think I'd like to reverse engineer C4220 MID to a C source file.

larsbrinkhoff:
Feel free to comment on my C42 C.  It can probably be improved upon, but I'm a bit exhausted spending a few days writing it.

It compiles to a MIDAS file which is almost idential to C4220.MID[C,SYS] from SAIL.  Many numbered labels are off by one.  The original file was compiled with an optimization which isn't present in our version.

PR #1370, opened 2018-10-28
Be more careful about sending input to MAGDMP.

larsbrinkhoff:
Using expect's `send` sometimes fails.  Exact cause unknown, but it looks like characters get dropped.


larsbrinkhoff:
I'll rerun this on CircleCI a few times to see if it's an improvement.

larsbrinkhoff:
I think this did the trick, or at least it's not making things worse.

PR #1374, opened 2018-10-30
Change GitLab CI to use the KA10 simulator.

larsbrinkhoff:
Because it doesn't work with KLH10.  For unknown reasons, it seems to work fine everywhere else.


larsbrinkhoff:
Because it's hard to debug a remote computer to which you have no access.

PR #1375, opened 2018-11-01
Clean up C documentation.

larsbrinkhoff:
Forgot this:

- Rename CC.PAT[C,SYS] to CC PATCH.
- Remove E padding from CCDOC.


ams:
Maybe use a .gitattributes file to explicitly mark them as text?

larsbrinkhoff:
Not a bad idea, but I don't think GitHub honors that.

larsbrinkhoff:
The `diff` attribute forces git to display a text diff even though it detects a binary file.

larsbrinkhoff:
I checked.  Neither `linguist-language=Text`, nor `diff` affects GitHub.

PR #1376, opened 2018-11-02
.gitattribtues (doc/c/cc.pat, doc/c/ccdoc.text): Mark as text files.

ams:
See https://github.com/PDP-10/its/pull/1375

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
-/src/*/* linguist-language=assembly
+/src/*/*	diff linguist-language=assembly
+/doc/*/*	diff linguist-language=Text


larsbrinkhoff:
Thanks.  Everything in /doc and /src should be text.  Can you extend the scope of this pull request?

larsbrinkhoff:
Oops, /src is already "Assembly".  That's not correct, but better than Text.

So just make /doc Text.

larsbrinkhoff:
Please also add the `diff` attribute to /doc/*/* and /src/*/*.  It doesn't affect GitHub, but it does git.

ams:
What do you want the diff attribute to be set too?

larsbrinkhoff:
Checks passed.  @ams, would you do the honors?

Issue #690, opened 2018-02-28
MINI on ITS?


ams:
When creating a new load band on the CADR, we use a small service called MINI which provides a very basic Chaosnet file transfer protocol to load the core system into memory.

Does anyone know if MINI is available on ITS?

larsbrinkhoff:
I'm glad you asked!  There's not only MINI, but also MINIC and MINI36.  I have been wondering what all those are.  So now you can be the expert to explain all this, yay!

larsbrinkhoff:
See this: https://github.com/PDP-10/its-vault/commit/4c8fa2581abc2c85f96fea5b3501c8ba1233ae4d

ams:
Ah, perfect! Also some documentation in https://github.com/PDP-10/its-vault/blob/4c8fa2581abc2c85f96fea5b3501c8ba1233ae4d/files/syseng/minisr.5.

The client program for CADR can be found here:  https://github.com/LM-3/sys/blob/master/lmio/mini.lisp

ams:
I’ll see if I can figure out the difference between MINI, MINIC and MINI36.

larsbrinkhoff:
Thanks.

Maybe MINI36 was used to send files to a KS10 running NSALV.  It was an option to bring up a new computer instead of using magtape.

larsbrinkhoff:
I'm changing the label from question to todo.  Fixing this is up for grabs!

larsbrinkhoff:
Ping @ams regarding MINI vs MINIC vs MINI36.

ams:
`MINISR` is the source for `MINI`.  `MINIC` is the early contact name, it was then renamed to just MINI.

larsbrinkhoff:
Thanks.  So if you need MINI we can add MINISR to ITS.  Presumably the latest version.

@earl-ducaine, is this something you want?

larsbrinkhoff:
Maybe possibly MINIC was used with the CONS?  But then I don't think ITS got Chaosnet until 1978 which would really be a bit too late for CONS I think?

So I think the right thing to do at this point is to just add the MINI server.

larsbrinkhoff:
The Lispm minisr.mid.44 is not for ITS. :-(

larsbrinkhoff:
The only version for ITS is MINISR 5 which uses the MINIC contact name.  I'll see about getting a newer version.

ams:
[...]

PR #1377, opened 2018-11-03
MINIC - mini Chaosnet file server.

larsbrinkhoff:
For booting Lisp machines.  @ams is this useful, or do you need MINI?


larsbrinkhoff:
See #690 for not a lot of more information.

ams:
I need MINI.

PR #1378, opened 2018-11-03
Chaosnet DOOR service

larsbrinkhoff:
DEVICE; CHAOS DOOR  
LMIO1; DOOR 1

```
;A chaosnet server which performs useful functions like opening the
;9th floor door to let urchins in to rip off virtually everything in the place.

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -111,6 +111,10 @@ respond "*" ":midas sysbin; minic bin_syseng; minisr 5\r"
 expect ":KILL"
 respond "*" ":link device; chaos minic, sysbin; minic bin\r"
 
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
Sure, if you have a PDP-11 on the 10-11 interface, and hardware register
764050 responds to the right bits!

larsbrinkhoff:
I added one more commit.  ESCE is the Chaosnet client.

Probably so named because it's the same function as typing ESC E on the TV keyboard.

eswenson1:
I remember ESCE and the ESC E!  Used it to call the elevator when I was ready to go back to my dorm.  :-)

PR #1391, opened 2018-11-09
Documentation updates

larsbrinkhoff:
- Make INFO; INFO > from INFOD > using EMACS' *DOCOND* library.
- Add miscellaneous documentation files.


larsbrinkhoff:
The new INFO; INFO > is almost identical to the previous one.  I made it version 48, which is now a fixed number in the build script.

PR #1267, opened 2018-11-10
LLOGO add-ons

larsbrinkhoff:
Get these LLOGO files built:

- TURTLE
- TVRTLE
- GERM
- MUSIC

Also fix features currently not supported:

- Compilation (needs LLOGO; DECLAR >)


eswenson1:
I don’t think you need to declare SUBRs. That is the default, right?

larsbrinkhoff:
No, I didn't need to add them.  It was just in case.

eswenson1:
You can get TVRTLE to compile by fixing the incorrect syntax that surrounds the definition of wipeclean inside the [color ... ] conditionalization.  In that definition of wipeclean, you can see that:

```
NO-VALUE)
```
doesn't belong there.  This should be replaced by a "]", which will close off the [COLOR conditionalization.  If you make that change, it compiles cleanly.

larsbrinkhoff:
Thanks!  I think the `NO-VALUE)` line should be removed.  There's a closing `]` further down with the comment "END OF COLOR CONDITIONAL SECTION."

larsbrinkhoff:
On second thought, I think NO-VALUE should be returned, like other primitives do.

eswenson1:
Yes, I just looked at this again. The correct fix is to simply remove the extraneous ")" before the NO-VALUE) at the end of the WIPECLEAN function inside the [COLOR ...] conditionalization.

larsbrinkhoff:
Yes, I did that and the build went through.

larsbrinkhoff:
Now that I rebased the pull request on top of the updated files, I found that the newer TVRTLE didn't have any problem.  So this just fixes the 340 TURTLE, and compiles the files.

PR #495, opened 2018-11-12
VIEW - read file

larsbrinkhoff:
SYS3; TS VIEW  
SYS3; TS VDIR - link to above

Dumped from TECO?  I don't know how to find out what it is, or where the source code is.


eswenson1:
Does that DUMPT command line really work for you?  It doesn't for me.  I need to say:

   :DUMPT $turnip;view comprs

without the comprs, I get a file not found error.

eswenson1:
Lars: are you sure you want to be changing historical files (e.g. TURNIP; VIEW 1) and not incrementing the version number?  I've always incremented the version number unless the file was actually damaged.  I think it is important to be able to "see" where we've made changes to the original files.  I would commit either both VIEW 1 and the modified VIEW 2, or just VIEW 2 (with your changes).

eswenson1:
I can't seem to get this to work on ES using your approach.  Also, I'm not sure you are being historically accurate here.  There was no TURNIP; TURNIP TECO (e.g. on MC) and no link to .TECO.;TECPUR > in TURNIP.  These existed in KMP.  And the contents of KMP; KMP TECO used KMP; as the default directory name.  I'd recommend:

1) adding KMP; KMP TECO
2) adding link for TS DUMPT in KMP;
3) using the DUMPT command line:  $TURNIP; VIEW COMPRS

I say 3, because I see no historical evidence that there was a TURNIP; VIEW START or a KMP; VIEW START.  It works fine if you specify TURNIP; VIEW COMPRS on the DUMPT command line.

eswenson1:
I can now reproduce your way of building (although I still object to move from KMP to TURNIP) on ES. You had made the change to the TECO init file that KMP suggested -- I didn't have in my KMP; KMP TECO because I had been successfully using the original KMP; KMP TECO.  However, it only worked for me because I had been specifying TURNIP; VIEW COMPRS on the DUMPT command line rather than relying on the KMP; DSK: FOO START defaults.  When these defaults were merged with the JCL, the bad filename resulted (as we've discussed).  With KMP's fix to the TECO INIT file, the merge works successfully, and I can dump out TS VIEW using only $TURNIP;VIEW (or $KMP;VIEW) on the DUMPT command line.

larsbrinkhoff:
Good points.  I have moved things around but I'll move back to KMP.  I'll bump the version number too.  As you probably saw by now, I made a link from VIEW START to VIEW COMPRS.

There actually is evidence to suggest the file name passed to DUMPT was `turnip;view`.  In the original SYS3; TS VIEW, right after the first copy of KMP TECO, that exact string appears.  In our newly dumped TS TECO, it reappears in the same place.

I also forgot to move the dumped file to SYS3.

eswenson1:
Thanks. Yes, I did see the link to TURNIP; VIEW COMPRS.  The reason this didn't work for me was because in order to get things working using your code, I had simply copied my KMP; KMP TECO to TURNIP; TURNIP TECO, but of course, the directory default was wrong as a result.

Since you've found evidence that the JCL passed to DUMPT was indeed TURNIP; VIEW, that does mean that you should have TURNIP; VIEW START as a link to TURNIP; VIEW COMPRS.  I don't see evidence that there was a KMP; VIEW START in backup tape info. I do, however, see evidence of KMP; KMP TECO and not TURNIP; TURNIP TECO.  (I suppose Kent could have logged in as TURNIP to create TS VIEW, but I think more likely that he did it as KMP).

larsbrinkhoff:
Agreed.  I'm about to push an updates pull request.

One remaining detail is the link from KMP; TS DUMPT.  To be really historically correct, it should be to SYS; TS NT, but DB doesn't have such a file.

eswenson1:
Probably should add sys; ts nt. Should be a link to .teco.:tecpur >. I think our link for sys; ts teco should probably NOT be a link to tecpur >, but rather to a specific version.

Interestingly, ES/PI had TECO at 928 and NTECO as >.

netsettler:
In reviewing this file, I noticed an oddity.  It is possible this is just a transport problem with me looking at the file, but since I can see this issue independently in several different ways, I'm doubtful.

The character at position 1721 in TURNIP;VIEW 1 and VIEW 2 (the old and new versions of that file) shows up as a small letter I with circumflex (code 238), which is out of range for the ITS 7-bit character set, on my local machine.  If I go to Eric's machine (which may not be true to these sources), it shows up as a small letter n. (which might be due to incorrect processing of a newline [control-j] character as \n, somehow losing the slash.  In any case, it's not the right character for the desired sequence.  it should be control-J, and this could be the reason that some of the JCL is not getting that character removed.  

I downloaded the file from github (which I hope happens in binary mode, since it thinks it's a binary file) and opened the file in Common Lisp in (unsigned-byte 8) format and read up to that character. These are a few of the preceding characters, starting with the start of that line:
```(27 32 104 107 32 102 106 32 122 106 32 45 64 102 2 13 238)```
Here is what they decode to:
```
(mapcar 'code-char *)
(#\Escape #\Space #\h #\k #\Space #\f #\j #\Space #\z #\j #\Space #\- #\@ #\f
 #\Stx #\Return #\LATIN_SMALL_LETTER_I_WITH_CIRCUMFLEX)
```

Now the fact of this odd character is probably just how my system is perceiving it, thinking it's unicode when it's not. Maybe newline is represented by 238. I note that lopping off the high order bits and just looking at this as a 7-bit quantity shows me:
```
(logand 127 238)
110
(code-char 110)
#\n
```
It is possible you're using a Newline character as something high up in the 8-bit ascii space, but I don't think TECO is going to like 8-bit ASCII. ITS teco was based on 7-bit ascii, packed 5 to a 36-bit word with one bit left over. It expects to see control-J for newline (10) and control-M for carriage return (13).  If you are having trouble with the newline not being stripped, it may be because it is looking for 238 (decimal) and not for 13 like it's supposed to.

(It's also probably a bad idea to follow control-M with control-J because various processors will try to optimize that with special CRLF handling. I would suggest it would be better to do control-J followed by control-M as happens elsewhere in this file.)

To verify that this was not just my own transfer issue, I went onto Eric's ITS emulator and pulled in the files VIEW 1 and VIEW 2 from the TURNIP directory into an E (emacs) job.  I did control-U 1 7 2 0 control-F to get to the position in question and did control-X = to see in the echo area what character I was looking at. In his case, it was a small letter n, which is "coincidentally" the character that would show up if you masked 238 to 7 bits but which is NOT the same as Control-J (10).

So something worries me here still.

**Comment by eswenson1** [comment]
I just looked at the file I received from the TURNIP; VIEW 1 file I retrieved from MC backup tapes, and can confirm that at position 1721, the character there is 238 decimal, #EE (hex), and 356 octal. When an ITS backup tape was created with this file on it (by itstar) and this was loaded onto ITS (ES or DB), the character at this position becomes a lower-case "n" (octal 156).  The binary for 356 is 011101110 and the binary for 156 is 001101110, so as you can see, itstar is stripping the high bit of the 8-bit character. 

Now, we've had other cases of files that were retrieved from MC backup tapes and loaded onto to Unix machines and then converted to Unix (binary) files where characters have been damaged.  I suspect the original TURNIP; VIEW 1 did not have a 356o character there -- and most likely not, either, a 156o character there. And I've seen cases where a newline characters have been changed to "n" characters before. Perhaps Lars understands why this has happened (I don't).

In any case, it does seem that the original file we got retrieved did have this 238d/356o character there, so it is likely that it was damaged on the host that processed the tape image (some machine at MIT).

**Comment by eswenson1** [comment]
Maybe we should (temporarily) revert our changes to VIEW 1 (which, on ES, I've called VIEW 2) and replace the bad 238 character with a 10 and see if the JCL parsing works.  

Actually, I just tried that and the resulting TS VIEW didn't work.  It gets the same errors as before:

```
file "DSK: .INFO.; TECO ORDER
" not found?
 EMACS◊J
*
```

**Comment by larsbrinkhoff** [comment]
Thank you both for your close scrutiny!

There is an explanation, but possibly an error in execution.  The key is to understand that the files stored outside ITS are encoded in the "evacuate format" that I believe was invented by Alan Bawden.  This encoding has the property that ASCII text files are readable both on ITS on Unix, but also CRLFs are translated to single LFs on the Unix side.  Furthermore, binary files are encoded in a lossless way.

I'll spare you the details, but a single CR in an ITS file is encoded as octal 356 on the Unix side, and a single LF as a CR.

**Comment by larsbrinkhoff** [comment]
I have made a full build from scratch, so I'm checking the result now.  The intent was to copy from line 43 in VIEW 1.

In EMACS, I see this on line 11 in KMP TECO:

```
  0,.k [1 zj -@f^B
^C^@^_$k hfx1 et KMP; DSK: FOO START $ et ^]1 $ ]1
```

The line break ought to be the CR and LF.  The new TS VIEW seems to work, except the file is not displayed immediately when VIEW has started.  Maybe same problem we had with INFO?

**Comment by larsbrinkhoff** [comment]
Right, transferring these files with FTP will probably not work in all cases.

I'm reversing the order of the characters.  The CR will still show up as î, but it will not look like a line break in EMACS.

I'll make a full rebuild again, to make sure the end result is OK.

**Comment by eswenson1** [comment]
@larsbrinkhoff Did you remember to create a link for TS VDIR => TS VIEW?  VIEW checks for a "D" in the JNAME and historically, I believe the FN2 used was VDIR, and if found, it expects to view a directory rather than a file.  I didn't see this in your emacs.tcl changes.

**Comment by netsettler** [comment]
I don't really mind changing this data in any way that works, I just wanted to highlight that the problem might be just the format of the file, and not some compatibility issue with another software program. Now that we understand the issue, I'm susceptible to any plausible plan for moving forward. You guys have been doing this for a while and probably have a sense of how to balance history and just getting things working.

**Comment by larsbrinkhoff** [comment]
Thanks Eric, I was about to forget that.

Kent, the way we do this is usually to commit the orginal files first, then do the udpates in later commits.  If a file has a version number, we increase it.  Most updates are minor fixes to get things working, but we don't shy away from major enhancements as long as they are in the (perceieved) spirit of ITS.

**Comment by eswenson1** [comment]
I'm wondering if the need to add the fix to TURNIP; VIEW 1 is due to the fact that between TECO 1171 and TECO 1212, long filename support was added to TECO. This might have changed how filenames in the JCL were processed. Otherwise, we still do have a bit of a mystery, since Kent's code clearly must have worked back in 12/1983, when VIEW 1 was created.  The long filename support was added to TECO in 1985 sometime, as far as I can tell.  The existing TS VIEW, of course, would have continued to work, being based on TECPUR 1171.  But new builds of it using TECO 1212 would have the JCL parsing problem.  Thoughts?

**Comment by eswenson1** [comment]
Doing an SRCCOM (on ITS) of VIEW 1 versus VIEW 2, we have:

```
*:srccom turnip;view 1,view 2

;COMPARISON OF DSK:TURNIP;VIEW 1 AND DSK:TURNIP;VIEW 2
;OPTIONS ARE    /3

**** FILE DSK:TURNIP;VIEW 1, 1-67 (2636)
**** FILE DSK:TURNIP;VIEW 2, 1-67 (2636)
zj -@f^B
^C^@^_$k                    !* Remove troublesome characters    !
***************
:KILL
*
```

Of course, by the time Lars merges this to master, he may have swapped the ^J and ^M that shows above as just a newline between the ^B and ^C.

**Comment by netsettler** [comment]
Eric, I downloaded and did a comparison with m-X Compare Windows in Emacs 24, and got the same difference, so yes, I believe that. And I'm OK with these changes in spite of the encoding issues, I just wanted to make sure we did it knowingly.

**Comment by larsbrinkhoff** [comment]
Thanks, Kent.  You're right to be cautions.  We do want to handle these files with care.

I have now rebuilt everything from scratch with the latest updates.  VIEW and VDIR seem to work as intended.  Line 65 in TURNIP; VIEW 2 looks like this in EMACS:

```
zj -@f^B^J^M^C^@^_$k                !* Remove troublesome characters    !
```

I added the TCTYP change around dumping TS VIEW, and now it displays correctly.  That's an interesting bug I suppose, but for now I'm content with this as is.

netsettler:
Lars, that's not what I see when I download view.2 from the most recent commit that has it (which is not recent) nor from the PR "Files changed" list just in this moment. In both cases, I see a linefeed after the control-M, not before.

teco screen shot 2018-11-13 at 9 31 45 am

 I'm thinking you made the change locally but have not uploaded your fix.

netsettler:
Sorry, for clarity, that's a compare windows with view.1 in the top window and view.2 in the bottom window.

larsbrinkhoff:
Yes, that's the expected result when viewing the host ("Unix") file.  A ^M here will become a ^J in ITS, and the î becomes a ^M.

The line I pasted in comes from ITS EMACS.  Once everyone agrees everything is in order, I'll make sure my local commit ID is the same as the one displayed here on GitHub, and that I don't have any local edits pending.  Then I'll do a full rebuild and check things in ITS.  Better be extra careful with TECO code, it's not my forte.

netsettler:
I see. I missed the detail where those got swapped. Ugh. OK, this looks good.

> Yes, that's the expected result when viewing the host ("Unix") file.  A ^M
> here will become a ^J in ITS, and the Ã® becomes a ^M.
>
> The line I pasted in comes from ITS EMACS.  Once everyone agrees
> everything is in order, I'll make sure my local commit ID is the same as
> the one displayed here on GitHub, and that I don't have any local edits
> pending.  Then I'll do a full rebuild and check things in ITS.  Better be
> extra careful with TECO code, it's not my forte.

You seem to be doing quite well under pretty extreme circumstances.

larsbrinkhoff:
I did the checks I promised.

PR #1394, opened 2018-11-14
Build CGOL FASL from source.

### LINKED ISSUE #1285: LISP; CGOL FASL
Needs to get compiled from source. Source is currently missing. May have found it, but need to check.

eswenson1:
Resolves #1285.

### DIFF
--- a/build/lisp.tcl
+++ b/build/lisp.tcl
@@ -508,6 +508,12 @@ respond "_" "liblsp;_pratt;cgrub\r"
 respond "_" "\032"
 type ":kill\r"
 
+# compile cgol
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
Is the CGOL FASL still needed for bootstrapping?

eswenson1:
In general, I've kept all the bootstrapped LISP; * FASL files even after I've found sources and compiled them. This is because sometimes there are circular dependencies and it isn't really possible to compile from source.  However, CGOL is unlikely to be used by anything required to bootstrap -- but until we know this for sure, we shouldn't remove it.

Issue #1027, opened 2018-06-27
DM demons


larsbrinkhoff:
The DEMSIG documentation explains the DM demons:

```
        A demon is a program which runs (and perhaps even exists)
        only when a request has been signaled for it.  If the
        demon with the specified demon is not currently running,
        the system job creates a demon job and loads it from the
        file SYS:ATSIGN <name>, where <name> is the name of the
        demon.
        When the demon runs, it can acknowledge receipt of one
        or more signals by using the .DEMON uuo (q.v.).
        The status of a demon can be examined and altered by
        using the RDDMST and STDMST symbolic system calls.
        When the system is started up, there is a once-only
        signal for the DEMSTR demon pending; this demon normally
        signals requests for all other standard systemic demons.

        Demons were much used on the DM machine for various purposes.
        Standard demons which are started by the DEMSTR demon include
        (as of July 23, 1975):
        NAME    REPT RATE       PURPOSE
        TCTYP   0               Set terminal types?? (Doesn't exist on SYS)
        UNSPOO  0               Line printer unspooler (TPL).
        SURVEY  10.             Network survey generator.
        LDRDEM  0               Loader demon (obsolete).
        BATCHM  0               Old batch monitor (obsolete).
        COMSYS  0               Old message demon (obsolete).
        NETRJS  0               CCN remote job server demon.
        ZONE    0               MUDDLE compile batcher.
        MG      0               Maze Guncher - kills games of MAZE
                                in an obscure way.  AV doesn't like them.
        COMDMN  0               Communications demon (replaces COMSYS).
        BATCHN  0               Batch monitor (replaces BATCHM).
        RFC402  0               Message archival demon.

larsbrinkhoff:
Links to:
- UNSPOO: #1026
- SURVEY: #762 
- LDRDEM: #1025 
- BATCHN: #235

larsbrinkhoff:
DEMST (#1028) reads internal ITS data structures and prints out the status of demons.

larsbrinkhoff:
DEMSTR has some more demons:
- EGG - Puff the magic dragon.
- PAPSAV - Spyer on T00.  Paper saver; console not connected to paper terminal?
- COMSAT - Mail server (#34).
- GUNNER - Garbage man (#233).
- EXPIRE - GMSGS cleaner.

larsbrinkhoff:
RJE protocol probably used by NETRJS: https://tools.ietf.org/html/rfc407

[...]

Issue #235, opened 2016-12-12
BATCHN - Batch monitor


larsbrinkhoff:
DM batch monitor.   Replaced earlier BATCHM.  There doesn't seem to be any source or binary.

larsbrinkhoff:
Maybe DMCG; MONIT 200 and DMCG; TS MONIT.

larsbrinkhoff:
No, not MONIT.

larsbrinkhoff:
Maybe .BATCH; BATCHN UJM 149 on TID 7005366.

larsbrinkhoff:
And .BATCH; BATCHN UJM150 on ToTS 9006235.

There are both .BATCH; BATCHN SAVE and NBATCH SAVE files.

larsbrinkhoff:
```
<PACKAGE "NBATCH">
;"NEW BATCH INTERFACE (MUDCAL) by S. W. Galley (SWG) and Jerry Morrison (JHM)
-- TALKS DIRECTLY TO DISK QUEUE, BY WRITING OUT SPECS AND MODS FILES IN .BATCH
DIRECTORY, AND IPC'S THE DEMON OR MODIFIES ITS SCHEDULED AWAKENING TIME."

**Comment by larsbrinkhoff** [comment]
@eswenson1, did you get these files out of ToTS?

**Comment by eswenson1** [comment]
Yes. I haven’t tried to prove that it works yet though.

**Comment by eswenson1** [comment]
SYS.14.12 describes the DMS Batch System.  Found on (among other tapes) 7005366 in the file %SYS;1412 DOC.

**Comment by eswenson1** [comment]
I followed an example in SYS.14.12, and I got an error:

```
CALICO Version MB11 in operation

@create.task◊ named (LINE):  example◊ system program (FILE):  ts muddle◊
terminal input from a file? (SYM):  y◊ which file? (without "s) (FILE):  example input◊
terminal output to file named (without "s) (LINE):  example output◊
type (SYM):  short◊ run at (LIST|ATOM):  t◊
rescheduling interval or sequence (LIST|VECTOR|OBJECT):  ((1 0 0) ())◊
field name (SYM):  print.specifications◊

*ERROR*
NTH-REST-PUT-OUT-OF-RANGE
#WORD *000000005547*
DTNORM!-TIMFCN!-PACKAGE
LISTENING-AT-LEVEL 2 PROCESS 1
@
```

eswenson1:
[...]

PR #1392, opened 2018-11-14
COMBAT

larsbrinkhoff:
<sys.unsupported>combat.mid.151

MIDAS program with ITS conditionals.


larsbrinkhoff:
Wong issue, sorry!

larsbrinkhoff:
According to #1027, there's something called ZONE which is the "MUDDLE compile batcher".  Also BATCHN (#235), the "new batch monitor".

larsbrinkhoff:
The valret is because the  .CALL RDDMST in SSTAT1 fails.  The call tries to read the status for the ZONE demon.  Or course, we don't have ZONE running (#1401).

PR #1393, opened 2018-11-14
MUDCOM

larsbrinkhoff:
<sys.unsupported>mudcom.mid.118

`MUDDLE SRCCOM`  MIDAS program with ITS conditionals.

> `MUDCOM`, a program for comparing versions of a Mdl program. It is used by combat (see section 5.2) to aid in the preparation of compiler plan files. It has several useful aliases.


larsbrinkhoff:
I randomly put this in SYSEN3 and SYS3, but I don't know if that's appropriate.

larsbrinkhoff:
JNAME aliases:

- MUDCHK
- MUDLST
- MUDFND

larsbrinkhoff:
> Mudcom is an assembly language program (not written in Mdl), which nonetheless understands the syntax of Mdl programs. It is used for comparing two versions of the same program, and also (under the name Mudchk) for checking the syntax of Mdl source files more rapidly than they can be loaded into a Mdl.  Mudcom is not interactive; all instructions must be passed on the jcl line.

larsbrinkhoff:
Not to be confused with COMMUD.

larsbrinkhoff:
Now that you mention it, I do.

PR #1404, opened 2018-11-16
MUDINQ - Muddle inquirer

larsbrinkhoff:
Inquires a Muddle job.  When run as MUDINQ, send a Muddle expression to the target, and waits for an answer.

Requires the IPC device.


larsbrinkhoff:
Sample session:

```
*:mudinq PDL MUDDLE

Inquiry: <+ 1 2>◊
ERROR: NO SUCH DEVICE
2323>>XCT 1   1/   .OPEN 3,1355      /   1,,516043   '◊1'  !IPC◊   ◊↑X.

larsbrinkhoff:
Looks like IPC: is enabled when ITS is assembled with MSPP on.

larsbrinkhoff:
I don't know.  It'd like to try it out for a bit.  It could be harmless, other than consuming some memory.

I feel an increasing need to be able to build different configurations.  DM would include Muddle and the DM demons, and IPC.  AI would have the 10-11 interface, the GT40, and a PDP-6.  MC would have Macsyma.  (ML... what?)

The current state is that DB is a superimposition of all ITS machines, and also everything in the time dimension.  There can be conflicts.

larsbrinkhoff:
I see now there's a conflict between the Muddle WHOM and the WHOJ alias.  Maybe the former could be named MUDWHO instead?

larsbrinkhoff:
I added message slurper and DM demon calls to my ITS.  Now IPC opens fine.  But there's no Muddle listening, so I'll have to check the manuals to see if there's something I need to call from Muddle.

larsbrinkhoff:
I'm not so sure about PURGE and MAKSRC.  I think the first sends `; \"Dungeon\" <FLUSH-ME!-INITIAL!- >`, and the second `; \"Dungeon\" <MAKE-SCRIPT>`

@36bit, @atsampson any insights?

PR #1409, opened 2018-11-17
Added emacs pascal mode pl mode and documentation.

### LINKED ISSUE #1303: EMACS Pascal mode
EMACS1; PASCAL 51
EMACS; PASCAL SYN  
EMACS; PASCAL DEFN  
INFO; EPASC 32

eswenson1:
Resolves #1303.


larsbrinkhoff:
FYI, I haven't added all EMACS libraries, because doing so would overflow the directories.  We discussed making EMACS0 and EMACS2 when that happens, but I think maybe the "most important" (for some value thereof) libraries should have priority.

larsbrinkhoff:
There's a comment in PASCAL SYN:

```
;Format for this file is
;described in INFO:EPASC.INFO,
```

That file is available from the EMACS 170 tape, from SV, and from KANSAS tapes.  The one from SV differs from the other.

- https://github.com/PDP-10/emacs/blob/extracted/info/epasc.info
- http://sv.svensson.org/INFO%3bEPASC
- http://www.unlambda.com/download/cadr/kansas.tar.gz

Can you please pick one and add it here?  We can ponder later which version is best, or whether to merge the two different versions into one.

larsbrinkhoff:
I had the wrong tarball above.  Even though that one is very interesting.

I meant one of these: http://unlambda.com/download/mit/

eswenson1:
I couldn't find in http://unlambda.com/download/mit/ where there was an info/epasc.info.  Can you be more specific? I looked in all of the tarballs for an "info" directory.  I also looked in http://www.unlambda.com/download/cadr/kansas.tar.gz

eswenson1:
Of the two other pointers you provided, the one at https://github.com/PDP-10/emacs/blob/extracted/info/epasc.info appears to be newer than the one at http://sv.svensson.org/INFO%3bEPASC.  I'll commit the first one.  If we find a better version, we can always update.

eswenson1:
I extracted the kansas-orig.tar.gz and ran the extract.sh and found the three (identical) copies of epasc.info. I compared that version with the two others you referenced above. The one I committed to my ejs/ticket_1303 branch is the one that seemed the most up-to-date and complete.  And that was the one from the EMACS 170 tape.

eswenson1:
By the way, my extract of the kansas-orig.tar.gz stuff didn't proceed without errors/warnings.  Does it do the same for you?  Here is what I got:

```
eswenson@eswenson-ThinkPad-W520:~/ITS/tapes/kansas-orig$ ./unpack.sh
uncompress 606
uncompress 671
606:
?Unexpected end of file
-rw-rw-r-- 1 eswenson eswenson       320 Nov 17 21:35 file001
-rw-rw-r-- 1 eswenson eswenson 161657440 Nov 17 21:35 file002
ERROR - Date earlier than Jan 1,1970!!!
ERROR - Date earlier than Jan 1,1970!!!
ERROR - Date earlier than Jan 1,1970!!!
ERROR - Date earlier than Jan 1,1970!!!
671:
?Unexpected end of file
-rw-rw-r-- 1 eswenson eswenson       320 Nov 17 21:36 file001
-rw-rw-r-- 1 eswenson eswenson 162134000 Nov 17 21:36 file002
-rw-rw-r-- 1 eswenson eswenson    714811 Nov 17 21:35 files.606
ERROR - Date earlier than Jan 1,1970!!!
gz/temp/temp.txt: Permission denied
ERROR - Date earlier than Jan 1,1970!!!
eswenson@eswenson-ThinkPad-W520:~/ITS/tapes/kansas-orig$
```

larsbrinkhoff:
I squashed your two commit together, I hope that's ok.

PR #1423, opened 2018-11-26
Documentation for the ML CH10 Chaosnet interface.

larsbrinkhoff:
```
   It seems more reasonable for ML to have an interface on it directly
than to use e.g. a crufty DL10.  Maybe other 10s want to go direct, also?

   What needs to be done to attach a caios net interface directly
to a pdp10 I/O bus, rather than via an 11.


larsbrinkhoff:
@rcornwell, read this.

larsbrinkhoff:
Hello @mcallester,

I see the initials DAM here: https://github.com/PDP-10/its/pull/1423/commits/5b7a16bae6923eaede8f09f7969fb31a4ffc4003

Did you design the Chaosnet device for the MIT ML PDP-10?

mcallester:
1976 is a couple years before I was DAM.  Maybe David Moon?

PR #1426, opened 2018-11-28
DMARCD - old ARCDEV

larsbrinkhoff:
The binary file DEVICE; OARCDV BIN was built from SYSEN2; DMARCD 30.

Before knowing about this, I reconstructed ARCDEV 66 from OARCDV, copying source text from ARCDEV 23.  Also since then, the original ARCDEV 66 has turned up, but it's not the same as my reconstruction.


larsbrinkhoff:
Checked binary identical with SRCCOM /$.

Issue #1431, opened 2018-11-30
TV wholine Y2K bug


larsbrinkhoff:
Setting the date to 2018 says 118 in the wholine.

larsbrinkhoff:
Fixed by #1432.

PR #1432, opened 2018-11-30
Fix TV wholine Y2K bug.

larsbrinkhoff:
2018 shows up as 118.  The simplest fix, and consistent with previous updates, is to add 1900 to get a four digit year.


larsbrinkhoff:
I'm holding off with the reviews a bit.  I'll check the TV-11 source code to see if there are is more date handling.

Also, I'm not sure about the system wholine versus a user defined one.  Is the TV-11 built in wholine the system wholine?  What happens with a user defined wholine?

larsbrinkhoff:
Resolves #1431.

larsbrinkhoff:
I'm not sure if it's best to try to find some more bits for the year, or make it really be mod 100.

larsbrinkhoff:
ITS' sysjob sends the left half of QDATE to the TV-11, losing the top two bits in the process.  To make this work beyond 2027, we'll have to fix ITS to send the year modulo 100.

eswenson1:
@atsampson isn’t the IDIVI instruction going to update the D register with the result of the integer division? Why are we storing the contents of E?

eswenson1:
Never mind. I looked up IDIVI and it seems remainder is put in E.  Sorry for not checking first.

larsbrinkhoff:
Thanks for the close scrutiny, Eric!

PR #1229, opened 2018-12-02
TVFIX bugs

larsbrinkhoff:
There are two problems with the TVFIX code in LOCK:

- First, the ^M in the host source file is translated to a ^J in ITS.  The host character should be octal 356 instead.
- Second, it still doesn't work since it's started from TARAKA.  So when TVFIX does .VALUE, it does to TARAKA, not DDT.  And TARAKA doesn't accept the request.


larsbrinkhoff:
It could also first check if TT11P already is 1.  If so, no need to stuff.

larsbrinkhoff:
Sometimes I get this:

```
TOP LEVEL INTERRUPT 400 DETACHED JOB # 7, USR:TARAKA TVFIX
```

So it's not working every time.

larsbrinkhoff:
Oh, just the tv11 emulator that wasn't connected.

larsbrinkhoff:
It was I who added it in the first place, so I'm not removing code from back then.  It worked fine when I tested it from DDT, of course.

PR #1418, opened 2018-12-02
CLOCK - Knight TV clock

larsbrinkhoff:
KLH; CLOCK 84  
KLH; DUUOS 38  
KSC; MSTAT > *missing*


larsbrinkhoff:
Big analogue clock.
big

larsbrinkhoff:
Small version in upper right corner.  Runs in the background.
tiny

larsbrinkhoff:
Or digital.
digi

larsbrinkhoff:
We don't have a policy, so it's open for discussion.

My thinking is that programs.md should list all programs.  It's not entirely clear exactly what a program is.  We don't have individual EMACS libraries listed, for example.

Another thought is that not all programs must be in SYS*n*.  They can still be called with the :*FOO*;*BAR* syntax.  We already have ALAN;GENSYM, C;CC, BKPH;ZAP, KSHACK;MICRO, GAMES;CHESS, BBN;LOGO, AAP;FORTH, etc.  Some of those are historically correct, others are recent inventions.

eswenson1:
I wonder if it would be a good convention to specify how to invoke the program (e.g. :alan:gensym) if the program doesn't live in SYS:

larsbrinkhoff:
Good idea.  I think that would be an improvement.

larsbrinkhoff:
I'm weary of adding these half crappy names I came up with, but I don't want this to linger on.  I'll make a note that they can be reconsidered.

eswenson1:
Or just include (in parens) the install directory? Maybe anyone using ITS will be smart enough to use FIND to find programs that don’t just work by virtue of being in SYS:.

larsbrinkhoff:
The big watch should be SYS1; TS CLOCK.

larsbrinkhoff:
The small clock should be TS WATCH.

PR #515, opened 2018-12-05
GUNNER - gun down jobs

larsbrinkhoff:
DEVICE; JOBDEV SHOE  
AR2: RWK; GUNNER 1

The source file is titled "GUN: JOBDEV", but the installed file is for SHOE:

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -669,6 +669,10 @@ expect ":KILL"
 respond "*" ":link device;jobdev hs,device;jobdev hsname\r"
 respond "*" ":link device;jobdev hf,device;jobdev hsname\r"
 
+# gunner
+respond "*" ":midas device; jobdev shoe_rwk; gunner\r"
+expect ":KILL"
+
 # pr
 respond "*" ":midas sys1;ts pr_sysen1;pr\r"
 expect ":KILL"
--- a/doc/programs.md
+++ b/doc/programs.md
@@ -112,6 +112,7 @@
 - GO, the Go board game.
 - GOTO, go to specified terminal line.
 - GUESS, a very silly game.
+- GUNNER, gun down jobs.
 - GMSGS, copy system messages to mail file.
 - H3MAKE, a job that requests DRAGON to build host table.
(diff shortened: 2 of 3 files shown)


BobKerns:
I don't think I can really add much.

Version 31 seems a lot more plausible than 1—I'm sure version 1 didn't even assemble.

I'm amused seeing the term "PDL" (Push Down List) in place of the term "stack". I'd pretty much forgotten the term existed.

SHOE was crypto-mnemonic for SHOO, as in "go away".

I'm pretty sure DDT had a way to kill a foreign job, so creating this as a JOBDEV was probably for the purpose of remote usage from the onset. But I ended up using it locally due to convenience, with the code finding the HACTRN or TELSER process.

I'm not sure if this was the first JOBDEV I wrote.

I may have written the first network print server. It was a rasterizer for the XGP, that did all the work in one process if ran on AI, but split and did a sort of RPC if ran elsewhere, with the per-scan-line stuff happening on AI, while the document parsing and font handling done on the client side. I doubt the source survived, given the retirement of the XGP after the Dover arrived.

As for foreign jobs and DDT: I added the ability to do $$^R to a foreign job after you attached it, in response to some hard-to-discourage troublesome net random.

I added the code initially by self-patching my own live DDT process. Not without a few missteps, but none that crashed my DDT. So I reverse-engineered my patches-to-patches-to-patches and added the code to the source.

The purpose was to avoid engaging the random as a human, but to convince him the system was too flakey to bother with. Without any form of security, we dealt with bad actors through psychology. Most people would stop or leave if asked, or if that didn't work, go away if gunned down a few times, but this guy was unusually persistent.

Mostly I hacked the prompting code in his DDT, inserting .sleep calls to simulate a slow system, and random error messages.

It did work.

larsbrinkhoff:
Thank you @BobKerns, I think that was a valuable addition!  I always thought SHOE was about kicking someone off.

Also interesting about $$^R.  I have seen the code but I didn't consider that it was for foreign jobs.  I'll go test that right away.

PR #117, opened 2018-12-06
LINE - displays mode line with date/time/uname/hsname/etc

larsbrinkhoff:
SYS2; TS LINE
source missing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ SRC = syseng sysen1 sysen2 sysen3 sysnet kshack dragon channa	\
       draw wl taa tj6 budd sharem ucode rvb kldcp math as imsrc gls demo \
       macsym lmcons dmcg hack hibou agb gt40 rug maeda ms kle aap common \
       fonts zork 11logo kmp info aplogo bkph bbn pdp11 chsncp sca music1 \
-      moon teach ken lmio1 llogo chsgtv clib sys3 lmio turnip mits_s
+      moon teach ken lmio1 llogo chsgtv clib sys3 lmio turnip mits_s rab
 DOC = info _info_ sysdoc sysnet syshst kshack _teco_ emacs emacs1 c kcc \
       chprog sail draw wl pc tj6 share _glpr_ _xgpr_ inquir mudman system \
       xfont maxout ucode moon acount alan channa fonts games graphs humor \
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -940,6 +940,10 @@ respond "*" ":link sys3;ts vrfy,sys3;ts expn\r"
 respond "*" ":midas sys2;ts wholin_sysen2;wholin\r"
 expect ":KILL"
 
+# LINE
+respond "*" ":midas sys2;ts line_rab;line\r"
+expect ":KILL"
+
 # WHOIML
 respond "*" ":midas sysbin;_sysen1; whoiml\r"
 respond "FILE:" "whoiml\r"
--- a/doc/programs.md
+++ b/doc/programs.md
@@ -148,6 +148,7 @@
 - KN10, KS10 frontend software.
 - LAY, SUDS layout program.
 - LD10, Lisp display slave.
+- LINE, display wholine.
 - LINK-10, DEC linker.
 - LIMERI, print limericks.
 - LIMSER, Chaosnet limerick service.
--- a/src/rab/line.2
+++ b/src/rab/line.2
@@ -1,0 +1,504 @@
+	title WHOLINE display
+	subttl Definitions
+
+; Reconstructed 2018 from disassembly of SYS2; TS LINE dated
+; 1977-09-05, by editing SYSEN2; WHOLIN 53 dated 1977-12-16.
+
+; I make no claims that this crock is at all well-written.
+
+.insrt MRC;MACROS
+
+; Assembly switches
+
+nd. slptim==5.				; seconds of sleep time
+
+; Accumulators
+
+acdef. [x y z v u w b c j t l]
+
+acdef. [wrd chr unm jnm f sp d e ptr str]
+
+; I/O channels
+
+acdef. [tto]
+.hkill tto
+
+define type string
+ irpc char,foo,[string]
+  movx x,<"char>
+  chrout x
+ termin termin
+
+;  No attempt is made to optimize the AOS'ing of the character count, since
+; this way it is more straightforward and anybody who cares about a few extra
+; words or a couple of usecs saved is obviously losing!
+
+define chrout ac
+ idpb ac,b
+ addx c,1
+termin
+
+line:	block 20.			; who-line block
+
+	subttl Main program
+
+; Start of program
+
+wholin:	.open tto,[%tjdis\.uao,,'TTY]	; get a TTY output channel
+	 .lose				; failed
+	movx x,%zeros			; initialize page counter
+	syscal CORBLK,[	clarg. (%cbred)	; get read access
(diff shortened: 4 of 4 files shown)


larsbrinkhoff:
This assembles to a binary identical to the original SYS2; TS LINE.

There was a file ACW; LINE 2 from December 1977.  This is close to the TS LINE timestamp, which agrees with the value of the %ASDAT symbol.

larsbrinkhoff:
@leighklotz and @cstacy wondered about this program.

> Date: 2 March 1984 12:58-EST  
> From: CSTACY
>
> > Date: 1 March 1984 22:29-EST
> > From: KLOTZ
>> 
> > WHAT IS SYS2;TS LINE?  It was written 9/5/77, and just says it's
> > a PDUMP file.  I did :LINE instead of :LINK.  It printed
> > Guess What!!??
> > Your program has just gone off into HYPERSPACE!^G
> > Of course, it was disowned.
> 
> It's appears to be a wholine program of some sort, but I don't know
> what the message means or where the source is.

leighklotz:
I'm not sure :LINE ever worked for me.  I remember using :WHOLIN though.
Looking at the "hyperspace" error it seems like it gives that if it can't get a TTY or if it can't process something about the JCL, but I'm a bit rusty. Certainly I gave it no args, but it's also possible that I ran this under a CRTSTY, if that caused TTY problems.

larsbrinkhoff:
Agreed, it'll go to hyperspace if it's upset in any way.  I think it would have been better to check the inputs first, before proceeding itself into the background.

It works for me without arguments, when I'm logged in through SUPDUP.  I tried to pass in a few different things on the JCL, but no luck so far.  I haven't checked in detail, but it looks like it wants a UNAME and a JNAME, or something like that.

cstacy:
I can't remember why you have to run WHOLIN; doesn't CRTSTY have it 
built in?
Or was the WHOLIN just for regular display terminals?

How does it work to output atomically to the wholine area without 
controlling
all output that is going to the screen?  (Does it automagically give and 
take
the TTY away from the other jobs?)  I can see how tricking the terminal 
(support)
into thinking it has one less  line than really exists, but how does the 
output
actually get onto the screen?


I suppose I could go read the source for WHOLIN and see what it really does.
But does anyone remember how it works offhand?

larsbrinkhoff:
Yes, CRTSTY has a wholine built in.  I didn't know that.  I rarely use CRTSTY though; I either log in with SUPDUP, or use :TCTYP AAA with a VT100 emulator.

WHOLIN isn't very smart, it just writes to the bottom line.  If some other program writes there, the output will get clobbered.

I just see one call to get an output channel:  

    .open tto,[%tjdis\.uao,,'TTY]

The wholine is written using SIOT on the tto channel.  It does check the CNSL user variable first; if there's no console available it will not output anything.

cstacy:
On 12/7/18 11:15 AM, Lars Brinkhoff wrote:
 >
 > WHOLIN isn't very smart, it just writes to the bottom line. If some 
other program writes there, the output will get clobbered.
 >
 > I just see one call to get an output channel:
 > .open tto,[%tjdis\.uao,,'TTY]
 >
 > The wholine is written using SIOT on the tto channel. It does check 
the CNSL user variable first; if there's no console available it will 
not output anything.

But suppose the SIOT is interrupted, and meanwhile some other program
writes anything at all.   What keeps the output buffer from mixing the
contents of the two partially done SIOTs?    One program outputs half
of a control sequence, and the next program outputs an ordinary character
that would be misinterpreted as the control argument (e.g cursor address).
That's for maximum fun; just mixing ordinary characters would be chaos 
enough.

I think the above would actually happen, if the wholine was implemented by
using control codes to address the bottom of top of the screen.

But the way it really works is that the WHOLIN program creates an "echo 
area"
and opens an output channel with the %TJECH mode bit set.
That directs all that channel's output to the echo area, which is maintained
separately by the system.  (Multiple programs trying to write into the
echo area at once would still be chaos, though, I suppose.)

I don't think CRTSTY uses the echo area, though.  It has full control over
the output buffer, so it can do an atomic output.  CRTSTY can even put its
wholine at the top of the screen rather than the bottom!

For anyone following this discussion who doesn't know about "wholine",
it's a dedicated line at the bottom of the screen that shows real-time
status information (such as the UNAME, which JNAME Is running,
whether IO is being done, IO operation progress, whether you are
paged out at this instant, system load info, etc.)

I always thought the wholine feature came from the Knight TV system,
because it could do it "for free" by peeking into the user variables
on the PDP-10.  (On a regular terminal, the user has to run some
job such as CRTSTY or WHOLIN to generate the wholine information
to be shown in the echo area.  And jobs and channels are expensive.)
Whereas the TV wholine is generated on the PDP-11: that line on the
screen is not even visible to ITS. (I think the echo area works just the
same when you're on a TV; just not used for wholines.
Unless you want two of them!)

There's not much special about %TNTV in TS3TTY.
There is the PDP-11 program in SYSTEM;TV.

The WHOLIN system call on ITS has something to do with this,
but I don't remember what it does at all.

The wholine feature also existed on the Lisp Machine operating system.
Pretty sure it all must have originated at SAIL.

I had forgotten about the separate "echo area" on ITS!

It's basically a separate output buffer ("piece of paper") for each
TTY I spent a few minutes looking at TS3TTY but am not familiar with it.
Somewhere in there it must compute the cursor addressing for the echo area
(screen lines minus the echo lines) and put the appropriate control
code for each piece of paper's vertical addressing.  I'm not even sure
whether the papers are drained separately at the device level
(asynchronously from each other) or whether they are consolidated into
one stream.  But either way, the system can keep the output of the two
pieces of paper separate from each other.  Every TTY is allocated two
pieces of paper: one main and one echo area.  Whether characters are
echoed there (as opposed to the main piece of paper), and whether you
can write junk (such as a wholine) there is determined by the SCML
("Set Command Lines") system call that configures it.
"Echo area", "echo region", and "command line (area)"
are all nomenclature for the same thing.

It's called the "echo area" because if you turn it on, you can see
your input characters down there.  This is addressing issues of
latency, and of having a command input area that is distinct
from the main program output.  Think TECO.

Someone...maybe KLOTZ or PGS...understood TS3TTY at one point.

Unrelated to all of this is a feature that we put into ITS at a very 
late date.
There's a software interrupt (%PJJST) that's a sort of job-control thingie
that you can use to request the job should update its progress display.
So if you're sitting there wondering how long your network file
transfer or your gene sequence analysis is going to take, you can
press a magic key (^_J) -- the program can handle the interrupt by
outputting some kind of progress information.  It's totally up to
the program what to do, and it could potentially write something
to the echo area.  I think there was maybe one program that used
this feature.  The idea for this feature came from TOPS-20,
which had ctrl-T, and from sitting around frustrated while
waiting for archival/evacuation FTPs to complete in the
waning days of ITS.  ("Well, on TOPS-20 you could just type ^T
to see how it's going..." "Hmph! We'll see about that!")

larsbrinkhoff:
Thanks, I stand corrected.  WHOLIN is smarter than I thought.

I had reason to look at the TV wholine recently, because it didn't display dates after year 1999 correctly.  I first looked at the PDP-11 code, but since the PDP-10 only sends 7 bits of year I had to fix it there.

larsbrinkhoff:
I asked Bruce Baumgart about wholines at SAIL, and got this:

> the term “wholine” appears in the Triple-I display service source code starting at 1973-12-23
> 
> The Triple-I vector displays predate the Data Disc by years
> 
> The first WHO program source code is dated 1972-10-27 at WHO[CSP,SYS]

PR #1440, opened 2018-12-07
Fix typoed 'DBP' in ITS.

larsbrinkhoff:
Crashed ITS on first use of TV wholine.


larsbrinkhoff:
I think I typed in the #1431 fix manually rather than copying the file from ITS.

PR #451, opened 2018-12-10
NEWDEC - DEC-MAC tape converter

larsbrinkhoff:
SYSEN1; NEWDEC 18

> THIS PROGRAM IS STARTED AT LOCATION GO AND TYPES "?" AFTER WHICH YOU ANSWER WITH A DIGIT 0-8 THE DRIVE (OR 0 IF ALREADY ON DISK) OF A NEW FMT DECTAPE AND WRITES THE ENTIRE TAPE ONTO THE DISK AS FILE .TEMP. .FILE.   IT THEN GROVELS OVER THE CRAP AND PRODUCES UP TO 22 FILES ON THE DISK BY THE SAME NAME AND EXTENSION AS THE ORIGINAL DEC TAPE.  IT ALSO CLEANS THE COB-WEBBS OUT OF THE DISK DRIVE.


larsbrinkhoff:
Thanks.  Added a commit to do so.

PR #255, opened 2018-12-10
Generate Library AUTO-S doesn't work

larsbrinkhoff:
See #254.

Error:

    Bad macro documentation beginning?


larsbrinkhoff:
I'm not sure the timestamp is necessary, but it doesn't hurt.

Maybe consider a solution to #275.

larsbrinkhoff:
Actually, it does hurt because it's not the 1981 version.

larsbrinkhoff:
Successfully generating AUTO-S :EJ looks like this.  "Compressing file" means the COMPRS file is written.  "->" means the :EJ file is written.  The TCL script looks for ":EJ", so it *should* time out if there's a failure.

```
M-X generate Library$emacs;auto-s$emacs1;auto-s
Compressing file DSK: EMACS1; AUTO-S >
 -> DSK: EMACS; AUTO-S ^Q:EJ

PR #609, opened 2018-12-10
Fix PEEK crash

larsbrinkhoff:
I built PEEK in KA ITS without Chaosnet, ran it, and typed K.  It crashed.


larsbrinkhoff:
This looks more complicated than necessary, because apparently J also calls KAOS.

PR #778, opened 2018-12-14
340D - 340 display Datapoint emulator

larsbrinkhoff:
SYSDOC; ITSTTY 41 says:
> For example, the 340D program does a CNSSET to the sty it uses, setting it to be a datapoint with a peculiar screen size.  This causes ITS to send datapoint cursor motion codes to the tty.  The 340D program reads them from the sty and simulates their effect in the display that it maintains on the 340.

.INFO.; TCTYP ORDER says:
> 340D            Declares the terminal to be the 340 datapoint simulator.

If this was a separate program, I can't find it.

But maybe it was enabled by setting `:TCTYP 340D`?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,8 @@ SRC = syseng sysen1 sysen2 sysen3 sysnet kshack dragon channa	\
       draw wl taa tj6 budd sharem ucode rvb kldcp math as imsrc gls demo \
       macsym lmcons dmcg hack hibou agb gt40 rug maeda ms kle aap common \
       fonts zork 11logo kmp info aplogo bkph bbn pdp11 chsncp sca music1 \
-      moon teach ken lmio1 llogo chsgtv clib sys3 lmio turnip mits_s rab
+      moon teach ken lmio1 llogo chsgtv clib sys3 lmio turnip mits_s rab \
+      stan_k
 DOC = info _info_ sysdoc sysnet syshst kshack _teco_ emacs emacs1 c kcc \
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
Almost done with first cut.  Posting a pull request preview.

larsbrinkhoff:
Thanks.  It was fun!

larsbrinkhoff:
LCM has one... except they rebuilt half of it.

larsbrinkhoff:
More from Stan Kugell.  I asked about fitting 73 characters on the display versus 79 as specified by TCTYP.

> Vague recollections... I think you could use 340D using the same size as a Datapoint or more lines/width if wanted. There were probably a number of programs not flexible enough to use different size displays so running the same dimensions as the Datapoints may have been helpful. Maybe it was the default to launch 340D - I don’t recall. Also the slow refresh rate of the 340 had to be considered. Using full width and height, full of text, the refresh rate was painfully slow. Better than using TECO on a TTY, but not optimal either. There are some things it would be difficult to simulate - color for example. The display’s phosphor had a bluish tiny when painted with the electron beam, and a long yellow-green decay. When full of text, the sweep of the beam painting letters from top to bottom, the color and intensity gradient behind the beam, was very noticeable.

larsbrinkhoff:
It seems there was a file called STAN.K; 3406 40.  Worth looking into.

larsbrinkhoff:
3406 40 recovered and found to be the PDP-6 part, see #1600.

PR #1455, opened 2018-12-22
Updated PAPSAV to use SYSMSG macros for maintaining patch table 

### LINKED ISSUE #1454: Update PAPSAV to use SYSMSG macro for patching
SYSMSG uses a clever set of macros to create a patch table and to perform patching of addresses in its code using addresses from ITS. PAPSAV does this less elegantly (manually).  This ticket is for updating SYSEN3; PAPSAV 2 (to SYSEN3; PAPSAV 3) to use the SYSMSG macros.

eswenson1:
Updated PAPSAV to use SYSMSG macros for maintaining patch table and for doing more elegant patching of ITS addresses in the code.

Resolves #1454.

### DIFF
--- a/src/sysen3/papsav.3
+++ b/src/sysen3/papsav.3
@@ -5,8 +5,55 @@
 	; source did. The original PAPSAV shared some code with SYSEN1;SYSMSG > (which
 	; also inserted SYSENG; CALRET >). Where applicable, the source from SYSMSG was
 	; used in this reconstruction.
+	;
+	; EJS (2018-12-21): This version uses macros from SYSMSG to build table
+	; of patches (ABSTB1).
 
 .INSRT SYSENG;CALRET >
+
+;;; DEFINE MACROS SO THAT SYSTEM VARIABLES ACCESSED
+;;; THROUGH ABS PAGES CAN BE REFERENCED IN A NATURAL WAY
+
+DEFINE CONC A,B
+A!B!TERMIN
+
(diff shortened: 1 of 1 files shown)


src/sysen3/papsav.3
```diff
@@ -68,8 +115,8 @@ PROGRAM SYSMSG
 	CALL RITUAL		;ASSURANCE OF PURITY
 
 	MOVEI COUNT, 1		;SET UP COUNT OF MESSAGE SLOTS
-	LSH COUNT, 0		;instruction patched through abstb1 (SYSMLN)
-	movei pt,0              ; instruction patched through abstb1 (SYSMBF)
+	LSH COUNT, SYSMLNG	;instruction patched through abstb1 (SYSMLN)
```
larsbrinkhoff: The comment doesn't match the symbol.

eswenson1: Yeah. Squoze symbols are only 6 chars long. So that last G is irrelevant. I coded my initial version with SYSMLN, which worked fine, but SYSMSG used the 7-character symbol. I could drop the G, or change comment to use the 7-char symbol. Which would you prefer?

eswenson1: I guess i’d prefer to update the comment, since ITS (sysjob) uses the 7-char symbol, even though the G will be ignored.

larsbrinkhoff: Sounds good, then.  Update the comment.

PR #939, opened 2018-12-25
IPLJOB - write 340 display list to plotter

larsbrinkhoff:
SYS; ATSIGN IPL  
SYSEN2; IPLJOB 47  
SYSEN2; AXPLOT 48  
.INFO.; IPL INFO  

When a program opens the IPL: device, ITS loads this program to interpret the display list and draw it on a Calcomp plotter.

[AI memo 109](http://bitsavers.org/pdf/mit/ai/aim/AIM-109.pdf) describes a non timesharing version.


eswenson1:
Actually, I do have a comment. Why, if IPL is placed in the T register prior to calling NUBJST, are you calling the binary ATSIGN IPLJOB. Shouldn’t it be ATSIGN IPL?

larsbrinkhoff:
Good point!  I'll fix that.

PR #498, opened 2018-12-26
TBMOFF - turn binary mode off

larsbrinkhoff:
SYS; TS TBMOFF
AR1: CSTACY; TBMOFF 1

Turn off BINARY mode if on a TAC or TIP.  Good to have if someone wants to set up an old-school Arpanet.


larsbrinkhoff:
Not tested.

PR #1460, opened 2018-12-27
NTS DDT with 340 support

larsbrinkhoff:
This is an old version of SYSTEM; DDT 68.  It only works on a KA10, but the 340 support is functional.


larsbrinkhoff:
`$V` toggles display on/off.  `$U` is supposed to go to "DECDMP" (i.e. MACDMP or DSKDMP), but doesn't.

larsbrinkhoff:
I first tried MIDAS' /T switch to add `.MLLIT=0` rather than editing the source.  But for some reason, that didn't work.

larsbrinkhoff:
Fun fact.  The ITS 724 binary comes with DDT built in, and it does display on the 340.

Issue #1433, opened 2018-12-03
LLOGO 340 display turtle problems


larsbrinkhoff:
LLOGO doesn't work with the 340 display turtle:

```
*llogo^H
!
LISP 2154
LLOGO 1700
DO YOU WANT TO USE THE DISPLAY TURTLE?
<y
DO YOU WANT TO USE THE GT40 RATHER THAN THE 340?
<n
LLOGO LISTENING
?cs
;COULDN'T EVALUATE WIPE NIL
;BECAUSE DISCRIBE IS AN UNDEFINED PROCEDURE
```

I'm guessing this is because LISP; SLAVE FASL isn't loaded.  DISCRIBE is defined there.

larsbrinkhoff:
But even if it's loaded, the 340 slave has a problem:

```
l^H
!
LISP 2154
Alloc? n
*(load "lisp; slave")
;Loading SLAVE 8
127303
(disini)
PDP6 NOT AVAILABLE TRYING PDP10 SLAVE
BADPI;(ILOPR;) 212>>0

larsbrinkhoff:
@eswenson1, do you know how to debug this?  I tried setting breakpoints in SLAVE FASL, but DDT doesn't know about the symbols.  ($^K doesn't help.)

eswenson1:
I don't. But you could assemble SLAVE and produce. a listing, and using two SUPDUP sessions, side-by-side, bring up the listing in emacs in one and debug without symbols in the other. I've done that before and once you have "found" where you are, it is pretty easy to debug.  Are you sure the ILOPR is happening as a result of code in SLAVE FASL, or might the ILOPR be happening in the interpreter/runtime?

larsbrinkhoff:
Thanks, I'll try that.

I'm not sure about anything yet.  "PDP6 NOT AVAILABLE TRYING PDP10 SLAVE" is a message that comes from SLAVE FASL, and the ILOPR comes right after that.

larsbrinkhoff:
I did this to find the start of the string " TRYING PDP10 SLAVE": `6462715156$W`  Then this to find the instruction to print the string: `7000xxxxxx$W` where the x'es are that address.

After printing the string, OPNTEN is called.  It opens an inferior job and loads SYS; ATSIGN 10SLAV.  Shortly after this, I get: `BADPI;(ATTY;)`  I'm not sure what to make of this, or where it's coming from.

larsbrinkhoff:
[...]

PR #1461, opened 2018-12-28
Fix Lisp display slave and LLOGO when using PDP-10.

larsbrinkhoff:
As discussed in #1433.  This moves the code in OP6D2 to a separate subroutine.  OPNTEN is updated to call it.


larsbrinkhoff:
Also fix Lisp Logo TURTLE to explicitly load the SLAVE FASL.  I don't see how it could work otherwise.

Issue #279, opened 2016-12-16
Knight TV emulator


larsbrinkhoff:
*For many years, the primary way to access Al was from the Knight TV terminal system, which was designed by Tom Knight.*

*The Knight TVs are bit-mapped frame buffer displays controlled by a small PDP-11. The TV-11 is connected to the PDP-10 via a special PDP10-to-PDP11 interface which allows the PDP-11's memory to be mapped into the PDP-10. A Knight TV console consists of a CRT display, and a keyboard similar to those on the SAIL TV system used at Stanford.*
-- Getting Started

![](https://hack.org/mc/tmp/knight-console.jpg)
![](http://world.std.com/~jdostale/kbd/Knight1.jpeg)

aap:
Is there any documentation for it? I'd love to try to get it working.

ams:
Is "Knight TV" the same thing as what was used as the TV subsystem on the CADR?

larsbrinkhoff:
I search for AI memos, but came up with nothing.  I think you'd better ask TK himself.

I'll try to help out by reading the monitor sources.

larsbrinkhoff:
@ams That sounds likely.

ams:
If it is the same as the CADR TV system then there are schematics for the board, source code for the TV sub-system.  The CADR TV subsystem is just a memory mapped display.  Here is the code that is used in the MIT CADR simulator (https://svn.common-lisp.net/mit-cadr/trunk/emulator/usim/x11.c)

Schematics: http://bitsavers.informatik.uni-stuttgart.de/pdf/mit/cadr/LM-2/05_TV.pdf

TV subsystem bits: https://svn.common-lisp.net/mit-cadr/trunk/lisp/window/tvdefs.lisp

If it is something different .. then I have no idea. :-)

aap:
The framebuffer would probably be the easiest part in this. There has to be a pdp-11 program to drive the terminals and talk to the 10. Then there is software on the 10 side as well. If he have all the code I'm sure the rest could be reconstructed even if there is no other documentation available. Of course having documentation would be of great help.

larsbrinkhoff:
SYSDOC; .CALLS >

- CNSGET and CSNSET - get or set console parameters
- CORBLK - map TV frame buffer
- T11MP - map PDP-11 memory
- TVWHER - tell where a TV is
- VIDBUF - request/release video buffer
- WHOLIN - get/set TV wholine

SYSDOC; UUOS >

[...]

PR #431, opened 2019-01-03
TVEDIT - paint program

larsbrinkhoff:
SYS2; TS TVEDIT
SYSEN2; TVEDIT 111  
.INFO.; TVEDIT ORDER

The assembly title is TVDRAW, which may be more appropriate.  There was a TOPS-20 text editor called TVEDIT.

Depends on #279.

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -1116,6 +1116,10 @@ expect ":KILL"
 respond "*" ":midas sys2; ts record_sysen2; record\r"
 expect ":KILL"
 
+# TV paint program.
+respond "*" ":midas sys2; ts tvedit_sysen2; tvedit\r"
+expect ":KILL"
+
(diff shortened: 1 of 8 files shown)


larsbrinkhoff:
This program can read the CHANNA; NAMPIC files, which are just binary images of the frame buffer as the PDP-10 sees it.

The TVREAD/TVWRIT programs uses a slightly more dense format, and a header.

larsbrinkhoff:
Original FILEXT,  sources are in .DEMO.; AR9 1.  It has TVDRAW 109, TVEDIT ORDER, and INSRT files.

There used to be a mailing list BUG-TVDRAW, but it's unclear if it refers to the ITS program or to a Lispm program by the same name.  The only recipient in the list is FRD.

PR #1472, opened 2019-01-06
Added new program, UPTINI, to create .temp.;UPTIME DATA. 

### LINKED ISSUE #1471: Add UPTINI program for creating UPTIME DATA
This ticket is for adding the new program, EJS; UPTINI 1 and building EJS; TS UPTINI for creating the file .TEMP.;UPTIME DATA.  A subsequent ticket may call for using this program to build the data file and then copying the result to BAWDEN; UPTIME DATA rather than loading a pre-built one from tape.

eswenson1:
A future commit can run this program, create the file, and copy to bawden
directory rather than using a pre-built version.

Resolves #1471.

### DIFF
--- a/src/ejs/uptini.1
+++ b/src/ejs/uptini.1
@@ -1,0 +1,41 @@
+title UPTINI - Create UPTIME DATA
+
+chdsko=10
+
+flose=.lose %lsfil
+
+data:
+	IRP M,,[DB,NO,UP,ES,LC]
+        sixbit/M/		; sixbit host name
+	0			; sixbit date
+	0			; sixbit time
+	TERMIN
+	0	;; end of data
+dataln==.-data
+
+datptr: -dataln,,data
+
+go:	.call [ setz
+		sixbit /open/
+		[.bio,,chdsko]
+		dbdev
+		dbfn1
+		dbfn2
+		setz dbdir ]
+	 flose
+	.call [ setz
+		sixbit /iot/
+	        %climm,,chdsko
+		setz datptr ]
+	 flose
+	.close chdsko,
+	.logout 1,
+
+dbdev:	sixbit /dsk/
+dbdir:	sixbit /.temp./
+dbfn1:	sixbit /uptime/
+dbfn2:	sixbit /data/
+
+cnstnt:	constants
+
+end go
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -1557,3 +1557,7 @@ expect ":KILL"
 # TBMOFF
 respond "*" ":midas sys; ts tbmoff_cstacy; tbmoff\r"
 expect ":KILL"
+
+# UPTINI
+respond "*" ":midas ejs;ts uptini_uptini\r"
+expect ":KILL"
--- a/doc/programs.md
+++ b/doc/programs.md
@@ -309,6 +309,7 @@
 - UP/DOWN, check if host is online.
 - UPTIME, Chaosnet uptime server.
 - UPTIME, record ITS uptimes.
+- UPTINI, program to create new UPTIME DATA.
 - URUG, GT40 debugger.
 - USQ/TYPESQ, unsqueeze/uncram or type a file.
 - VERSA/SPOOLR, Versatec/Gould printer spooler.


eswenson1:
Current list of hosts added is:  DB,NO,UP,ES,LX

Feel free to suggest changes.

src/ejs/uptini.1
```diff
@@ -0,0 +1,41 @@
...
+		sixbit /open/
+		[.bio,,chdsko]
+		dbdev
+		dbfn1
+		dbfn2
+		setz dbdir ]
+	 flose
+	.call [ setz
+		sixbit /iot/
+	        [%climm,,chdsko]
```
larsbrinkhoff: This doesn't work as intended.  Either `%climm,,chdsko` for an immediate argument, or `[chdsko]`.

src/ejs/uptini.1
```diff
@@ -0,0 +1,41 @@
+title UPTINI - Create UPTIME DATA
+
+chdsko=10
+
+flose=.lose %lsfil
+
+data:
+	IRP M,,[DB,NO,UP,ES,LX]
```
larsbrinkhoff: I don't remember an LX.  Did you mean LC?

src/ejs/uptini.1
eswenson1: Then why does the code work? I think I tried one of the two alternatives you suggested and it didn’t work. Running my code does write the file. 

Why do you think it is wrong? Sure you’re  it thinking of .iot uuo?

eswenson1: I did find this email:

```
Date: 14 January 1983 02:57-EST
From: Alan Bawden <ALAN @ MIT-MC>
Subject:  IOT control bits
To: BUG-ITS @ MIT-MC
cc: GSB @ MIT-MC

This doesn't work as advertized:

	.call [setz ? sixbit /iot/
		[%tipek,,ttyich]
		setzm t]

While this does:

	.call [setz ? sixbit /iot/
		%clbit,,%tipek
		movei ttyich
		setzm t]

We could just fix the documentation of the IOT call to not claim that the
left half of the first argument is XORed with the control bits...
```

which tends to support what you're saying. But when I tested my code, it worked fine.

eswenson1: Since it is slightly more efficient, I changed the [%climm,,chdsko] to %climm,,chdsko, and it still works.  Since this was one of your suggested alternatives, and since I see *NO* code in ITS that does it my way or, for that matter, the %climm,,chdsko way -- everyone does a MOVEI CHDSKO, I'm happy to change it to not use an indirect reference.  But, as I said above, it appears to work fine.

larsbrinkhoff: Sure, `[%climm,,chdsko]` works, just not what you (probably) intended.  From the use of %climm, it looks like you intended an immediate argument.  But to do that you'll need to remove the brackets, because the %climm must be in the argument block.  It works with the bracked because then it's a %clin argument instead and then I suppose ITS takes the channel number from the right half or maybe even just the low 4 bits.

eswenson1: Ok. I did intend an immediate argument. I understand (now) your point. I did update the code to not indirect. What I was confused about was that the code worked fine. Thanks for your %clin explanation — i’ll Read up on it, since i’ve forgotten these details with the long passage of time.

larsbrinkhoff: Of course.  Reading source code doesn't make it more clear, since you can see `foo` or `%clin,,foo` or `move foo` and they all work the same.  Likewise `%climm bar` or `movei bar`.  And then there are everyone's macro packages on top of that.

eswenson1: Also, %CLINN,,FOO is the same as FOO, but MOVE FOO isn't (quite) the same, since MOVE is 200000, and therefore MOVE FOO is 200000,FOO.  Maybe the 2000000 bit will be ignored.  In any case, MOVE obscures what is really intended here.

eswenson1: To get the correct bits for %CLIMM,,FOO, which is what I wanted, I would have had to use MOVEI FOO, which would have a) included the 200000 bit (unwanted) and b) obscured the intent.  MOVE FOO doesn't signify "immediate".

larsbrinkhoff: Beating a dead horse here I suppose, but...  Note that the %CL bits agree with the semantics of the MOVE instructions.  %CLIMM - MOVEI (immediate), %CLIN - MOVE (from memory), %CLOUT - MOVEM (to memory).  I think it's mnemonic.

eswenson1: Sure, but we're not "moving" anything, nor executing any PDP-10 instructions.  We're specify a parameter to a system call.  Obviously, some clever ITS people picked the bits for signifying memory and immediate cleverly.  But clever people aren't always people who are making things easier for others to understand.  %CLIMM,,FOO is the *right* way of doing this.  MOVEI FOO works, but is oblique.

PR #1449, opened 2019-01-06
340D improvements

larsbrinkhoff:
Use Stanford keyboard instead of TTY.  In the simulator, the keyboard input is from the 340 window.

### DIFF
--- a/build/sims/config.202
+++ b/build/sims/config.202
@@ -229,6 +229,7 @@ DEFOPT PTRP==1		;HAS PAPER TAPE
 DEFOPT PTPP==1		; And punch works.
 DEFOPT PDCLKP==1	;"DeCoriolis" CLOCK
 DEFOPT 340P==1		;HAS 340 DISPLAY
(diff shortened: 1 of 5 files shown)


larsbrinkhoff:
Ah, some characters doesn't advance the "cursor".  The table above uses bit 100 to indicate that.

```
        TRNN I,-200
        JRST 340K       ;FAKE OUT 340 NON-SPACING CHRS

larsbrinkhoff:
I'm updating the 340 simulation to add characters as per the new evidence.

I'll have to rework this pull request to make use of them.  I think I'll keep the atsign character.  There's the matter of deciding wether to display old ASCII ↑ ←, or use the incremental mode ^ and _ new ASCII characters.

larsbrinkhoff:
This depends on changes in the KA10 simulator.

larsbrinkhoff:
I think this is ready now.  @eswenson1, any comments?

PR #1476, opened 2019-01-10
If TVFIX runs without a TV-11, exit gracefully.

larsbrinkhoff:
Using .LOSE results in a disowned 10!0 job hanging around.


larsbrinkhoff:
Turns out this isn't enough.  Even if TT11P is present in ITS, it could be that the TV-11 unresponsive and this makes an .IOT fail later.

larsbrinkhoff:
To be more precise, 11HALT is called to tell the user to start the PDP-11.  The .IOT that fails is to print to the TTY, but TVFIX is running without one.

PR #1328, opened 2019-01-10
QSEND

larsbrinkhoff:
EMACS; QSEND :EJ
EJS; QSEND 149  
Z; QSEND 166 *missing*  
Z; QSEND DOC *missing*


larsbrinkhoff:
Let's go with QSEND 149 for now.

larsbrinkhoff:
I'm noting that ToTS has Z;QSEND 166 and QSEND DOC.

eswenson1:
Good find. We should remove the one from EJS (older version) and add QSEND 166 and QSEND DOC to Z.  And update the build to build the emacs package from Z.

PR #1477, opened 2019-01-11
Rearrange SALV and NSALV binaries.

larsbrinkhoff:
(N)SALV BIN is the MIDAS output.  @ (N)SALV is the executable with DDT merged in.  Update dumping of ITS to use SALV BIN.

### DIFF
--- a/build/ka10/include.tcl
+++ b/build/ka10/include.tcl
@@ -59,7 +59,7 @@ proc make_ntsddt {} {
 }
 
 proc make_salv {} {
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
I forgot to write about the context here.  Eric asked me why it wasn't possible to start @ SALV from DSKDMP (by just entering SALV<CR>), whereas it works with @ NSALV.  It turns out that's a mistake on my part.  I thought they should be just MIDAS output without DDT, but when I examined historical @ SALV and @ NSALV files I saw they all had DDT merged in.

PR #774, opened 2019-01-15
Add conditionals for code using the shared I/O bus

larsbrinkhoff:
@philbudne discovered 340 code in ITS which makes use of bit 0 in CONI.  This is a feature of the shared I/O bus, described in Hardware memo # 1.

I think it would make sense to put such code inside IFN DSDP.


larsbrinkhoff:
Work in progress.

larsbrinkhoff:
I tried to capture everything that uses DSDEV or DSDEVN.  And also everything that uses CONI bit 0.  Not sure I got everything yet.

It looks like the build fails when purifying PEEK.

larsbrinkhoff:
Hello @eswenson1, @atsampson, @bictorv, and everyone else,

This one is a bit tricky, so I could use some extra eyeballs on this if you can spare them.

I'll explain the background.  DSDP enables the use of the "deselection device".  That was some machinery to allow the PDP-6 and PDP-10 to share some devices on the I/O bus.  One processor at a time could "select" a device for use, and "deselect" it when done to make it available to the other processor.  Selection was automatically done when a processor issued a CONO to a device.  Deselection is done by sending a special code to the DSDN device.  So far not very complicated.

The tricky bit is that a processor can check the selection status from CONI bit 0, i.e. the sign bit.  It's not always obvious when this happens, because maybe the CONI happens in one place, and the JUMPL/SKIPL somewhere else.  So if you see I made a mistake, or if there's some place I have missed, please let me know!

larsbrinkhoff:
As it turns out, I *have* made some mistake because there's a consistent crash when PEEK is built.

src/system/itsdev.976
```diff
@@ -4726,7 +4744,7 @@ OMPXS:	XCTR XRW,[DATAO OMPX,(C)]	;ZAP CHNL
 	POPJ P,
 
 OMPXCL:	SOSGE OMXNTO
```
atsampson, line 4746: Can that SOSGE skip over the POPJ P,  if DSDP==0?

src/system/itsdev.976
```diff
@@ -4600,7 +4615,7 @@ AIMXW5:	IDIVI A,3
 IMPXCL:	TRNE A,4	;SKIP ON WAS READING AT MP
 IMXCL1:	SOS IMPXF	;WAS READING AT PI
 	SOSGE IMXNTO
```
atsampson, line 4617: As below - can that SOSGE skip over the POPJ P,?

larsbrinkhoff:
Thanks @atsampson, good catch.  The SOSGE instructions are meant to skip over the DATAO when the device is still in use, or deselect it when the counter goes negative.  I updated to use a single SOS in the DSDP==0 case.

larsbrinkhoff:
[ITS hardware memo #1](https://raw.githubusercontent.com/larsbrinkhoff/its-archives/master/ailab/ITS_Hardware_Memo_1.pdf) has a table showing "assignable" devices.  The ones we are concerned with are: DIS, MPX, OMPX, NVD(XYT).  Possibly also RBTCON and TIPDEV, but they don't seem to have any deselect code in ITS.

src/system/itsdev.976
```diff
@@ -2645,12 +2655,15 @@ SUBTTL	340 DISPLAY
...
+IFN DSDP,[
 	CONI DIS,T	;DIS AVAILABLE
 	JUMPN T,ADSDV1	;YES
 	CONO DIS,0
 	CONI DIS,T
 	JUMPE T,ADSDVN	;DIS NOT AVAIABLE
-ADSDV1:	SKIPL T,DISUSR
+ADSDV1:
+]
+	SKIPL T,DISUSR
```
larsbrinkhoff, line 2666: This also checks the CONI bits.

larsbrinkhoff, line 2666: Why store in T here?

larsbrinkhoff:
I think I understand the PEEK "crash".  Before, ITS would check the 340 CONI bit 0.  During the build, the emulator dpy device is not enabled, so CONI would return all zeros.  So ITS would consider the 340 offline.  Hence PEEK could not use the 340 to display, which is the default when running from the console teletype.

With my change, ITS no longer checks the bit so it considers the 340 available always.  So PEEK will probably succeed in trying to open the 340 and thus output is no longer going to the console teletype (which is what the build script sees).

This is supported by the fact the problem only happens on KA10.  I tested running PEEK manually, and I get the same behaviour with a printing terminal (although not with AAA).

larsbrinkhoff:
Next roadblock: 

```
:pdump datdrw; nodips bin
*:go

PDP-6 not available, type "C" to continue without it *CC

The last command timed out.
```

larsbrinkhoff:
Previously we'd see this:

```
PDP-6 not available, type "C" to continue without it *C
Display not available type "C" to continue without it *C
TOP MODE!
```

Now that the 340 display *is* available, NODIPS is somehow stuck and will not output TOP MODE or accept input.  My theory is that it's trying to use the 340.  This used to work way back, but doesn't any more.  So we're kind of stuck here.

I have a workaround in mind, which is to temporarily patch ITS to have the 340 fake being unavailable.

larsbrinkhoff:
wip

But now, finally, it seems to come together.

larsbrinkhoff:
No second thought, rather than working around the problem, it would be better to fix ITS to check if the 340 interface is online.  The implicit assumption is, if 340P is on the 340 is always up and running.  This need not be the case.

build/basics.tcl
```diff
@@ -99,7 +99,7 @@ expect ":KILL"
 respond "*" ":job peek\r"
 respond "*" ":load sysbin;peek bin\r"
 respond "*" "purify\033g"
-respond "Idle" "q"
+respond {$G} "q"
```
eswenson1, line 102: What's "{$G}"?

eswenson1 (approved):
This looks good to me. I assume you've tested with both DSDP=0 and =1, right? I didn't understand the "{$G}" in the PEEK initialization code, but I'm sure once you explain, I'll be ok with it.

build/basics.tcl
larsbrinkhoff: `$G` is the output from DDT.  `{` and `}` is TCL syntax for a string that doesn't do $ variable expansion (but keeps track of balanced braces).

larsbrinkhoff:
Actually, I haven't tested with DSDP==1 because it has been disabled for a long time even on AI (KA10).  So we never tried it with KA ITS.  But it's a good idea; I think it should work.  The CONI bit 0 code was already there and DSDP==1 adds some code to DATAO the device deselection codes to the DSD and DSDN devices which should be harmless.

larsbrinkhoff:
SYSTEM; CONFIG 54 from 1978-06-29 disabled these: PDP6P, 340P, DSDP,  IMXP, OMXP, and HCLKP.  Some of those had some bit rot when we first re-enabled them.

build/basics.tcl
eswenson1: Ah, thanks.  Didn't know that.

larsbrinkhoff:
Now also tested with DSDP=1 and starting TS SPCWAR.

PR #1482, opened 2019-01-17
Make NAMDRG more graceful about missing TV-11

larsbrinkhoff:
@eswenson1 has been testing running NAMDRG with less than stellar results.  It seems the IO channel to the frame buffer accepts data from a SIOT, but fails when FINISH is called to flush the buffer.  Quite possibly, NAMDRG is started too soon, before TVFIX aka STUFF has finished loading.

I suggest to fix this by having NAMDRG check ITS' TT11P first.  If it doesn't exist, ITS is built without TV-11 so just .LOGOUT.  Otherwise ensure it's -1, or else wait until it becomes -1.  I think this should be a more robust behaviour.

### DIFF
--- a/src/sysen2/namdrg.149
+++ b/src/sysen2/namdrg.149
@@ -254,9 +254,9 @@ name:	.close 1,	;close the channel we were loaded on
 	jrst name2
 
 name1:	;char unit out, disp. mode, no continuation lns
-        syscal OPEN,[ [%tjmor+%tjctn+%tjdis+.uao,,tyoc] ? setz b ]
+        syscal OPEN,[ [%tjmor+%tjctn+%tjdis+.uao,,tyoc] ? b ]
 	 .value
-	syscal OPEN,[ [%tjmor+%tjctn+%tjdis+.bao,,tybc] ? setz b ]
+	syscal OPEN,[ [%tjmor+%tjctn+%tjdis+.bao,,tybc] ? b ]
 	 .value
 name2:	.call [	setz ? 'cnsget ? 1000,,tyoc ? 2000,,tcmxv ? 2000,,tcmxh
 		2000,,a ? 2000,,a ? setzm a]
--- a/src/sysen2/namdrg.149
+++ b/src/sysen2/namdrg.149
@@ -226,9 +226,6 @@ init:	setz a,
 
 name:	.close 1,	;close the channel we were loaded on
 	move p,[-lpdl,,pdl-1]
-	.suset [.roption,,a]
-	tlo a,%opint+%opliv+%opopc
-	.suset [.soption,,a]
 	move b,[squoze 0,nf11ty]
 	.eval b,	;TTY # of console free display
 	 .value
@@ -273,6 +270,9 @@ name2:	.call [	setz ? 'cnsget ? 1000,,tyoc ? 2000,,tcmxv ? 2000,,tcmxh
 	pushj p,clrpts	;clear user name ptrs
 	skipl @tt11p	;wait for TV 11 to come up
 	 .hang
+	.suset [.roption,,a]
+	tlo a,%opint+%opliv+%opopc
+	.suset [.soption,,a]
 	pushj p,map11
 	move a,@dmnbd	;find our initial dmnbf ptr
 	movem a,logcnt
--- a/src/sysen2/namdrg.149
+++ b/src/sysen2/namdrg.149
@@ -226,6 +226,9 @@ init:	setz a,
 
 name:	.close 1,	;close the channel we were loaded on
 	move p,[-lpdl,,pdl-1]
+	move b,[squoze 0,tt11p]
+	.eval b,
+	 jsr die
 	move b,[squoze 0,nf11ty]
 	.eval b,	;TTY # of console free display
 	 .value


larsbrinkhoff:
Made this a pull request to show my commits so far.

I tested this three ways:

- ITS without TT11P.  The early .EVAL takes care of this.
- ITS with TT11P but tv11 not started.  NAMDRG stays in the .HANG.
- ITS with TT11P and tv11 started.  The free display comes up.

I don't know how to trigger the problem @eswenson1 has.  Evidence points to tv11 running and talking to ITS well enough to have TT11P set to -1, but maybe later it's not well.

After FINISH fails, is TT11P still -1?  ITS will usually detect if the TV-11 isn't responding within a second, and set TT11P to 0.

larsbrinkhoff:
I got this call to RELOAD now.  Only, the return address points to .CALL FINISH, which means it was the SIOT that failed.

eswenson1:
Hmmm... I thought my return address was after the call to FINISH. But it is possible I’m mistaken.

larsbrinkhoff:
This is from your email:
```
reload/'NAME$: CAM TTYFRC+11 =310000004375
...
TTYFRC+11/ .CALL 41502 (FINISH)
```

So actually it looks the same as mine.

larsbrinkhoff:
I added an %CLERR argument to the SIOT.  I got this:

```
12      %ENSMD          MODE NOT AVAILABLE

**Comment by eswenson1** [comment]
Is it referring to the open modes? Or are there intervening calls to change the console modes prior to the SIOT? I would think that invalid open modes would have resulted in errors at OPEN time or changes modes at CNSSET time. Or it referring to the block mode SIOT and the unit mode open — as a mismatch of modes?

**Comment by larsbrinkhoff** [comment]
I don't know, and I can't find %ENSMD anywhere in ITS except in the BITS file!

```
TTYFRC+7/   .CALL 41501 (SIOT)   :calprt
Arg #1: B   0,,2
Arg #2: B/   ANDCB X,SCREEN   440700,,12131   $1'D' !19$
Arg #3: A/   PATCH+23   0,,123   $1'    !3$
Error return: C/   V   0,,12   $1'     *$
```

Side note: Would be good to have CALPRT decode the error code.

**Comment by eswenson1** [comment]
Agree on the CALPRT enhancement.  We should enter a ticket for that.

I wouldn't be the least bit surprised if the ITS code that returned that error code used "12" or "10." rather than the constant. The code is notorious for using magic numbers!

**Comment by eswenson1** [comment]
A clue:  An old version of itsdev (101) has a reference to %ensmd.  

```
aopen:	tlz %flin\%flout	; Now closed
	hrrz t,4(bp)
	trnn t,%doblk		; Block mode only
	 jrst rerr+%ensmd
```

And the check is for a block mode / versus non-block mode situation.

**Comment by eswenson1** [comment]
Where is the TV11 i/o handler code (in the source tree)?  It must be that code that is returning %ensmd. I suppose it is possible that there is code in ITS that is doing a mode check before invoking the device driver, but I think that is less likely.

**Comment by larsbrinkhoff** [comment]
I think the PDP-10 side is mostly in in TS3TTY.

**Comment by eswenson1** [comment]
Yeah, I had looked there first and didn't see anything that appeared to return an error code of 12 or 10 -- and no references to the constant either.

**Comment by larsbrinkhoff** [comment]
I put a breakpoint on NSIOT.  I got to this JRST OPNL12.

```
NSIOT:  TLNN B,1000     ;OBJECT IF B.P. OR COUNT IS IMMEDIATE.
         TLNE C,1000
          JRST OPNL33
        CAIGE W,4
         SETZ D,
        MOVEM D,IOTBTS(U)
        MOVE D,IOTTB(H)
        TLNE D,%IOTBK+%IOTSP    ;SIOT ALLOWED ONLY ON UNIT MODE CHANNELS.
         JRST OPNL12            ;"MODE NOT AVAILABLE".
```

I wonder if I messed it up when I changed an .OPEN to a .CALL instead?

**Comment by eswenson1** [comment]
No wonder we couldn't find a reference to %ensmd.  Someone was trying to be super clever.  Damn macros.

**Comment by eswenson1** [comment]
If so, it must be subtle -- probably worth stepping through and using :CALPRT.  I don't see it.  Here is the original code:

```
name1:	.open tyoc,[%tjmor+%tjctn+%tjdis+.uao,,'t52]	;char unit out, disp. mode, no continuation lns
	 .value
	.open tybc,[%tjmor+%tjctn+%tjdis+.bao,,'t52]
	 .value
```
and here is your updated code:
```
name1:	;char unit out, disp. mode, no continuation lns
        syscal OPEN,[ [%tjmor+%tjctn+%tjdis+.uao,,tyoc] ? setz b ]
	 .value
	syscal OPEN,[ [%tjmor+%tjctn+%tjdis+.bao,,tybc] ? setz b ]
	 .value
```

**Comment by larsbrinkhoff** [comment]
When I change back to .OPEN, the SIOT seems to go through.  I guess I have to read UUOS and .CALLS really closely to see if there's a difference between .OPEN mode bits and OPEN control bits.

**Comment by eswenson1** [comment]
I don't see anything obvious. Are you sure the SIOT has NEVER succeeded since you changed from .open to OPEN?  I assumed you'd tested this after you updated it to pass in the TTY rather than modify the code at runtime to supply the TTY.  And if so, this means that it works, at least sometimes.  So perhaps you are just getting lucky when you switch back to .open?

**Comment by larsbrinkhoff** [comment]
Well, SIOT certainly failed every time I tested it today.  And now I ran with .OPEN successfully twice in a row.

I'm questioning my own sanity.  My recollection is that I tested this and the free display came online on the first try.  But now I'm doubting whether I tested the right program.

**Comment by eswenson1** [comment]
Do you know whether you had made the change to NAMDRG before you made the picture that you posted on Twitter very recently?  Because that certainly proved that THAT version of NAMDRG worked with the new TTY number!

**Comment by larsbrinkhoff** [comment]
That was probably name with the self modifying .OPEN.

**Comment by larsbrinkhoff** [comment]
When NSIOT is entered, A is 25001,,2, which is %tjmor+%tjctn+%tjdis+.uao,,tyoc.  However, CTLBTS(U) is 3 and this is xored with the right half.  So the mode bits are changed to 25002.  2 is no longer .uao but .bai.

So why is CTLBTS(U) not 0?  I assume this should be set from a %CLBTW or %CLBIT argument.

**Comment by eswenson1** [comment]
Interesting!  And that certainly explains the mode error from the SIOT.

**Comment by eswenson1** [comment]
Where do you see NSIOT using CTLBTS(U)?  I don't see the control bits being referenced using that location until much later in the file (in NSOPEN).  On the other hand, I do see NSIOT using IOTBTS(U).  I'm looking in ITS 1648.

**Comment by larsbrinkhoff** [comment]
Sorry, I meant NOPEN.  Since changing between .OPEN and OPEN, I looked at AOPEN and NOPEN.  NOPEN+22 uses CTLBTS.

**Comment by eswenson1** [comment]
Right.  NOPEN uses them.  How much pain would it cause to put a breakpoint at the system call dispatcher where, presumably, the CTLBTS are set up, before the dispatch to NOPEN?  Maybe kill off a bunch of other jobs to prevent too much noise first?

**Comment by larsbrinkhoff** [comment]
I just tried it but without killing off anything.  But that was too much noise; at least a dozen system calls.  I will have to try more tomorrow.

**Comment by larsbrinkhoff** [comment]
Is there a better way to examine CTLBTS(U) than this?

```
U[   2662
CTLBTS+2662[   0

larsbrinkhoff:
I'm tracing the OPEN call now.  It entered NCALL and cleared CTLBITS.  So far so good.

Then we get to ASYSC5, and now R is 025000,,000003.  This is an immediate control bits argument.  I don't know where it came from, but W is 40462 so I'll have to check that in the timesharing DDT.

larsbrinkhoff:
The problem is that SYSCAL doesn't terminate the argument with the SETZ bit!

larsbrinkhoff:
OH I SEE MY MISTAKE!  See next commit.  What an embarrassment.

larsbrinkhoff:
I suppose I should build this from scratch and run it a zillion times to make sure it *really* works this time.

larsbrinkhoff:
Seems to work as far as OPEN and SIOT are concerned, but now I get no NAMPIC.

larsbrinkhoff:
Oh, it's probably just this: https://github.com/aap/tv11/issues/19

larsbrinkhoff:
Yes.  I believe this is good to go now.  Please review.

eswenson1:
> Is there a better way to examine CTLBTS(U) than this?
> 
> ```
> U[   2662
> CTLBTS+2662[   0
> ```

I can't find any. Spent the last hour reading ddtord and don't see anything.  That seems really strange to me, since this would be an obvious thing to do.

PR #1427, opened 2019-01-21
Use H3MAKE

larsbrinkhoff:
Use H3MAKE to build the host table from the source file.

This also fixes the bug that the host table should be named SYSBIN; HOSTS3 > rather than SYSBIN; HOSTS3 BIN.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ DOC = info _info_ sysdoc sysnet syshst kshack _teco_ emacs emacs1 c kcc \
       combat pdl minits mits_s chaos hal
 BIN = sys2 emacs _teco_ lisp liblsp alan inquir sail comlap c decsys \
       graphs draw datdrw fonts fonts1 fonts2 games macsym maint imlac \
-      _www_ hqm gt40 llogo bawden
+      _www_ hqm gt40 llogo bawden sysbin
 MINSRC = midas system $(DDT) $(SALV) $(KSFEDR) $(DUMP)
 
 # These are not included on the tape.
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -178,9 +178,13 @@ respond "*" ":midas syshst;_syshst;h3make\r"
 expect ":KILL"
 respond "*" ":link syshst;ts h3make,syshst;h3make bin\r"
 
-# build binary host table
-respond "*" ":syshst;hosts3 /insert syshst; h3text > /outfil sysbin; hosts3 bin\r"
-expect ":KILL"
+# Run H3MAKE to make the SYSBIN; HOSTS3 > database.  H3MAKE looks for
+# an older version, so there is an empty placeholder to make this
+# work.  H3MAKE runs in the background and leaves either SYSHST; H3TYO
+# or H3ERR with the output from the HOSTS3 program.  The placeholder
+# file is removed later.
+respond "*" ":syshst;h3make\r"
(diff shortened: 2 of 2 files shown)


larsbrinkhoff:
Now purifying NAME fails.  This is usually due to a bad HOST3 BIN database.

I don't know what H3MAKE did in the background.  I'd say it's actually better to run HOSTS3 BIN (the program) and see it's messages in the build log.

eswenson1:
I just merged in the changes from H3TEXT > in DB to to my copy on ES.  Then I ran :syshst;h3make bin.  I had no issues.  I saw a new sysbin;hosts3 >.  I ran NAME and had no errors.

larsbrinkhoff:
I think we can make this work.

But the problem is when someone edits the host table and HOSTS3 complains.  If we run H3MAKE we never get to see the error message.  Or maybe it writes them to a file somewhere?  I'll check.

eswenson1:
H3MAKE writes files to SYSNET:  H3TYO NNNNNN are status messages about the building of H3TEXT NNNNNN.  H3ERR NNNNNN are error messages. It also appears that COMSAT will send you email of a failed attempt to build with H3MAKE.

eswenson1:
And what about a new table in SYSBIN? Did it create one?

larsbrinkhoff:
No HOSTS3 in SYSBIN.  It's as if H3MAKE was never run at all.

It should be run here: https://travis-ci.org/PDP-10/its/jobs/484021028#L2305  
And here are the file listings: https://travis-ci.org/PDP-10/its/jobs/484021028#L2611

eswenson1:
I guess the next step is to make a TCL build that does everything up until the point of the H3MAKE command and then stops.  Then we should boot/login and debug this.  I'm happy to do it unless you want to.

larsbrinkhoff:
Thanks, please do!

eswenson1:
Ok.  Running a build that will stop right before the h3make call.  Will let you know what I find.

eswenson1:
The reason H3MAKE is failing is because it attempts to find out the correct FN2 (version of host table) to use by opening DSK:SYSBIN;HOSTS3 >.  This file doesn't exist (yet) and hence the OPEN fails. This is a fatal error.  I created a dummy file DSK;SYSBIN;HOSTS3 2013 (with text contents "foo") and then proceeded the stopped/disowned H3MAKE job.  It completed successfully and created SYSBIN;HOSTS3 2014 and SYSHST;H3TYO 2014.  The contents of SYSHST;H3TYO 2014 was:

```
Reading text file DSK:SYSHST;H3TEXT 2014
Processing tables...
Warning: Duplicate host def for DB-ITS.EXAMPLE.COM
105 nets, 21 sites, 39 names; length 905
Writing HOSTS3 format binary file DSK:SYSBIN;HOSTS3 2014
```

Note: the warning is correct.  We have two entries in H3TEXT for DB-ITS.  I think one of them is supposed to be KA-ITS.  In any case, that is a warning and not a fatal error.

I suggest we either include a real host table (as a binary file in the build) (e.g. SYSBIN;HOSTS3 2013) or create a dummy placeholder file (as I did).  Then, after HOSTS3 2014 is created, we can have the build script delete the dummy file.

larsbrinkhoff:
Thanks.  I added a placeholder.

larsbrinkhoff:
MacOS build failed again.  Why?!?

PR #1502, opened 2019-01-22
Imlac additions

larsbrinkhoff:
- Adding @livingcomputermuseum's Imlac simulator.
- Have it boot with an appropriate SSV or SST.
- Add back the old Imlac terminal type which works with SSV.

WORK IN PROGRESS


larsbrinkhoff:
The Imlac simulator works OK with SSV 22.  Well enough to run PEEK.

larsbrinkhoff:
Not sure this is useful for ITS, but here's the SAIL Imlac directory:  
https://www.saildart.org/[IML,SYS]

Maybe that DataMedia terminal emulator?

larsbrinkhoff:
https://www.saildart.org/FIXIML.ME[UP,DOC] - How an Imlac was loaded from the PDP-10.

larsbrinkhoff:
https://www.saildart.org/IMLAC.BO[UP,DOC]  
https://www.saildart.org/IMLAC.PMF[UP,DOC]  
https://www.saildart.org/IMHAK.BO[UP,DOC]  
https://www.saildart.org/IMKEY.BO[UP,DOC]

larsbrinkhoff:
Add a link from IMLAC; .PRGM. NORMAL to SSV or SST.  Or IMLAC; .PRGM. Txx to a particular terminal.

larsbrinkhoff:
I have sImlac bootstrapped from ITS.

A `notelnet` terminal port was used.  Commands in sImlac were:

```
set bootstrap stty
attach tty telnet hostname port
go
```

On the ITS side, run `IMLOAD`.  Type the tty number without T.  Confirm with Enter.

larsbrinkhoff:
This is still a bit experimental, but the basics are there.

I believe the SSV 22 provided may not be directly from MIT, but rather possibly a patched version from somewhere else.  But other SSV versions don't work well with sImlac; investigation pending.

The "start" script can run sImlac and attach it to ITS terminal T50.  But IMLOAD'ing it from ITS always seems to fail.  Running the same steps manually has a better success rate.  Again, the reasons are unknown.

The one version of SSV that should be compatible with MAZE always fails the bootstrapping.

larsbrinkhoff:
Yes, they self-kill when the user types ^Q.  Meaning quit, presumably.

I try to use the application built in quit command when possible.

larsbrinkhoff:
Your review wasn't an approval, so I'm re-requesting.

larsbrinkhoff:
Thanks!  It's good to close a pull request that has been open over a year.

PR #1507, opened 2019-01-24
Fixes file corruption in common/lins.463.  

### LINKED ISSUE #1500: COMMON; LINS 463 is damaged
I have a good version on ES.  At line 796, things go astray.  DB version has this:

```
Almost anything derogatory you could say about todays's software design
would be accurate.
                                 -- Kenneth E. Ia Masters (of M&J fame)^@
Since aerosols are forbidden, the police are using roll-on Mace!^@
I can't understand it.  I can't even understand the people who can
understand it.
                        -- Queen Juliana of the Netherlands.^@
```

ES version has:

```
Almost anything derogatory you could say about todays's software design
would be accurate.
                                 -- Kenneth E. Iverson^@
        Famous last words:^@
We are getting into semantics again.  If we use words, there is a lot of
relatives on the train for home.^@
A beginning is the time for taking the most delicate care that balances are
correct.
                 -- from "Manual of Maud'Dib" by the Princess Irulan^@
```

DB version ends with:

```
Almost anything derogatory you could say about todays's software design
would be accurate.
                                 -- Kenneth E. Iverson^@
        Famous last words:^@
We are getting into semantics again.  If we use words, there is a lot of
relatives on the train for home.^@
A beginning is the time for taking the most delicate care that balances are
correct.
                 -- from "Manual of Maud'Dib" by the Princess Irulan^@
```

ES version ends with:

```
I die.  My replacement reads my uncommented code and deletes a
fragment he doesn't understand.  Eventually the subroutine is sold to
the government, and the bug causes nuclear missles to be launched by
accident.  The other side retaliates, and all die.  O the
embarrassment.^@
Think of C++ as an object-oriented assembly language.^@
The only thing better than TV with the sound off is Radio with the sound
off.
                        -- Dave Moon
```

DB version is 3519 lines long.  ES version is 3518 lines long.   Seems some disk blocks got shuffled in the DB version.

eswenson1:
Resolves #1500.


eswenson1:
:-)

PR #1512, opened 2019-01-26
Added CHATER. This programs displays active com links.

### LINKED ISSUE #517: CHATER - show com links
SYS1; TS CHATER  
AR3: GREN; COMS 17  
KP;MACROS > *missing*

eswenson1:
Resolves #517.


eswenson1:
Hmm.... looks like I need to find an inserted MACROS > KP; file, that I don't believe I have. May have to withdraw this one.

eswenson1:
Got it to build.

eswenson1:
@larsbrinkhoff Ready for review.

eswenson1:
Well, those were the only two macros used from whatever was in the original. And it can be replaced with the real file if we can ever find it. As reconstructed, the chater binary matches exactly. It probably would have been better to provide accumulator defines in the macro file, but then they would have been duplicated in COMS as well.

PR #812, opened 2019-01-28
DFTP - Transfer files to/from the Datacomputer

larsbrinkhoff:
SYS1; TS DFTP  
INFO; DFTP 2  
.INFO.; DFTP CHART  
.INFO.; DFTP INFO

Not sure where the source code is, but this looks promising:  
MRC; DFTP 200  
DATCOM; DFTP 191  
DATCOM; -READ- -THIS-

The symbol table mentions JOBDAT, so maybe it was linked together.

The DATCOM directory also has many FASL files, so it looks like there was some Lisp program too.


larsbrinkhoff:
I have this built now.  Minor inconveniences are the lack of an ARPANET and a Datacomputer.

larsbrinkhoff:
I don't think anyone is.  But setting up some ARPANET links shouldn't be too far off.

eswenson1:
Can't imagine it would be hard to write a little server that supports the DFTP commands/transfer and stores in a file system on a host. I assume the protocol is pretty trivial?

larsbrinkhoff:
I haven't looked into it.  I think it's documented, maybe in an RFC or maybe elsewhere.

The protocol is probably simple; maybe just sending text lines back and forth.  But there's some kind of query language which may be trickier.  DFTP maybe just needs a trivial subset, though.

larsbrinkhoff:
datacomputer

brfid:
> I don't think anyone is. But setting up some ARPANET links shouldn't be too far off.

IMP code from about that period is available: https://walden-family.com/impcode/

larsbrinkhoff:
Thank you @brfid.  Yes, Dave Walden, Bob Armstrong, et al has run a small network of simulated IMPs using SIMH:

https://github.com/simh/simh/tree/master/H316/tests

But having just the IMPs is quite boring, of course.  I occasionally take tiny steps towards adding the necessary IMP-host interface to make it possible to make a small Arpanet with interacting hosts.  Another obstacle is that most host software has disappeared.  We have code for several PDP-10 systems, and the Illinois' PDP-11 Unix, but not much more.

brfid:
> Thank you @brfid. Yes, Dave Walden, Bob Armstrong, et al has run a small network of simulated IMPs using SIMH:

I should have know that you were all over that! ...I've also been thinking about what's left in terms of IMP/host as well as the NCP implementations. I'll comment more once I've got a better idea of what you've done.

larsbrinkhoff:
Sorry, it's not much so far.  I updated @charlesUnixPro's patch for inclusion in SIMH:  
https://github.com/simh/simh/commit/e8bceec8e2e681bec741796fa99d3dcd22e64ad8

I started a standalone IMP emulator/reimplementation in C, but with the original IMP code running in SIMH that may not be necessary.  Or maybe it will be, if the 1973 code doesn't support 1822 long leaders.

PR #1522, opened 2019-01-29
Updated emacs from version 162 to version 163.

### LINKED ISSUE #1520: Incorporate patches in EMACS;PAT162 > into appropriate emacs sources
We should incorporate the emacs patches in EMACS; PAT162 > into the appropriate source files and generate a new version of emacs (with no patches).

eswenson1:
This incorporates the patches from 162 into files 434 and isearch
75. Ignoring the patch to & Read Command Name since the version in the
patch file is older than the one in crl 208. 

Resolves #1520.


eswenson1:
@larsbrinkhoff Note: I think someone twiddled a couple characters when TS 126 was created (in PI).  I think the version is supposed to be the same as the emacs version number (which was 162).  In any case, I made the FN2 of the TS file in EMACS; be 163, which matches the version of [pure] and the version reported by Emacs when it starts up.

eswenson1:
However, looking at backup maps, I see that there is only a TS 152 (after TS 126). Still, I think it is wrong to name the TS file 126 (or 127) when the version of emacs is based on the FN2 of [pure] and we know that version is 163 (now -- previously it was 162).

larsbrinkhoff:
I dunno.  Even back at TS 13/14, the corresponding pure files are [PURE] 24 and 25.

larsbrinkhoff:
On the other hand, we have a chance to make this better.

eswenson1:
I like making things better!

Issue #1324, opened 2018-10-08
Missing program aliases.


larsbrinkhoff:
- ETTIME, alias for VTTIME.
- GCBULK, alias for GCMAIL.
- Aliases for SENDER: CHSEND, LMSEND, FSEND, SSEND, SS, PIX, SHIT, OHSHIT, RP, R, RA.
- Aliases for PR: TVDOC, TVDOCT, MUD, MUDDLE, INTRUP, PRIM, LSC, NLSC.
- Aliases for TELNET: TCPTN, TTN, NCPTN, PTN.

larsbrinkhoff:
Maybe some more here?

https://github.com/PDP-10/its/blob/4e2ea8e4d851a0ea1f910c3001e5c2ddb9ff528c/src/sysen1/jobs.87#L66-L118

larsbrinkhoff:
TS BOOT11 is an alias for TS 11STNK.  It's a timesharing version of .; @ BOOT11, i.e. it takes 11DDT and IOELEV and stuffs a PDP11 over DL10.

larsbrinkhoff:
STYLOC is an alias for TTLOC.

larsbrinkhoff:
Here's a long list of links:

http://sv.svensson.org/KLH%3BXLINKS

eswenson1:
Any idea what CHANNA; TS SAM -> CHANNA; MAHASA MATMAN is/was?  This link exists in PI/ES but the link target is non-existent. I see no trace of "TS SAM" and MAHASA or MATMAN anywhere.

eswenson1:
I read the one I picked up in Tallinn, but evidently all the cool ITS names came from the other. Can’t even remember the names of the books — just the name Amber. In any case, I didn’t the book compelling enough to read the other. It was ok — just not my genre.

larsbrinkhoff:
I think most names come from Lord of Light.  Channa, Taraka, Rakash(a), and Mahasamatman if the link counts.  The Amber series is just Chaos(net) and Amber.

eswenson1:
Yeah. I just read the first book of the Amber series. How was Lord of Light?

larsbrinkhoff:
[...]

PR #1532, opened 2019-02-11
Link from TS BOOT11 to 11STNK.

larsbrinkhoff:
11STNK doubles as a timesharing BOOT11 command.  I haven't tested it, obviously, but I would expect it to send something to the DL10 PDP-11.

I can't see any link in the records, so I'm guessing the . directory is a good location.

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -1418,6 +1418,7 @@ expect ":KILL"
 # PDP-11 linker.
 respond "*" ":midas sys1;ts 11stnk_kldcp;11stnk\r"
 expect ":KILL"
+respond "*" ":link .; ts boot11, sys1; ts 11stnk\r"
 
 # CARPET, remote PDP-11 debugger through Rubin 10-11 interface.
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Because 11STNK checks its JNAME and if it's BOOT11 it'll take a special action.

```
ITS,[   .SUSET [.RXJNAME,,A]    ;See if this is BOOT11
        CAME A,[SIXBIT/BOOT11/]
         JRST CMD               ;No, take 11STNK commands
        MOVEI Q,DDTFN           ;Yes, load .;11DDT 16K
        PUSHJ P,DDTLD
        MOVEI Q,PRGFN           ;and .;IOELEV BIN
        PUSHJ P,PRGLD
        MOVEI TT,[ASCIZ/Stuff/]
        PUSHJ P,ASCOUT
        PUSHJ P,STUFF           ;ask user then stuff it
        .LOGOUT 1,              ;go away if user types N
];ITS

larsbrinkhoff:
We have more of these: #1324

Issue #1534, opened 2019-02-14
Color scope


larsbrinkhoff:
There was a color display device.  Device code 420 (same as Spacewar consoles), symbolic name CLR or CSCOP.

larsbrinkhoff:
Michael Beeler wrote:
> The color scope was a prototype made by DEC.  They were considering producing a color scope for sale, but never did.  There were two prototype scopes built; one they gave to the AI lab, and I don’t know what happened to the other one.
> 
> It used a regular color TV CRT tube, but all the driving electronics were special purpose stuff built by DEC.  It had only 9 bits of deflection in X and Y, so it was less precise than the 340.  Its spot size was larger than the 340, partly because it had to cover enough of the red, green and blue phosphor spots to have a consistent color.  It was slow!  I don’t remember just how long it took the deflection amplifiers to settle, but it was well longer than the 340.  The color was 6 bits of red, 6 of green, and 6 of blue.  It was only point-plotting, no line, character, or other capability.
> 
> There was no connection between the 340 and the color scope.  “Solar Corona” was shot on the color scope for the sun images, and on the 340 for the titles.  The developed, original film was spliced by hand to make the movie.
> 
> The color scope had a major problem with “convergence”, that is, making the beams from the red, green and blue electron guns all hit the screen at the same point.  This was a problem in color TVs too, but was especially so with the color scope.  For each gun, there was a correction factor that was a weighted sum of a constant factor, X, X^2, Y, and Y^2.  That is, there were constant, linear, and quadratic dependancies on both X and Y, for each of the three electron beams.  The weighting was done with potentiometers (variable resistors) on printed circuit borads, very hard to access.  I made new PC boards with shielded cable to a panel that sat on the tabletop next to the display (where the electronics were under the table).  The panel had a whole bunch of potentiometers, each with a knob and graduated scale, so it was easier to adjust the convergence.  Nevertheless, it was a frustrating thing to work on, and it was always an approximation — no setting ever made all spots just right.
> 
> There also was in interesting feature; if you sent the command to display a point (turn on the beam for a few microseconds) before the deflection amplifiers had settled, the point would be smeared into a short streak between (0,0) (the bottom left corner of the screen) and its correct position.  This is because the
[...]

PR #1543, opened 2019-02-18
TITLER - display text on color scope.

larsbrinkhoff:
See #1534 for information about the color scope.

The program is fine as is, but color scope simulation is a work in progress.


larsbrinkhoff:
The program runs in timesharing if IOTLSR mode is entered at the start.

larsbrinkhoff:
Simulation screencast:  
 https://youtu.be/ymJ5D9exvUw

PR #931, opened 2019-02-20
GAS - Michael Speciner's gas simulation

larsbrinkhoff:
MS; DISP GAS3


larsbrinkhoff:
I first tried, some time ago, to make the source work.  But I didn't get very far.

Obviously the binary works, so I started reconstructing the corresponding source.  It's showing promise.  The source and binary are largely the same, but have some crucial differences.  However, I hit a snag midway with a large section in the binary that doesn't have a corresponding source section that I see so far.

Rather than have that work dragging on indefinitely, it's nice to have the working binary available.

larsbrinkhoff:
Should include a link to Speciner's memo here: ["How the GAS Program Works"](https://dspace.mit.edu/bitstream/handle/1721.1/5829/AIM-272.pdf).

PR #1545, opened 2019-02-22
Recent changes to DDT have broken various control characters

larsbrinkhoff:
The most recent changes to DDT (to support bucky bits) have broken DDT's ability, when run over SUPDUP, to handle various control characters, including ^F and ^X.  Reverting DDT to the previous version makes these work again.


larsbrinkhoff:
I changed DDT to always be prepared to handle the bucky bits.  My expectation was that the reading from a %TIFUL channel wouldn't return 12-bit characters unless %TOCFI is set.  But it seems that ITS at a low level intercepts the <kbd>^\</kbd> character and uses it to send commands in the input stream.  Thus any source can introduce bucky bits regardless of %TICFI.

larsbrinkhoff:
Also add this update when it's done:  
https://github.com/PDP-10/supdup/pull/19

larsbrinkhoff:
@eswenson1, either of these two commits should fix your problem.   Or maybe it's my problem.  Anyway.

eswenson1:
By the way, I did update EX with the updated version of DDT (are you sure we shouldn't be incrementing version numbers from 1547 to 1548 here), and now control characters work fine from the original linux supdup. I then updated the linux supdup client to the latest in this branch, and tested it. It also worked fine.  Thanks.  (Still, I cannot approve this PR, because GitHub keeps thinking I'm trying to comment rather than approve -- this from both my mac and iPhone).

larsbrinkhoff:
Thanks for checking.  Strange that you can't approve, but it's not a blocker.

Having trouble getting this build with Circle CI using the SIMH KS10.

eswenson1:
Feel free to merge whenever the build succeeds.

larsbrinkhoff:
It seems there's an actual problem, not just a temporary glitch.  The SIMH build always hangs here:

```
MAXIMUM TOOLAGE>gen-tags
; Generating NTAGS file.
:$ I'm using a VALRET $:TAGS MACSYMA;MACSYMA NTAGS
:KILL  MAXTUL$J
*:CONTINUE
; Now, do :MOVE MACSYMA;MACSYMA NTAGS, MACSYMA TAGS
; after verifying that things are ok.
MAXIMUM TOOLAGE>quit
QUIT
* 
(quit)
:KILL 
*:move macsym;macsym ntags,macsym tags
*aljabr$^S
*:lisp

**Comment by eswenson1** [comment]
Because ALJABR has a * LISP file, the customary prompt that LISP emits on startup is not present. It does appear that LISP does print a newline before entering the REPL. In any case, we normally use a "respond" TCL command to provide the next input line, but in this particular case we don't.  We simply provide the "(load \"lisp;mlsub\")" input using "type".  Not sure why this isn't being sent -- doing so would cause LISP to display something that the next "respond_load" would handle.  

Does the build hang and then eventually TCL times out?  Is it waiting on the response from the loading of MLSUB, but that never appears?  That would suggest that LISP isn't seeing the load of MLSUB request.  And that might suggest that TCL never sent it.

Not sure what to suggest here.  This code hasn't changed in eons and clearly doesn't have issues on other emulators.  Perhaps changing the "type" to a "respond" if you can match on the newline emitted by LISP on startup in this case?

**Comment by larsbrinkhoff** [comment]
Build hangs at the same place on my laptop too, so I'll have to look into this.

**Comment by eswenson1** [comment]
Things are definitely wrong with the new DDT.  Various characters are not echoed correctly now.  For example, typing <esc><esc>v now echoes as the <esc> as a diamond.  And typing ^r as an upside down u.  While I understand how this might be "correct" for some terminals, this is not how ITS displayed these characters in supdup terminals.  This may be due to the new supdup.  I haven't reverted to the old supdup with the new ddt yet to find out.  But if it is not related to the new supdup, it might explain why the build is failing too (maybe).

**Comment by larsbrinkhoff** [comment]
Thanks, I'll look into it.  It's not supdup, it's DDT.

**Comment by larsbrinkhoff** [comment]
Apparently, when a control character is echoed to SUPDUP, it'll display as the SAIL character.  But when it's echoed to a TV, it displays with the ^ prefix.  That was unexpected.

**Comment by larsbrinkhoff** [comment]
This is very strange.  According to expect, `aljabr$^S` and `:lisp` are echoed back normally.  But when `(load "lisp;mlsub")` is entered, nothing is echoed back.

**Comment by larsbrinkhoff** [comment]
If I add `(PRINC "HELLO")` to ALJABR; * LISP, the build can proceed.  Just a random observation, not a proposed solution.

**Comment by larsbrinkhoff** [comment]
I came up with a different solution.  If %TOFCI isn't set, read from the ordinary TTY input channel rather than the %TIFUL one.  Now we get ASCII control characters, there is no messing with an altmode prefix, and ITS can do the echoing.

**Comment by eswenson1** [comment]
Just tried this version of DDT. Seems to echo control characters fine.  I haven't found anything awry yet.

**Comment by larsbrinkhoff** [comment]
Thanks.  I'll have to test it with the Knight TV too.

**Comment by larsbrinkhoff** [comment]
Test cases:
- [ ] Console.
- [ ] Serial line.
- [ ] Previous version of supdup.
- [ ] Latest version of supdup.
- [ ] Knight TV.

For all above:
- Does <kbd>^F</kbd> work?
- Does <kbd>^\\</kbd> work?
- Does <kbd>Meta</kbd> work?  E.g. logging in with <kbd>Meta-U</kbd>.

**Comment by larsbrinkhoff** [comment]
I'm thinking about the JFCL no-skip return, and a better name for BLAH.

**Comment by larsbrinkhoff** [comment]
Yes, that's what I mean.

**Comment by larsbrinkhoff** [comment]
Ok, how about this now?

**Comment by eswenson1** [comment]
Can you provide a brief summary of the net change from the original DDT (and supdup)? I think the diffs I see are against a previous updated version and i’m having trouble seeing the what the actual change is now against the version of DDT before any (related) changes were made.

**Comment by larsbrinkhoff** [comment]
Fair enough.  It's not easy to diff commits for renamed files.  I did this:

```
git diff -M 084d79d3c5932be32969afd1c78f64ac44dff717 HEAD -- src/sysen1/ddt.1546 src/sysen1/ddt.1547
```

The change is to use the %TIFUL (channel full character set) input channel only when %TOFCI (TTY capable of full character set) is set.  Characters are then stripped of top and super, and control characters are folded.  Meta is converted to an altmode prefix.

To make this look good, DDT does its own echoing for %TOFCI terminals.  This is just to echo ◊ for the altmode.

eswenson1:
Once again, GitHub is not letting me approve this PR. It forces me to leave comments only. Selecting Approve tells me that a comment is needed, and adding a comment (with Approve) just adds a comment.  So consider this approved by me.

larsbrinkhoff:
I tested it before, but I'll do a final round of testing before merging.

Issue #564, opened 2017-02-27
MIT- prefix


larsbrinkhoff:
Some programs accept an MIT- prefix before hostnames, and strip it out for output.  No much harm done.

However, some programs will add an MIT- prefix when printing an ITS hostname.  That's no longer accurate.

larsbrinkhoff:
Programs adding MIT- for output:
- CFTP

eswenson1:
PWORD fixed in this PR:  #1551

PR #1551, opened 2019-02-23
Update PWORD to not add MIT- prefix to machine names in emails

eswenson1:
Also, add ES to list of known machines.


eswenson1:
This is one of several programs that add the MIT- prefix.  See #564.

PR #1567, opened 2019-03-12
C library for TV raster graphics.

larsbrinkhoff:


eswenson1:
I don't see this library being built. Is it supposed to be?

larsbrinkhoff:
No, it's not built.  I was thinking it doesn't need to be until there's an application that uses it.  What do you think?

eswenson1:
Hmm....I think if it is buildable, and could be used for linking with if there were an app that needed it, we should build it.

larsbrinkhoff:
Fair enough.  I updated the commit.

PR #995, opened 2019-03-20
EMACS Macsyma mode

larsbrinkhoff:
EMACS1; MAXIMA 26  
EMACS1; MAXIMA INFO  
EMACS; MAXIMA :EJ

Now that we have Macsyma!


larsbrinkhoff:
Some obvervations:

- I don't see that MAXIMA INFO is linked from INFO or .INFO.
- I don't see any files with `-*- maxima -*-`.
- There is a "Macsyma Mode" in USRCOM.

eswenson1:
Isn’t maxima the Common Lisp rewrite of Macsyma — the latter being the maclisp one? ITS macsyma sources would have -*- Macsyma -*- and would use Macsyma mode, not Maxima mode.

eswenson1:
Given the above, why do we need this in ITS?

larsbrinkhoff:
I think it's just **named** "maxsima".  The INFO file says it's from 1979 where Maxsima or even Common Lisp didn't exist.  The TECO file says " EMACS macros for editting Macsyma code".

I don't know if this adds much value, but I don't think we should discard it either.

PR #1584, opened 2019-04-10
PDP-6 WAR 44

larsbrinkhoff:
PDP-6 Spacewar! version 44.  From Samson's DECtape "PSamsonIV".

Work in progress branch:  
https://github.com/PDP-10/its/tree/lars/war44

Scanned listing!  
https://www.computerhistory.org/collections/catalog/102680667


larsbrinkhoff:
Note that the consoles are totally different from the ones used by ~1975 SPCWAR.

Device code 724 can be DATAI'd to get 36 bits.  Each half describes one console.  There are bits for:

- Rotate clockwise
- Rotate counter clockwise
- Faster (thrust)
- Slower (thrust)
- "Button", Torpedoes
- "Xmit data", same as torpedoes
- "Pedal", hyperspace
- "General cancel", self destruct
- Time 0
- Time K
- Beacon beam

I don't know what the last tree do.  The "beam" is indeed a beam, but I haven't discerned any effect.

larsbrinkhoff:
Beam in action.
beam

larsbrinkhoff:
I asked Michael Beeler about the beacon beam.

> That was added (I don’t know who put it in, or when) as an aiming aid.  One of the challenges in Spacewar is being able to point your ship accurately so that your torpedoes will hit the other ship.  With both ships moving, perhaps accelerating due to thrust, and being pulled by the star’s gravity, it
was a good aspect of the game that required some learning and skill.  The idea was that the beam was sort of like a laser pointer.  It would show you where torpedoes would go if you fired them now.  I myself think it was sort of cheating.  In any case, as far as I know it never became very popular.

> That was also the fate of some other variations, such as four players.  In that version, two players
would almost always get shot down very quickly, and the game then became just like normal, with only two players.  Another variation was to have angular momentum; the turn switches were like turning thrusters, and put a turning force on the ship.  This version was very hard to play.  You were too much absorbed in building up some turning rate or slowing down the turning rate, and you didn’t have enough attention to think about speed and where the other ship was and whether you were going to fall into the star, etc.  I think the only variation that survived well was the option for the star to have gravity or not.

PR #1582, opened 2019-04-11
Standalone Lisp

larsbrinkhoff:
@ LISP

Check if it works, and if so, add a copy to the . directory.

EDIT The binary is updated thusly:
- Load @ LISP.
- Load the in-core symbol table.
- Remove in-core symbol table and DDT.
- Insert patch to replace FSC.
- Dump back to @ LISP.


larsbrinkhoff:
My patch is to replace 

```
        FSC     A,@ECHOCC
        FSC     A,@TICC
        TLNE    A,200000
         JRST   DING
```

with 

```
        PUSH P,B
        MOVE B,ECHOCC
        ADD B,TICC
        JRST PATCH
LOC PATCH
        TRNE B,200
         JRST DING
        JRST PITELE2-1
```

I see now I have to pop B somewhere.

larsbrinkhoff:
Obscure DDT command is very useful here.

@ LISP includes an NTS DDT at 34000, and there's an in-core symbol table before that.  By typing

    33777/ / ^Y

That symbol table is added to the TS DDT symbol table.

larsbrinkhoff:
Also add `◊q◊,777777◊◊z` to zero symbol table and DDT.

larsbrinkhoff:
This memo is probably close in time to @ LISP:  

https://github.com/larsbrinkhoff/its-archives/blob/master/ailab/AIM-116a.pdf

larsbrinkhoff:
Suggested test for the display:

```
(DISINI (QUOTE FOO))
(SETQ DISLIST (LIST GET (QUOTE FOO) QUOTE SUBR))
(DISAD (QUOTE FOO) NIL 'HELLOWORLD)

larsbrinkhoff:
This is only built for KA10.  The reasoning is that it's not much use to have a @ LISP on a KS10 system, since it will not run.

larsbrinkhoff:
Hmm, not sure it works.  Better test some more.

PR #1523, opened 2019-04-13
ARDS picture FOOBAR doesn't render well

larsbrinkhoff:
-PICS-; FOOBAR PIC doesn't render OK with either SUPARD on a TV, or with @bictorv's SVG translation.

foobar1

foobar2

### DIFF
--- a/src/dcp/sgincl.15
+++ b/src/dcp/sgincl.15
@@ -26,6 +26,7 @@
 (defvar sg-current-x 0) (defvar sg-current-y 0) (defvar sg-xor-ior 'ior)
 
 (defconst %TDGRF #o231) (defconst %TDNOP #o210) (defconst %TDRST #o230)
+(defconst %TDCRL #o207)
 
 (defconst tv:alu-ior 0) (defconst tv:alu-andcam #o40) (defconst tv:alu-xor -1)
 (defconst %goMVR #o001) (defconst %goMVA #o021)
@@ -159,6 +160,8 @@
       ((ior)	 (sg-out %goIOR) (setq sg-xor-ior 'ior))
       )))
 
+(defun sg-return ()
+  (+tyo %TDCRL sgos))
 
 (defun sg-set-point (x y)
   (sg-move-absolute x y))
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
@bictorv, I updated your ARDS converter here.

bictorv:
And the other pictures still look OK, I hope? (Any clue about the reason for all this?)

larsbrinkhoff:
To be honest, I only tested FOOBAR PIC.

The reason is that the very first command in that file is an invisible extended vector.  I.e. is starts with a relative position movement.  Then it draws some more vectors, and next it does a "set point", moving to an absolute position.

Since all other files look fine, I mostly assumed they start off with an absolute position.  Well, I suppose this is a good opportunity to test the "convert directory" function.

Another assumption I made is that FOOBAR PIC isn't malformed, i.e. that the relative movement makes sense.  Since the Imlac program does display FOOBAR PIC correctly, I take that as an indication the file is ok.  But it could be the case that the Imlac program isn't displaying the same was a real ARDS would, and that FOOBAR PIC was created on an Imlac rather than an ARDS.

larsbrinkhoff:
I ran `ards-to-svg-dir` on -PICS- and I see no obvious errors in the resulting SVG files.  HTTP to UP currently gives me Error 503.  I can do a side-by-side comparison when it's back up.

bictorv:
And now it's back UP.

larsbrinkhoff:
All pictures are indeed identical, except for these two exceptions:

- DOOR is positioned slightly more to the left with the new version.
- FOOBAR looks better.

I'd say that's a win, but not unequivocally so.

bictorv:
Let's call it a win. :-)

larsbrinkhoff:
Screenshot from a real ARDS.   http://www.bitsavers.org/pdf/mitre/ESD_Technical_Reports/ESD-TR-69-358_Microprogramming_Facility_Nov69.pdf

The display is 1081 x 1415 points across 6.5 x 8.5 inches.  Text is displayed as 80x50 characters.

ards1

larsbrinkhoff:
Maybe I should have updated the ARDS interpreter in VERSA too (CC @atsampson).

PR #1597, opened 2019-04-30
Add some old .INFO. files.

larsbrinkhoff:
ITS INFO dated 1974.  Note the 1973 info on the 10-11 interface.

? INFO and ^DDT INFO from 1972.


eswenson1:
I hadn’t approved this because I had a pending question/comment!

larsbrinkhoff:
Sorry, I was impatient after two days.

Feel free to post your questions and comments anyway.

larsbrinkhoff:
I have some questions of my own: What is `? INFO` about?!?  What system name?

eswenson1:
I guess I'm asking why we are including this?  What does it do?  It certainly doesn't have any valuable content.

larsbrinkhoff:
I can certainly agree ? INFO is arguable.  The ^DDT and ITS files I think are not useless.  Is the use of console lights to display running jobs documented elsewhere?

My reasoning is not to judge too much.  The files were there on an ITS machine at some point, and I don't know what future generations will think is valuable or not.  That said, I'm not adding every single version of everything; I try to stick to the latest versions.

eswenson1:
Ok

Issue #1598, opened 2019-05-02
Imlac PDS-4


larsbrinkhoff:
There was an Imlac PDS-4 attached.  Backup records show files first located on ML, later MC.

"PDS-4 at Ames" from http://www.digibarn.com/history/04-VCF7-MazeWar/stories/colley.html
![pds4-2](https://user-images.githubusercontent.com/775050/57062566-d7b51400-6cc0-11e9-9399-2fa4a7cf259d.jpg)

larsbrinkhoff:
INFO; TERMS 160 says

```
        Imlac PDS-1, PDS-4

Status: Supported by ":TCTYP IMLAC" if you're running SSV or SST.
Users:  MIT-DM, Plasma Physics (PLASMA-GROUP@MC)
Price:  PDS-1G is minimum $9K all the way up to PDS-4 with options >$20K
Screen: PDS-1 gets 44 lines by 88 chars with standard font.
        PDS-4 gets 50 lines by 89, but can hack smaller fonts yet.
Chars:  Stroke drawn, you need the long vector option!
Keybrd: no great feel, but stands up to an amazing amount of abuse.
        It takes a lot of multiplexing to generate all 128 Ascii codes.
Speeds: I have heard of Imlacs going 80Kbaud and better.
Has:    Everything.
Lacks:  If it doesn't have it, you can always write your own program.
Misfeatures:    Expensive.

It might be handy to attach your own keyboard.  Features are based on
use of the SSV or SST program of the MIT-DM group.  Since the display
processor is a full minicomputer, you can of course run any sort of
program you wish in the terminal.  In particular there is an excellent
set of games available.

CBF  8 March 1978

larsbrinkhoff:
From http://www.chilton-computing.org.uk/acd/icf/terminals/p008.htm

![pds4-1](https://user-images.githubusercontent.com/775050/57062437-637a7080-6cc0-11e9-9e2a-ad547022f080.jpg)

larsbrinkhoff:
"Imlac at Project MAC used for music - Vercoe" from http://www.digibarn.com/collections/photodocumenation/imlac-project-mac/

![pds4-3](https://user-images.githubusercontent.com/775050/57063228-d2f15f80-6cc2-11e9-9ec5-2aaa2e1e6c84.jpg)

larsbrinkhoff:
From http://www.digibarn.com/collections/reports/auerbach-imlac-report/

[...]

PR #1599, opened 2019-05-02
SSV4

larsbrinkhoff:
This is the SSV ("scroll saver") software for Imac PDS-4 (#1598).

Supplied is ASSV4, a program to assemble SSV4.


larsbrinkhoff:
ASSV4 runs MMIDAS, hence the added symlink.

larsbrinkhoff:
I discovered ^Q exit IMTRAN, which is more convenient than typing ^Z and then :KILL.

larsbrinkhoff:
Nothing I push passes the Circle CI build for some unknown reason.  They all break here:

```
*:cwd decsys
*:dec sys:macro
*macro=macro
?MCRPDL PDP OVERFLOW, TRY /P 
 ON PAGE 41 AT  + 1
PDLOV; 412730>>PUSH 17,10   

The last command timed out.

larsbrinkhoff:
The good news is that the build breaks on my laptop too, which should make it easier to debug.

The bad news is that even if I roll back to commits from January, they still break!

#1602 doesn't break?!?

Issue #841, opened 2018-04-26
PUB


larsbrinkhoff:
.INFO.; PUB DOC  
.INFO.; PUB RANDOM  
.INFO.; PUB TIP  
INFO; PUB TIPS  
INFO; PUBXGP DOC  
PUB; PUB2 LOADER  
PUB; PUB DOC  
PUB; PUB FILES  
PUB; PUB FTP  
PUB; PUB LOADER  
PUB; PUB LOG  
PUB; PUBMAC DFS  
PUB; PUBMAC DOC  
PUB; PUB PARC  
PUB; PUB PUB  
PUB; PUB SAILV  
PUB; PUBSTD AIL  
PUB; PUBSTD DFS  
PUB; PUBSTD MAX  
PUB; PUBSTD TES  
PUB; PUBSUM 1  
PUB; PUB TIPS  
PUB; PUB UPD  
PUB; PUB UPDATE  
PUB; PUBXGP 49  
PUB; PUBXGP DOC  
PUB; PUBXGP XGP

Typesetting software.

![](http://www.nomodes.com/images/cover.gif)

larsbrinkhoff:
Hello @nomodes,

PUB isn't forgotten!  In this repository we have, or will have, documents prepared with PUB.

Thank you for providing this manual:  
http://www.nomodes.com/pub_manual.html

larsbrinkhoff:
Looks like these are the source files on Saildart:  
https://www.saildart.org/[PUB,SYS]/

eswenson1:
I tried creating a TS PUB (and TS PUB2) from the PUB; PUB LOADER and PUB; PUB2 LOADER files (using STINK) as per the instructions in PUB; MAKE PUB, and produced the two executables.

However, attempting to run PUB on PUB; PUB PUB (documentation on PUB), gives me this:

> *:ejs;pub
> *pub
> INVALID INDEX NO. 1 FOR ARRAY MONTH
> CALLED FROM 500134  LAST SAIL CALL AT 511104
> ?

I seem to recall seeing email or an issue (authored by Lars?) that referenced this error, which presumably comes from one of the SAIL libraries or perhaps DECSYS; DECBOT BIN.  Is this some Y2K issue.  

Any ideas on how to pursue this?

larsbrinkhoff:
That rings a very faint bell.  I think it was something with Fortran though, so probably Adventure.

larsbrinkhoff:
I think maybe @atsampson was involved?

[...]

Issue #353, opened 2016-12-20
MAIL dates don't handle 21st century very well


eswenson1:
Mail received on ITS displays Date headers similar to this:

Date: Tue, 20 Dec ;6 15:18:33 EST

Note the ";6", which should be "16" or "2016".

larsbrinkhoff:
@Pul-Ess and @bictorv both have some changes to adress Y2K problems.  I tried them lightly, but so far didn't make any progress.

larsbrinkhoff:
I saw this too.  (Except it's been a while, so it was ;9.)

bictorv:
I fixed that in KSC;OUT around 2005, see http://up.update.uu.se/patch; for a description and patch.

larsbrinkhoff:
Thanks.  I chose to conservatively stay with RFC 822 and make the two-digit years work, and I see you updated them to RFC 2822.  That's something to consider.

bictorv:
Hmm, why not follow the current standard when it's so easy to do?

larsbrinkhoff:
In this case I saw the comment `; Year as 2-digit number.` so I stayed with that in order not to risk break any code.  I didn't check anything further, so I didn't notice it's actually for output of RFC 822 date & time.

larsbrinkhoff:
@eswenson1, have ES received any mail lately?  Was the Y2K bug fixed now?

eswenson1:
I just sent some mail from ES to ES and got this:

```You have mail from COMSAT:
Date: Mon,  5 Oct 20 09:36:32 EDT
From: COMSAT@ES-ITS.SWENSON.ORG
Subject: Msg of Monday, 5 October 2020 09:36-EDT
To: EJS@ES-ITS.SWENSON.ORG
Message-ID: <1214.201005@ES-ITS.SWENSON.ORG>

Queued: (FILE [EJS;EJS MAIL]) at ES-ITS.SWENSON.ORG```

Not really sure why it was queued, but all the dates look ok.  Does that answer your question?

eswenson1:
Seems my directory was full -- reason for the above. Just deleted some files and then sent another message.  Received it fine as:

```Date: Mon,  5 Oct 20 09:41:57 EDT
From: Eric J. Swenson <EJS at ES-ITS.SWENSON.ORG>
To:   EJS at ES-ITS.SWENSON.ORG

test```

larsbrinkhoff:
> I just sent some mail from ES to ES and got this:
>
[...]

PR #1602, opened 2019-05-03
Y2K fix for DECUUO's TIME emulation.

larsbrinkhoff:
Use RQDATE instead of .RDATE to get the full year.  The previous version was presumably written in 1972, breaking 28 years later in 2000.  The updated version will stop working in 2052 due to TOPS-10 limiting the returned value to 15 bits.

This has not been tested at all.  Maybe PUB will work better now, see #841.

I'm considering fixing #353 here too.


larsbrinkhoff:
Fixing this bug was very good timing.  Apparently, it just about now started causing problems assembling MACRO.

PR #1600, opened 2019-05-03
3406 - PDP-6 part of 340D

larsbrinkhoff:
3406 is a program that runs on the PDP-6 and displays characters on the 340 display.

See #778 for information about 340D.


larsbrinkhoff:
Sorry, now quite ready for review.

- Erase to end of line must not call INSERT for every space.
- Erasing an entire line should probably use a loop instead of .REPT.

PR #1608, opened 2019-05-15
Fix typo in script to set up a TAP interface

dugoh:
The logfile got truncated by the last command


larsbrinkhoff:
Thanks.  I had this fixed, but it seems I forgot to commit some last minute edits.

PR #1613, opened 2019-05-29
Build on slow computer

larsbrinkhoff:
Tune the build script timeouts to it can run on a slow computer, such as a Raspberry Pi.


larsbrinkhoff:
I'm tweaking the tweaks.

larsbrinkhoff:
Given input from @athornton in #1649, I'm planning to rewrite this to just increase the default timeout to... 200... 300 seconds?

But still, I'd like the CI build to use less.  So use an environment variable for that.

larsbrinkhoff:
I'm retweaking the tweaked tweaks.

larsbrinkhoff:
KLH10 no longer works with this.  It's not obvious why.

larsbrinkhoff:
This seems outdated now and no one has complained in a while.  There are ready built disk images available here: http://hactrn.org/images/

Issue #1614, opened 2019-05-31
Rmode


larsbrinkhoff:
DM's Emacs-like TECO-based editor.

https://youtu.be/FXdmo2j_CiQ?t=3026

larsbrinkhoff:
.TECO.; RMODE ARCHIV  
.TECO.; RMODE BITS  
.TECO.; RMODE CHART  
.TECO.; RMODE DOC  
.TECO.; RMODE RECENT

larsbrinkhoff:
Maybe:  
HAROLD; RMODE 3  
HAROLD; RMODE ABBREV  
HAROLD; RMODE CMNDS  
MRBSYS; RMODE 32  
WJL; RMODE NOTES  
WJL; RMODE TMACS  
WJL; ABOUT RMODE

larsbrinkhoff:
Probably ported to TOPS-20:  
<rmode>mm.teco.2
<rmode>moby.bkd.1
<rmode>moby.rvb.1
<rmode>modedf.teco.185
<rmode>modeld.teco.10
<rmode>modepr.teco.1
<rmode>modeta.teco.117
<rmode>modetw.teco.49
<rmode>pprint.teco.1
<rmode>rmode.ctl.1
<rmode>rmode.recent.7

larsbrinkhoff:
The files MODEDF, MODETA, MOD928, and TECPUR 928 are enough to run Rmode.  It displays a mode line similar to EMACS:

```
    <=<S-EXP ^I ^G SAFE>=> [MAIN]
```

^F moves forward one character.  ^P and ^N moves back and forward by word.  ^L moves to the next line.  ^B to select buffer.  ^A selects some kind of mode.  Altmode doesn't seem to do much.  ^E to load a macro library.

Ah, ^A can set a mode which is some kind of unit by which to move the cursor.  W is Word.  L is List.  S is S-exp.  N is Normal.  R is Raw.  T is Text.  There are also other modes like Safety and Auto Indent.

larsbrinkhoff:
A file called RMODE.INIT found with the German translation of Zork.  Maybe evidence someone at Infocom used Rmode on their DEC-20?

https://github.com/historicalsource/zork-german/blob/master/rmode.init

larsbrinkhoff:
Missing manual:   
"RMODE: A Real-time Edit Facility" (SYS.04.07-1), P. David Lebling, R. V. Baron, Bruce K. Daniels, 1977.

[...]

PR #1616, opened 2019-06-03
RMODE - text editor.

larsbrinkhoff:
This editor is a collection of TECO macros.  It's one of the forerunners to EMACS.  It was popular on the DM machine.  (And maybe Infocom.)


larsbrinkhoff:
Currently only a set of binary files.  See #1614 for some more information.  Maybe source code can be found in the future.

PR #1617, opened 2019-06-04
TST342 - display all Type 342 characters.

larsbrinkhoff:
This will run on a PDP-6 or 10 on ITS or out of timesharing.

The resulting binary is named MAINT; TST342 BIN.  Since it's both a timesharing and standalone program, TS TST342 or @ TST342 wouldn't be appropriate.


larsbrinkhoff:
CC @SMJ, this could be used to test a 342 character generator.

PR #909, opened 2019-06-15
FAILSA - Failsafe reader

larsbrinkhoff:
MOON; FAILSA 9


rcornwell:
What commit are you using for the simulator? Note all the tapes in http://sky-visions.com/dec/tops503/ which are all failsa tapes.

larsbrinkhoff:
It wasn't the latest commit with the tape updates, so that may explain it.  Anyway, the program read files just fine so I suppose it's OK.

larsbrinkhoff:
To reviewers: The binary historically was named MOON; FAILSA BIN.  There never was a TS FAILSA as far as I can see.  I think it's somewhat appropriate to keep it this way, since it's not exactly a user friendly program.

larsbrinkhoff:
It would be nice to be able to read BACKUP tapes too.  I'm making a stab at guessing the format:

- The entire tape (save set?) is one tape file.
- The first record is a header of some kind.  Tape label?
- The next record is a file header.  The file name is at offset 42 (octal). Two words of ASCII, another word, and then one more ASCII word with the file extension.
- Following records are file data, with 40 words of header in each.

PR #1629, opened 2019-06-15
Damaged file

### LINKED ISSUE #162: NSALV corrupts disk
If we try to load lots of source files using NSALV, the resulting system displays random failures.  The failures come and go, and show up in various places.  Also, it depends greatly on the host computer, with virtual machines perhaps being the worst.

The theory is that this is a timing-dependent bug in NSALV which corrupts the file system.

larsbrinkhoff:
Recently, the build fails intermittently due to SYS; PURQIO 2138 being damaged.  I believe this is because the initial TRAN$G from NSALV now loads more files.

The proper thing would be to fix #162, but as a stopgap measure, reduce the number of files on the minsys tape.  E.g. TS GAS and PURQIO 2138 doesn't need to be there.


larsbrinkhoff:
I made a new directory, bin/minsys, and moved files that *do* need to be in minsys there.  The minsys tape is now made from bin/\<cpu>/_, bin/\<cpu>/sys and bin/minsys/sys.  The other files stay where they are.  The Makefile BIN variable now includes sys, which is where PURQIO and TS GAS are.

larsbrinkhoff:
Thanks for asking. :-)

PR #829, opened 2019-06-17
EMACS CHART bug

larsbrinkhoff:
The generated file EMACS CHART has a Y2K bug.  The year is supposed to be two digit but overflows into the hundreds.


larsbrinkhoff:
The EG code at SYMDAT is somewhat hairy to fix, and EG is declared to be semi-obsolete anyway, so I'll just fix FS FD CONVERT$ for now.

PR #1558, opened 2019-06-17
Two NTSDDT, one overwritten.

larsbrinkhoff:
The misc.tcl script assembles .; @ NTSDDT twice.  The first time from SYSENG; NTSDDT (KA10 version with 340 support), the second from SYSTEM; DDT (KL10 version with disk support).  Obviously the second overwrites the first.


larsbrinkhoff:
Proposed solution:

Move assembling of @ NTSDDT to the <cpu>/include.tcl file.  KS10 doesn't change; use SYSTEM; DDT > as before.  KA10 gets both SYSTEM; DDT and SYSENG NTSDDT.  KL10 gets SYSTEM; DDT.

The KL10 include.tcl is just a placeholder for now, so the KL10 @ NTSDDT will actually not get built.  However, @rcornwell will soon have KL ITS running on his simulator!

PR #1634, opened 2019-06-27
Boot off MAGDMP tape

larsbrinkhoff:
The `magfrm` tool has been updated to make a bootable magtape.  There is no longer any need for a MAGDMP paper tape.


larsbrinkhoff:
The CircleCI build fails because it doesn't update the subrepository with magfrm.

larsbrinkhoff:
The updated magfrm just mimics what the real MAGFRM does:

1. Write a 16-word bootstrap first on the tape.
2. Open a MAGDMP binary file and skip words until the first JRST 1.
3. Copy the rest of the file to tape.
4. Write tape mark.

PR #1637, opened 2019-07-01
Make KL10 front end file system

larsbrinkhoff:
Store files in a way which make them visible to the KL10 front end.

The files are also visible in the ITS .KLFE. directory.  The directory data is stored in the file .KLFE.; KLDCP (DIR).  This file, all other front end files, needs to be in contiguous sectors on disk.  The directory is 64 sectors.

There is a pointer to the directory from the HOM block, which is in sector 1 and 10 on the very first track (cylinder 0, surface 0).


larsbrinkhoff:
The KLDCP DOC file looks strange, but it's fine when read from ITS.

larsbrinkhoff:
I'll say wow when we can take a disk image and boot it on MIT-MC!

PR #1640, opened 2019-07-04
Bug fix for TCP from Dave Conroy

larsbrinkhoff:
I compared Conroy's SYSTEM files against the originals, and found this bug fix.  %PKFLS is a 36-bit value, so its halves need to be swapped.


larsbrinkhoff:
Hey, maybe *this is the #1643 bug*!

eswenson1:
Now I have to apply this fix on ES and see if I ever crash again with the same error.

eswenson1:
Since applying this fix to ES, I have gotten this error twice again.  So this fix isn't a complete fix.

PR #1639, opened 2019-07-13
Update configuration for KA10 simulator

larsbrinkhoff:
@rcornwell wrote:
> For ITS you should disable RPA, RPB, TUA, FHA, DPB.

That's a fair point, those devices aren't in use.  However, I wonder what ill effects you have in mind from leaving unused devices enabled?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,7 @@ $(SIMH):
 	$(MAKE) -C tools/simh pdp10
 
 $(KA10):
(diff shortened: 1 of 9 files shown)


larsbrinkhoff:
Is CD a recent addition?

```
> set cd dis
Non-existent device

rcornwell:
Not sure what CD is... not a device. Might be DC.

PR #1650, opened 2019-07-16
Have "KL" ITS boot directly into NTS DDT.

larsbrinkhoff:
This avoids using DSKDMP for pdp10-kl, aka KL ITS (but really still KA10).  This brings it closer to how MC was booted, e.g. load NTSDDT from KLDCP.

CC @eswenson1


eswenson1:
You assigned this to me. Did you intend to add me as a reviewer?

larsbrinkhoff:
Sorry, yes.

larsbrinkhoff:
The pdp10-ka and pdp10-kl variants now hang unexpectedly in the middle of the build.  Since I updated the simulators from @rcornwell's upstream, I'll have to check to see if that is the cause.

larsbrinkhoff:
I rolled back to an earlier version.

PR #1652, opened 2019-07-18
New quantum timer simulation sometimes fails.

larsbrinkhoff:
Checking to see if something from @rcornwell's latest upstream changes has introduced a regression in the ITS build.


larsbrinkhoff:
It's the GitLab build that is under examination.

(CircleCI is stupid about subrepositories so disregard those.  Travis only does a subset of the full build, so it doesn't trigger the failure.)

larsbrinkhoff:
Background info.  I first saw builds failing on the lars/dc-10 branch but figured it had something do with the DC-10 simulation.  But then I also saw it on the lars/kl-ddt branch, pull request #1650).  Both branches have in common that they use a KA10 simulator based on the latest "sims".  I haven't seen this failure with an older version of the simulator.

rcornwell:
I did a build yesterday from the master branch with my current sims repo and the build completed fine.

larsbrinkhoff:
I'm hoping everything's OK.  But I'll be running these tests to make sure.

larsbrinkhoff:
The baseline commit checked out OK.

larsbrinkhoff:
Latest upstream master (14acd52b83eddd6e15b18e8fb6cd89605ac34bc2) gets this:

```
*:klfedr write .temp.; kldcp bin
MAR; 5412>>.IOT 4,7
PURPG;PURINS;MPV; 5412>>.IOT 4,7   
The last command timed out.

**Comment by larsbrinkhoff** [comment]
@rcornwell, some change in your repository does seem to have introduced a problem which breaks the ITS build.  I have seen this on GitLab and my laptop.  I will locate the exact commit and report back.

**Comment by larsbrinkhoff** [comment]
44f8310 "PDP10: Cleanup invalid argument." did not pass.

**Comment by rcornwell** [comment]
What is strange is that I can build from master branch with current sims without problem.

**Comment by larsbrinkhoff** [comment]
Yes, that is strange.  But I have noticed that sometimes the build will only fail on some specific machines.  Maybe differences in operating system, libraries, compiler, etc.

**Comment by rcornwell** [comment]
I re-cloned the repo and now I can't build anything. dasm is failing, klh10 is failing. etc

**Comment by larsbrinkhoff** [comment]
That's even stranger!

**Comment by larsbrinkhoff** [comment]
This one seems OK: 58d34a3 "KA10: Start of IMP support for WAITS. Not working yet."

**Comment by larsbrinkhoff** [comment]
So it's down to 54d1af5 - 12b5035.  Out of those six, 12b5035 seems most likely to be the culprit.  Maybe the quantum timer?

**Comment by larsbrinkhoff** [comment]
OK: 33c0fc7 SCP: "Updated to current."

Now testing the commit immediately before 12b5035.  Seeing as those two commits don't make any PDP10 changes, they should be ok.

**Comment by larsbrinkhoff** [comment]
585e372 "IBM360: Cleanup compile errors." is ok on my laptop.  Still waiting for GitLab.

**Comment by larsbrinkhoff** [comment]
Testing with the new quantum timer reverted.

**Comment by larsbrinkhoff** [comment]
@rcornwell, I believe something about your new quantum timer implementation broke the ITS build.  I'm testing this change: https://github.com/larsbrinkhoff/ka10-simh/commit/8a2c52c28ae9d7d5088202d7a4a5c6f1f3ef70ab, which revers back to the pre-12b5035 implementation.

I see the quantum timer has something to do with MAR break.  And I did see this a few times:

```
*:klfedr write .temp.; kldcp bin
MAR; 5412>>.IOT 4,7
PURPG;PURINS;MPV; 5412>>.IOT 4,7
```

But other times the build just hung, or there were other errors.  Does this clue tell you anything?

**Comment by rcornwell** [comment]
The quantum timer should be ok, also it should be more accurate. This uses the new timer code. I don't want to back up to older code.

**Comment by larsbrinkhoff** [comment]
I'm not saying back up.  I'm asking, can you tell anything from the clues?

**Comment by rcornwell** [comment]
I am not in town today will see what I can tomorrow. I do remember ITS does not really need the timer. Also it builds fine on both my systems. The timer generally never goes off.

**Comment by larsbrinkhoff** [comment]
Circle CI and Travis builds fine with the new quantum timer.

GitLab and my laptop do not.

**Comment by larsbrinkhoff** [comment]
I don't understand the code yet, so this is mostly a random shot in the dark.  But maybe this suggests something is strange about fault_data:

```
*:klfedr write .temp.; kldcp bin
MAR; 5412>>.IOT 4,7
PURPG;PURINS;MPV; 5412>>.IOT 4,7
```

Here we see some unexpected exceptions: MAR, PURPG, PURINS, and MPV.

I understand fault_data and the quantum timer occupy the left and right halves of a pager word.

**Comment by rcornwell** [comment]
Problem is, until I can reproduce this I can't track things down. The quantum timer change just computes the time until a timeout should occur and set a wakeup event that number of us in the future. Other then this the code is basically the same. Also quantum timer rarely interrupts the system. When the timer is read, the code computes what the timer value should be and returns that.

**Comment by larsbrinkhoff** [comment]
It seems to me the old implementation triggered on overflow to bit 18:

```
        if ((qua_time & 01000000LL) != 0) {
            mem_prot = 1;
            fault_data |= 1;
        }
```

But the new code uses RSIGN, so overflows when bit 17 is set.  Right?

As for which is correct, I have no opinion.

**Comment by larsbrinkhoff** [comment]
My reading of hardware memos 2 and 2A is that the former says bit 17 is overflow, but the latter says bit 18 is overflow.

**Comment by larsbrinkhoff** [comment]
I inserted an fprintf so show when the quantum timer overflows.  It seems to happen with long running CPU-bound jobs.  In the context of the build script, this is mostly MIDAS assembling large programs.

**Comment by larsbrinkhoff** [comment]
For example, assembling "@" triggers ten times on my laptop.

**Comment by larsbrinkhoff** [comment]
It does seem like the unexpected MAR etc does happen after a quantum timer overflow.

```
[QUANTUM OVERFLOW]MAR; 27145>>SKIPGE 1017
PURINS;(PURPG;MPV;) 27145>>SKIPGE 1017

larsbrinkhoff:
Using the "bit 17 patch" I had one successful build on my laptop, but GitLab still fails.  I get this:

```
*:c;occ c1.c c21.c c22.c c23.c c24.c c25.c c26.c
c1.c:
Unhandled interrupt, continue?  
The last command timed out.
```

"Unhandled interrupt" possibly suggests something about fault_data.

More investigation needed.

larsbrinkhoff:
Sometimes this fprintf triggers.

```
uint32
get_quantum()
{  
    uint32  t = 0;
    if (sim_is_active(&cpu_unit[1])) {
       double us;
       us = sim_activate_time_usecs (&cpu_unit[1]);
       if ((uint32)us >= qua_overflow) {
           fprintf (stderr, "[QUANTUM ERROR 2]");
           fflush (stderr);
       }
       t = qua_overflow - (uint32)us;
    }
    return t;
}  
```

This ought to be bad considering the use of the return value:

`MB = ((uint64)get_quantum()) | ((uint64)fault_data) << 18;`

I think this may explain why MAR, PURPG, PURINS may be set.

larsbrinkhoff:
@rcornwell, your fix seems to agree with the GitLab machines.  Thanks!

PR #1666, opened 2019-07-29
Added support for RP06 disks with RH10 disk controller. 

### LINKED ISSUE #1648: RH10 -> RP06 Support
I added conditionalization to SYSTEM; RH10 > to support either the existing RP04 geometry or RP06 geometry.  I then built DDT, SALV, ITS, and DSKDMP with this support.  I created a MAGDMP tape with magdmp.bin (unchanged), xddt bin, and xsalv bin.  

I booted MAGDMP, loaded ddt and xsalv, and ran MARK$G.  SALV hung after the PACK NO = prompt.  Since no ddt disk code should have been executed here, and only SALV code, and my SALV config was identical to that of pdp10-kl -- except for the RHRP06==1 setting that selected the RP06 geomery, I'm wondering if there is something in the RH10 support that would not accommodate this geometry (from my updated RH10 >):
```
;DISK PHYSICAL PARAMETERS (RH10 CONTROLLER)

IFE RHRP06 [
DEFSYM  NCYLS==406.             ;# CYLINDERS NORMALLY USED (RP04)
DEFSYM  XCYLS==411.-NCYLS       ;# CYLINDERS FOR SPARES, HACKS, ETC.
DEFSYM  NHEDS==19.              ;# TRACKS/CYLINDER
DEFSYM  NSECS==20.              ;# SECTORS/TRACK
DEFSYM  SECBLK==8               ;# SECTORS/BLOCK
]
IFN RHRP06 [
DEFSYM  NCYLS==812.             ; # CYLINDERS NORMALLY USED (RP06)
DEFSYM  XCYLS==815.-NCYLS       ; # EXTRA CYLINDERS FOR SPARES, HACKS, ETC.
DEFSYM  NHEDS==19.              ; # TRACKS/CYLINDER
DEFSYM  NSECS==20.              ; # SECTORS/TRK
DEFSYM  SECBLK==8               ; # SECTORS/BLOCK
]
DEFSYM  NBLKSC==NHEDS*NSECS/SECBLK ;# BLOCKS/CYLINDER
DEFSYM  NBLKS==NCYLS*NBLKSC     ;# BLOCKS NORMALLY USED
DEFSYM  XBLKS==XCYLS*NBLKSC     ;# BLOCKS FOR HACKS
DEFSYM  TBLKS==NBLKS+XBLKS      ;TOTAL BLOCKS
```
@rcornwell any ideas?

Note: it mattered not whether I used:
   DEFSYM  NTUTBL==2               ;2 BLOCKS PER TUT
or
  DEFSYM  NTUTBL==4               ; 4 BLOCKS PER TUT
Both caused the same hang.

eswenson1:
Moved RP04 disk parameters to separate file (system;rp04 >) from system;rh10 >.
Made build default to RP04 when RH10 is specified, but switches
can select RP06 for relevant components.  Resolves #1648.


src/system/ttytyp.319
```diff
@@ -184,6 +184,77 @@ MCONDX KA,{				;;; KA
 	TTDAAA 37,SPEED=9600		;T37
 	TTDAAA 40,SPEED=9600		;T40
 .ENDC ;}
+MCONDX KL,{				;;; KA
```
larsbrinkhoff: Why add a new section to TTYTYP?  Line 187 is a new KL, but the comment says KA.  Line 348 is the old KL, now renamed to KX, but the comment says KL.

build/kl10/include.tcl
```diff
@@ -66,11 +71,12 @@ proc make_dskdmp {} {
...
     respond "HRIFLG=" "N\r"
     respond "BOOTSW=" "N\r"
-    respond "RP06P=" "N\r"
-    respond "RP07P=" "N\r"
+    respond "R11R6P=" "N\r"
+    respond "R11R7P=" "N\r"
     respond "RM03P=" "N\r"
     respond "RM80P=" "N\r"
     respond "RH10P=" "Y\r"
+    respond "R10R6P=" "Y\r"
```
larsbrinkhoff: RP10R6P=Y seems to me DSKDMP is assembled for RH10 and RP06, right?

src/system/ttytyp.319
eswenson1: I'll fix. I'll have to look at what I did. I think I did that because I wanted to use the KL build with terminals, and you hadn't defined any. But I can put it back because I've added the channels to the EX TTYTYP entry.  I suspect the "KA" comment for KL was there originally.  I remember fixing a few comments after machine definitions since they were wrong.

build/kl10/include.tcl
eswenson1: Whoops.  Must be that we're not using this now -- since the build worked fine and the resulting system works.  I'll fix that.

src/system/ttytyp.319
eswenson1: So what I had done was this:  renamed your original KL entry to KX (to get it out of the way and preserve it); copied the KA entry to KL so that I could login to a KL system; added an EX entry once I had built/tested/deployed an EX system.  I forgot to get rid of the temporary KL entry and rename KX back to KL.  I've done that now.

build/kl10/include.tcl
eswenson1: I just fixed this (changed the Y to an N).  I'm going to retest the KL build locally.

eswenson1: I got past the build of DSKDMP in my local build. Our change to use DDT rather than DSKDMP now means that we don't test out the generated DSKDMP. Maybe we should load DSKDMP from DDT during the build one time and use it just to test out that it works.

src/system/ttytyp.319
larsbrinkhoff: Thanks!

build/kl10/include.tcl
larsbrinkhoff: Yes, good idea!

build/ks10/include.tcl
```diff
@@ -92,6 +92,7 @@ proc its_switches {} {
 proc make_ntsddt {} {
     respond "*" ":midas dsk0:.;@ ddt_system;ddt\r"
     respond "cpusw" "3\r"
+    respond "dsktp=" "1\r"
```
larsbrinkhoff: Since dsksw=0, I believe the value of dsktp doesn't matter.  Maybe the dsktp question shouldn't come up at all?

src/system/ddt.70
```diff
@@ -66,6 +66,14 @@ setf [Type of disk? (0 none, 1 Systems Concepts, 2 RP-10, 3 RH-10)]dsksw
 	define rh10
 	ife dsksw-3!termin
 
+ifdef rh10,[
```
larsbrinkhoff: I see dsktp= indeed should only be asked if rh10 is set.  So how come it shows up in the ks10 build?

larsbrinkhoff:
And thanks for fixing some bugs I left behind!

build/ks10/include.tcl
eswenson1: I'll enter a separate ticket on this.

src/system/ddt.70
eswenson1: I'll enter a separate ticket on this.

PR #1674, opened 2019-08-02
Fixed COMSAT and DQXDEV on KA10. 

### LINKED ISSUE #1669: DQX device doesn't work on pdp10-kl emulator (KA10s)
The DQ device (actually DQX) code appears to rely on the KS10 or KL10 processors. It fails on KA10 (and presumably KI10) processors.
 
It uses the adjbp instruction, which apparently is not supported on KA10 and KI10 processors:
 
```
IBP Increment Byte Pointer

Increment the byte pointer at location E, setting the byte alignment to zero
if the incrementing crosses a word boundary, as explained above.
Notes. Giving this instruction code with bits 9-12 nonzero produces the
ADJBP instruction described at the end of the section. In the KIlO and
KAlO, only the IBP form is available and bits 9-12 are ignored (but should
be zero).
```
 
MIDAS is correctly setting the 9-12 bits to non-zero:
 
```qt.ptr+14/   IBP 1,2225   =133040002225```
 
but since these are ignored on KA10 processors, the instruction is being treated and IBP instruction rather than an ADJBP instruction.
 
This results in PTR requests failing. This causes COMSAT to fail. 
 
I think we need to either a) rewrite this in an processor agnostic manner or b) conditionally assemble this based on target processor.

### LINKED ISSUE #1671: MAIL doesn't work on pdp10-ka emulator with chaosnet networking
I built an ITS with chaosnet networking (and RH10 disk controller, TM0B tape controller, and 3 RP06 disks drives) and can run it under the pdp10-ka.  

COMSAT .values on startup.  It also .values on MFINIT$G.  It gets an MPV fault either while processing .MAIL.; NAMES > or while processing any queued email message.

(In order to get this far, you have to update the DQ (DQX) device to work on a KA-10 processor.  See #1669 for details on how to do that).

You can get past the MPV during MFINIT by commenting out the following lines in .MAIL.;NAMES >:

```
;; Hosts that can receive *msgs
;;(*EX (EQV-LIST *MSGS-TO-ITSES))
 
;;(*MSGS-TO-ITSES (R-OPTION NOTDIST)
;;  (EQV-LIST (*MSG-SINK@EX (R-OPTION NOTDIST))))
```

(It doesn't matter what the hostname is -- DB or EX fail in the same way).

However, once the COMSAT initialization files are created with MFINIT, then running COMSAT normally will die with an MPV when processing any mail request.  

The ITS I'm using has 768k of memory.  I tried with a 1024k ITS, and although the system appeared to run fine, COMSAT had other issues. It isn't obvious to me that COMSAT needs more memory -- it ran fine on real ITS systems of less than 1024k. 

I believe the MPV is occurring in the ULNDEL UUO.  This UUO frees an LN entry.

Here is a trace of the ULNDEL code causing the MPV:

```
$1B; ULNDEL>>MOVE U3,U40   U3/   41   U40/   DMOVE C,27336      ^N
ULNDEL+1>>LDB U2,14777   U2/   50   14777/   ADD N,U3      ^N
ULNDEL+2>>JUMPE U2,15741   U2/   C      ^N
ULNDEL+3>>HRRZ U1,(U2)   U1/   20   C/   .IOT U4,5461      ^N
ULNDEL+4>>TRNN U3,-1   U3/   DMOVE C,27336      ^N
ULNDEL+6>>MOVE U2,(U3)   U2/   C   27336/   ORCMM OC,@452222(L)      "$0"MULTI$!
   ^N
ULNDEL+7>>CAIN U1,(U2)   U1/   5461   E.A. _ 452222   ^N
ULNDL4>>MOVE U3,U2   U3/   DMOVE C,27336   U2/   ORCMM OC,@452222(L)      ^N
ULNDL4+1>>MOVE U2,@32(L)   U2/   ORCMM OC,@452222(L)   557026/   ??    ^N
(MPV;) ULNDL4+1>>MOVE U2,@32(L)   U2/   ORCMM OC,@452222(L)   557026/   ??
```
This suggests that the LP to the next LN is bad (bad list pointer).

### PR DESCRIPTION
**Pull request by eswenson1**
Conditional assembly prompts for limiting to KA10 instructions. Also fixed DQXDEV (DQ) device to work on KA10.

Resolves #1671. Resolves #1669.

### DIFF
--- a/src/sysnet/netrts.357
+++ b/src/sysnet/netrts.357
@@ -6,6 +6,8 @@ IFNDEF $$FTP,$$FTP==0	; Do not assemble FTP routines anymore.
 IFNDEF $$TCP,$$TCP==1	; Normally assemble TCP routines.
 IFNDEF $$DQ,$$DQ==0	; Normally use host table, not DQ device.
 IFNDEF $$450,$$450==0	; Disable special casing of SMTP reply code 450
+IFNDEF $$KA10, $$KA10==0 ; default to KS/KL-10 instructions
+
 comment|   
          This file contains the network hacking code of COMSAT.
 
@@ -2165,12 +2167,19 @@ NHITS9:	POPAE P,[D,B,A]
 
 ; "NHMLTX" - routine skips if host in A is a Multics.
 NHMLTX:	PUSHAE P,[A,B,C,D]	; Save acs
-	DMOVE C,[ASCII "MULTICS"]	; What we are looking for
+
+IFN $$KA10,[
+	MOVE C,(B)
+	MOVE D,1(B)
+]
+IFE $$KA10, DMOVE C,[ASCII "MULTICS"]	; What we are looking for
+
 	JRST NHOPSY		; Join common code
 
 ; "NHITS" - routine skips if host in A is an ITS.
 NHITS:	PUSHAE P,[A,B,C,D]	; Save acs
-	DMOVE C,[ASCII "ITS" ? 0]	; What we are looking for
+	MOVE C,[ASCII "ITS"]	; What we are looking for
+	MOVEI D,0
 ;	JRST NHOPSY		; Join common code
 
 
--- a/build/ka10/include.tcl
+++ b/build/ka10/include.tcl
@@ -217,3 +217,7 @@ proc patch_clib_16 {} {
 
 proc copy_to_klfe {file} {
 }
+
+proc comsat_switches {} {
+    respond "Limit to KA-10 instructions" "y\r"
+}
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/sysnet/netrts.357 [hunk 1]
```diff
@@ -2165,12 +2167,20 @@ NHITS9:	POPAE P,[D,B,A]
...
+	MOVE D,1(B)
+]
+IFE $$KA10, DMOVE C,[ASCII "MULTICS"]	; What we are looking for
+
 	JRST NHOPSY		; Join common code
 
 ; "NHITS" - routine skips if host in A is an ITS.
 NHITS:	PUSHAE P,[A,B,C,D]	; Save acs
-	DMOVE C,[ASCII "ITS" ? 0]	; What we are looking for
+;	DMOVE C,[ASCII "ITS" ? 0]	; What we are looking for
```
larsbrinkhoff: Please remove this comment.

Issue #1468, opened 2019-01-04
Figure out how to use the OJB device


eswenson1:
Turns out there is an OJB device (in addition to the JOB and BOJ devices).  It *appears* that the difference between the JOB device and the OJB device is that whereas JOB launches a job to satisfy the request, OJB searches for an existing user job to satisfy the request. This may well be a way to debug JOB handlers (e.g. DEVICE; ATSIGN MLDEV).  My theory is that you could start up an MLDEV (normal user job) under DDT and then use OBJ: rather than JOB: to use it as the handler for an I/O request to device.  

SYSDOC; JOB 100 describes the JOB and BOJ devices, but there is no mention of the OBJ device.  I did find references to it in SYSTEM; ITS 1648 and the comments in ITS suggest that it behaves as mentioned above.

larsbrinkhoff:
The above says both OBJ and OJB.  The latter is correct, right?

eswenson1:
Right. OJB is correct. I can’t ever type that correctly!

larsbrinkhoff:
I see OPTOBJ is defined, but the symbol is never used.  At least it doesn't show up in searches.

larsbrinkhoff:
Oh sorry, it's %OPOBJ.

eswenson1:
I haven't been able to get this to work.  I created a job (MLDEV) in one HACTRN, logged in as EJS and loaded device;atsign mldev.  In another HACTRN (logged in as EJS0), I created a translation between JOB:device;atsign mldev and OJB: EJS; EJS MLDEV.  I then tried to do a ^R NO:lars; and just got:

```
NO: LARS; EJS MLDEV - DEVICE NOT AVAILABLE
```.

However, if I do a ^R OJB:EJS;EJS MLDEV, I return from the BOJ open call in the BDH -- so that definitely does work.  However, I haven't been able to find a translation that works so that doing a ^R no;lars; invokes my MLDEV BDH (under DDT).

I've even tried a translation of JOB:*;* * -> OJB:EJS; EJS MLDEV.  I know that target works (see above), but I can't get an ^R NO;LARS;, which should cause JOB:DEVICE;JOBDEV NO to get invoked, to work.  I've tried the following translations:

```
DB ITS 1648  Peek 630   1/8/2019 15:40:22  Up time = 4:16:43:37
EJS0 HACTRN & inferiors:
   IO   JOB:DEVICE;ATSIGN MLDEV =>  OBJ:EJS;EJS MLDEV
[...]

PR #1687, opened 2019-08-12
Tickets #1468 and #1685

### LINKED ISSUE #1685: Fix Other menu item info INFO
The menu "Other" doesn't work properly in INFO.  This is because there is no Node: section for it.  There seems to be some strange INFO formatting in this file -- not sure why -- where topics are duplicated.  I won't fix that, but I will add a Node: for Other so that it can be found from the top-level menu.

eswenson1:
Resolves: #1468. Adds INFO topic Tips that includes various tips on using and administering ITS, including one on how to debug OJB devices.

Resolves #1685: "Other" menu in INFO directory doesn't work properly.

### DIFF
--- a/doc/info/dir.230
+++ b/doc/info/dir.230
@@ -336,6 +336,7 @@ Operating System (ITS):
 			XXFILE is similar to XFILE except that it allows
 			you to get your tty back while the file is being 
 			executed
+
 Simulators:
 
 * DECUUO: (DECUUO)	DECUUO is a program which emulates the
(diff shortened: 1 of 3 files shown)


doc/info/dir.230
```diff
@@ -22,7 +22,7 @@ File: DIR	Node: Top	This is the top of the INFO tree
 * Debugging facilities::
 * Terminal Support::
 * ITS: ITS
-* Simulators::
+* Simulators::
```
larsbrinkhoff: Is this added space needed?

doc/info/dir.231
```diff
@@ -24,6 +24,7 @@ File: DIR	Node: Top	This is the top of the INFO tree
 * ITS: ITS
 * Simulators:: 
 * Other::
+* Tips: (ejs;tips >)Top
```
larsbrinkhoff: The TIPS file is missing from the commit, right?

eswenson1:
Whoops, I did this so many times I forgot tips and the fact that I moved it to info:. Will fix

larsbrinkhoff:
I moved the space character fix from the second to the first commit, rebased, and force pushed.

PR #1689, opened 2019-08-14
Use the Datapoint kludge

larsbrinkhoff:
Telnet to port 10002.  Ports are 4800 baud.


larsbrinkhoff:
Updated to latest KA10 with improved idling.

PR #1690, opened 2019-08-14
TWDEMO - scripted demonstration of SHRDLU.

larsbrinkhoff:
This is a program that just replays the dialogue in SHRDLU; DEMO FLICK.

CC @SMJ


larsbrinkhoff:
Not that I have found yet.  Searching for DEMO or FLICK in the SHRDLU files doesn't reveal anything useful.

eswenson1:
I assume all it is doing is firing up lisp, opening the demo file, and feeding lisp all the commands in the file (sort of like xxfile), right?

larsbrinkhoff:
Something like that, but not a bare Lisp.  It probably has some SHRDLU files loaded.  Note e.g. `(GRASP (QUOTE :B1))` to have the robot pick up a block.  GRASP is defined in TWUTIL.

larsbrinkhoff:
And text between @ characters is the dialogue.

larsbrinkhoff:
Hmm, the GRASP in TWUTIL isn't very useful!

larsbrinkhoff:
Ah, GRASP is in RBRN; GRAPH 26.

PR #1691, opened 2019-08-14
CI updates.

larsbrinkhoff:
The CircleCI KLH10 build seems to have met the same fate as @eswenson1's Jenkins and GitLab.  It fails just about every time, usually somewhere around purifying EMACS libraries.


larsbrinkhoff:
Also change the CirlcleCI job names from "job1" etc to more understandable "build_simh" etc.

PR #1703, opened 2019-08-19
Increase NAMDRG limits to accomodate larger system

larsbrinkhoff:
With the addition of 16 more Datapoint Kludge ports, the number of ttys has overwhelmed NAMDRG.   @eswenson1 found out that MAXTTY needs to be increased.

But there's still a problem at XGPSTS.


larsbrinkhoff:
@SMJ,

When I tvcon to hactrn.org, I get a blank display and F1 does nothing.  See this pull request for one more fix to NAMDRG.

PR #1710, opened 2019-08-28
Updated Emacs QSEND library to use Insert Date rather than Get Date.

### LINKED ISSUE #1709: QSEND uses non-existent EMACS macro Get Date
Get Date was renamed to Insert Date, according to EMACS; EMACS NEWS, in EMACS 141, in Nov 1979.

eswenson1:
This function was renamed in Emacs 141 (Nov 1979).  Resolves #1709.


eswenson1:
It is. s/Get/Insert/

PR #1714, opened 2019-08-30
Do something to alert users if DUMP has NEWCOD wrong

larsbrinkhoff:
It's very easy to miss setting NEWCOD to the right value in DUMP.  Do something to improve on this.

- Assembly-time check seems hard.  There is no information to say what the target processor is.
- Run-time check?  Determine processor, and barf is NEWCOD doesn't match.
- Select NEWCOD at run time.
- Remove NEWDOC==0 code in DUMP.  Requires updating ITS.


larsbrinkhoff:
I changed my mind.  DUMP already uses .EVAL so I added one more for KS10P.

eswenson1:
Not sure why you say we can’t check the target hardware at assembly time. I encountered code the other day that uses midas macros to figure out what Kx10 the code is being assembled on. We don’t do cross assembly in our build (except for, say PDP11 code). The code is assembled on an emulated machine reflecting the target hardware. I’ll look around for those macros and point you to them.

larsbrinkhoff:
When I say target, I mean what machine DUMP will run on.  If MIDAS is
ran on, say, AI, and you want to assemble DUMP for ML, how would MIDAS
know what processor ML has?

eswenson1:
Well, the only use of the MCOND macro in MIDAS is to control NEWCOD. And NEWCOD is only used to select KS10 or not KS10.  Whether the machine is a KS10 is determinable at assembly time.

larsbrinkhoff:
If you run MIDAS on AI, how can MIDAS know if ML is a KS10 or not?

eswenson1:
We do not build DUMP with MACHINE=AI, and expect to run it on ML.  We build DUMP with MACHINE=DB, and if we're running on a KS10 (e.g. SIMS or KLH10), then the binary that is produced will end up with NEWCOD=1 and only run on a KS10.  It is not intended to be copied to a KA10 and run there.  When we build DUMP with MACHINE=KA, on pdp10-ka, then the binary that is produced will end up with NEWCOD=0 and only run on a KA10 (or non=KS10).  It is not intended to be copied to a KS10 and run there.  

My point is that we are not doing cross architecture builds. It it irrelevant what the machine is here (AI, ML, DB, KA, KL).  What matters is whether the code uses KS10 instructions or whether it doesn't.  Or, more generally, whether it uses KA10, KL10, or KS10 instructions.  That we can determine at assembly time through macros that are cleverly written to work one way on each of these architectures.  

Any code that is conditionalized for a KS10 (like DUMP) will only work on a KS10.  So detecting that we're building on a KS10 emulator for a KS10 emulator will work.

larsbrinkhoff:
No, the binary will not end up with NEWCOD taken from the build machine.
NEWCOD depends on the target machine.

It's not unreasonable to build DUMP on one machine, for running on
another machine.  I do intend that DUMP can be copied around.  That is
exactly the use case I wanted to improve by this pull request.

This isn't something I invented.  DUMP will do a runtime check that it's
running on the machine it was built for.  If not, it will give the user
a slap on the wrist but continue running with some capabilities
disabled.

eswenson1:
We’ll just have to disagree here. There are two variables here: machine name and newcod. The latter generates code for ks10 code or not. This is totally independent from machine name, although historically the two were normally linked — at least for long periods of time. If you build with newcod =1 it just isn’t ever going to run on a KA10. So it makes no sense for the KA10 build to build with newcod=1, nor for the KLH10 build to build with newcod=0. The build for a KS10 is built on a machine which emulates the KS10. The build for KA10 is running on a machine emulating a KA10. Consequently, we can check at assembly time what Kx10 we’re running on, and set newcode appropriately. 

This is independent of the machine name.

Historically, if you were building for machine named AI, this meant KA10 (until AI was moved to a KS10). I think we still have the same case. DB is a KS10. KA is a KS10. And soon, KL will be a KL10.

larsbrinkhoff:
There has not been any restriction that a KA10 DUMP can not be built on
a KS10 ITS or vice versa.  I wouldn't want to add such a restriction.

eswenson1:
I think I just realized that we do want to be able to cross-assemble -- we need to produce a TS DUMP for KL10 (from some non-KL10 build) so that we'll have an ITS prepared for the KL10.

But the macros I saw do allow overridding -- for just this reason -- but if a processor isn't specified, can default to the one on which the assembly is running.

PR #1721, opened 2019-09-01
Update emulator for improved idling and throttling

larsbrinkhoff:
The -s argument is number of milliseconds to sleep ever 1000 PDP-11 instructions.

### DIFF
--- a/build/dependencies.sh
+++ b/build/dependencies.sh
@@ -11,9 +11,9 @@ install_linux() {
     sudo apt-get install -my git make gcc libncurses-dev autoconf
     case "$EMULATOR" in
         simh) sudo apt-get install -y simh;;
-        pdp10-ka) sudo apt-get install -y libegl1-mesa-dev libgles2-mesa-dev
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
@eswenson1, I retro fitted another commit to update the KA10 simulator which does 340 idling.

larsbrinkhoff:
I now get this.  I think the solution is to have libpcap-dev installed, but still, getting a linker error seems wrong.  Maybe add an #ifdef HAVE_PCAP_NETWORK somewhere?  CC @markpizz @rcornwell 

```
sim_ether.c: In function ‘eth_version’:
sim_ether.c:2441:1: warning: implicit declaration of function ‘pcap_lib_version’ [-Wimplicit-function-declaration]
 return pcap_lib_version();
 ^
sim_ether.c:2441:1: warning: return makes pointer from integer without a cast [enabled by default]
/tmp/ccDJqEoQ.o: In function `eth_version':
sim_ether.c:(.text+0x2817): undefined reference to `pcap_lib_version'
collect2: error: ld returned 1 exit status
make[1]: *** [BIN/pdp10-ka] Error 1
```

Issue #441, opened 2017-01-18
Machine name


larsbrinkhoff:
Every now and the I stumble upon a place which hardcodes the ITS machine names.  Those places should be recorded.

larsbrinkhoff:
SYSEN2; MAGFRM 17, label MCHNTB.

eswenson1:
This is a great idea. Eventually we should centralize.

-- Eric

larsbrinkhoff:
SYSEN1; PWORD >, label bltspc.

SYSEN1; DDT >, label KEXIST.

SYSEN1; SENDER >, labels Create and SC4.

SYSEN2; FRETTY >, label GO.

SYSEN2; FIND >, label ITSTBL.

SYSEN2; DIRED >, label ACT2.

SYSENG; DIRDEV >, label OPEN10.

SYSENG; DEMAIL >, label Begin.

SYSNET; TELSER >, label CHTELE.  <--- This one is especially funny!

SYSNET; SENDS >, label gotusr.

SYSEN1; NETIME >, label hosts.

INQUIR; INQUPD >, label NEWMAC.

SYSENG; RAYGUN >, label MCHNTB.

larsbrinkhoff:
SYSENG; MODEMS >, label mchtab.

SYSENG; @DEV >, label machtb.

BAWDEN; 11LOAD >, label ackrfc.

eswenson1:
SYSNET; COMSAT >, labels: bughst, domgat, and tcpgat.  Currently, binary is patched rather than source modified.

larsbrinkhoff:
SYSENG; DUMP >, macro MCOND.

SYSEN2; FUMP >, macro MCOND.

SYSTEM; CONFIG >, macro MCOND.

SYSTEM; TTYTYP >, macro MCONDX.

SYSTEM; IOELEV >, macro MCONDX.

SYSTEM; ITS >, macro MCOND.

SYSENG; LUSER >, macro MACTAB, label ITSTAB.

SYSTEM; SALV >, symbol MCHN.

KSHACK; NSALV >, symbol MCHN.

larsbrinkhoff:
From @bictorv's http://its.victor.se/wiki/rename:

SYSEN2; MLDEV >, many places after symbol GO.

L; *LISP >, symbol UDIRSET.

L; STRUCT >, `MRA [ITS` ...

bictorv:
Not only recoRded, but recoded, I think. But recording is the first step!

[...]

PR #1715, opened 2019-09-01
Rename MX back to MC.

larsbrinkhoff:
We all agree MC is the proper name for the KL10, right?


larsbrinkhoff:
Should look at all these places: https://github.com/PDP-10/its/issues/441

larsbrinkhoff:
I think this will do for now.

larsbrinkhoff:
OH wait, the IOELEV build script needs to be updated too.

PR #1750, opened 2019-10-14
Snug as a symbol table in a RUG

larsbrinkhoff:
When loading a program from a paper tape into RUG, the symbol table is lost.

Close inspection reveals that the first pass loads the application core image and stores it on the disk swap area.  The second pass writes the symbol table to core, but forgets to update the symbol table pointer.

I have applied and tested a fix to RUG; AR 613, but I need to check if this works for version 632 too.

Below L.LD2B find `ADD RELOC,L.ADR` and add:

```
        CMP L.ADR,O.SYME
        BHI L.LD3
        MOV L.ADR,O.SYME
```

L.ADR holds an address to a block of data being added to the symbol table.  This is compared against O.SYME which is the start of the table.  If L.ADR is lower, update O.SYME with the new value.  The symbol table grows towards lower addresses.


larsbrinkhoff:
Also fixes #1752.

PR #1084, opened 2019-11-05
LUNAR - Emacs library

larsbrinkhoff:
EMACS1; LUNAR 212  
MOON; LUNAR 222  
MOON; LUNAR :EJ


larsbrinkhoff:
@rcornwell, the kl10 build is prone to failing in the "load moby tape" part due to taking longer than 10 minutes.  Travis CI doesn't like that.  Howver, the ka10 build does fine.  Do you know why there would be a difference?

PR #1772, opened 2019-12-01
Reverted EMACS1;FILES 435 to EMACS1;FILES 434.

### LINKED ISSUE #1771: Emacs says "IFN     Illegal FS flag name?"
Whenever I try to execute some TECO code from Emacs, I get the message `IFN     Illegal FS flag name?`

Example 1:  type M-% for query replace.  Enter a string, altmode, another string, double altmode.  A wild error message appears.

Example 2: Type double altmode.  Enter e.g. 65I$$ to insert ASCII character 65 into the buffer.  Same error message.

eswenson1:
The changes in version 435 seem wrong in all cases.  And reverting
to version 434 fixes the Illegal FS flag name? error that results
when trying to execute TECO commands in the minibuffer.  It seems
like version 435 includes references to FS flag names that are not
properly terminated with $ -- hence the errors.  Resolves #1771.


atsampson:
Looks good to me. [its-vault](https://github.com/PDP-10/its-vault/tree/master/files/emacs1) has FILES 439 and FILES 440 which seem to be less mangled.

eswenson1:
My guess is that those fixes were malformed. I’ll look into it and enter another issue.

eswenson1:
I looked at FILES 440 and I see the same issue with:
```
.,.m(m.m&_Save_for_Undonsert_File^[
```
That looks mangled too.  

It would appear that you can do FS ECHO LINES"E ...' (without the terminating $), so that, in itself appears not wrong.

One of the changes in 435 appears to be that when we are writing to the echo area, we emit the string we want, then conditionally emit a CRLF depending on whether FS ECHO LINES is Non-zero (435) or FS ECHO LINES-1 is Non-zero (434).  I can't figure out which is correct.  FS ECHO LINES defaults to 3 (I think), in which case either of these two will emit the CRLF.  However, when I set FS ECHO LINES to 1, both versions appear to work fine for me.  And both versions appear to work sub-optimally when FS ECHO LINES is 0.  So I don't know which is correct.  Any ideas?

eswenson1:
The only relevant change in Write File (from 434 to the patch in EMACS;PAT162 12) is this:

```
    @FTWritten:_^]1^[  FS ECHO LINES"N @FT
^['
```
to
```
    @FTWritten:_^]1^[  FS ECHO LINES^[-1"N @FT
^['
```
So if we could determine whether this change is good, we could either create a FILES 435 with just this change in it, or we could add an EMACS;PAT163 1 with that single patch in it.

But somehow, when I created the original EMACS1;FILES 435 I introduced a bunch of bad stuff.  I think I found a FILES 435 somewhere and used it -- I don't think I would have made such hash of the file.  I would have done a diff between the Write Files implementations in FILES 434 and the patch file, and would have just copied the patched version.  The FILES 435 I had had more changes (and broken changes -- some of which I see in FILES 439 and FILES 440 from its-vault.

eswenson1:
See #1773 for a new ticket to update to FILES 440 (with fixes to damage in the file from its vault).

PR #1775, opened 2019-12-10
Updated emacs1;files 434 to version 440.

### LINKED ISSUE #1773: Update EMACS1;FILES 434 to later versions
In its-vault, we have FILES 439 and FILES 440, which appear to be later versions of FILES 434. They both have at least some file damage and cannot be used directly, but with inspection and correction, we should be able to pull in some fixes. As incorporating FILES 435 into the build (#1520) caused issues (IFN Illegal FS flag name?), and that file did have the same damage as FILES 439 and FILES 440, we need to edit FILES 440 to correct the damage and ensure that it doesn't break minibuffer commands.

eswenson1:
This is the latest version of emacs1;files that we have and appears
to have some fixes. It does appear to work. Resolves #1773.


larsbrinkhoff:
Have to restart the KL10 build a few times.  As I mentioned before, it's slow extracting the big sources tape.  When Travis CI doesn't see any output in five minutes it thinks the build has hung.  I do wonder why this doesn't happen to the KA10 build.  Maybe split the tape in two?  Awkward workaround, I know.

PR #1789, opened 2020-01-10
increase timeout for backup portion of the build

pgalbraith:
I needed this extra time when building for pdp10-ka


larsbrinkhoff:
Thank you.  We usually let the submitter merge the pull request when it's approved and CI builds pass.

PR #937, opened 2019-01-17
Enable IMX

larsbrinkhoff:
Make a list of IMX channels.  Check which channels are used by applications.

PR #1767, opened 2020-01-11
Update KA/KL simulator

larsbrinkhoff:
Update subrepo with @rcornwell's KA/KL simulator.  Maybe when the kl10 branch is merged?

Check if 7cd2eb2 is fixed and can be reverted.

### DIFF
--- a/build/pdp10-ka/run
+++ b/build/pdp10-ka/run
@@ -45,4 +45,6 @@ set mta type=b
 set dt mpx=6
 set imp mpx=4
 set dpa noheaders
+set wcnsls enabled joystick
+set ocnsls enabled
 b ptr


larsbrinkhoff:
@rcornwell testing your latest "kl10" commit now.

larsbrinkhoff:
Here we go with the merged branch.

larsbrinkhoff:
Looks like the 340 no longer works.  I'll run some tests.

larsbrinkhoff:
It's very strange, but I when I tried it now I haven't been able to make the 340 work in any version I tried, old or new.  The symptoms are that @ SPCWAR and PEEK ^Y works fine, but :GAMES;SPCWAR doesn't.  It just blinks some graphics and then ITS freezes solid.

I put breakpoints on DRECYC+1, SRECYC, and LPTBRK+1.  DRECYC never hits.  SRECYC hits a few times, and LPTBRK indefinitely.  It does look much like if the CONO PI hack hack fails.  But I also ran my HACK HACK test and it passed.

larsbrinkhoff:
After much head scratching, I found an old sims commit with a working 340: abaeb33.

larsbrinkhoff:
The exact commit that breaks is 94709e1f.

If I revert the change to use pi_vect (two places), the 340 works again.

larsbrinkhoff:
Solved.  `pi_enc` can be updated, but `pi_vect` isn't.  This happens when the 340 hack hack is in use.  @rcornwell has a patch in the works.

larsbrinkhoff:
By the way, I think the "hack hack" should have worked on the PDP-6 as well.  It probably originated there.

larsbrinkhoff:
With @rcornwell's recent "pi_vect" fix, this should be good to go.  But now we might as well wait for the official release of KL10 into SIMH.

larsbrinkhoff:
Apparently there's a delay getting the simulator upstreamed, so I'm going ahead with this.  When this is merged, #937 can also be closed.

PR #1797, opened 2020-01-17
Lorenz Attractor

larsbrinkhoff:
If Edward Lorenz had visited the AI lab.

Lorenz


larsbrinkhoff:
This is just a draft; no need to review yet.

larsbrinkhoff:
Compare with this video frame from @livingcomputermuseum's 340 display.  (From https://www.youtube.com/watch?v=V4oRHv-Svwc)  It's contiuously drawing a circle.  The circle completes in 11 frames.  Assuming 30 frames per second, that would be 37 milliseconds.

spiro

PR #1727, opened 2020-01-26
Make 340 display optional

larsbrinkhoff:
When built with the SDL library, the current use of the KA10 simulator will always open a window 340 display.

I think it would be better for users to explicitly say if they want the 340.  For example, `./start 340`.

### DIFF
--- a/build/pdp10-ka/start
+++ b/build/pdp10-ka/start
@@ -38,7 +38,24 @@ type340() {
     cp build/pdp10-ka/run out/pdp10-ka/run
 }
 
+help() {
+    cat <<EOF
+This start script takes several command line arguments:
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
This change makes the Type 340 off by default.  Maybe it should be on by default?

If would be good if the 340 could be enabled and disabled at run time, but maybe that requires fixing some things in the emulator.

larsbrinkhoff:
Actually, `set dpy enabled` and `set dpy disabled` seems to work fine.  Tested with timesharing Spacewar.

larsbrinkhoff:
Also added a long overdue help text to the script.

PR #1809, opened 2020-01-29
Fix TVFIX TOP LEVEL INTERRUPT

larsbrinkhoff:
If CHANNA; RAKASH TVFIX (linked to TS STUFF) runs but tv11 isn't started, it will get a TOP LEVEL INTERRUPT and hang.  This is because STUFF is trying to write to the TV-11 memory, but not succeeding.  Thus trying to print an error message.  But there is no TTY channel open, which results in the interrupt.

This changes STUFF not to print anything in TVFIX mode.  It will just go away.


eswenson1:
What about having it write a message to SYSMSG?

larsbrinkhoff:
Good point.  Can it be done outside of the ITS monitor?

eswenson1:
Didn't PAPSAV and other daemons use the daemon message capability to write SYSMSG messages?  I could be wrong here, but I seem to recall there was a mechanism for daemon jobs to do so.

larsbrinkhoff:
As far as I can see it's just to retrieve messages from the monitor buffer and print them to TTY/file (SYSMSG/PAPSAV).

eswenson1:
Looks like you're right. I checked SYSENG; PFT > and SYSNET; COMSAT > and didn't see any code in these daemons that wrote to the system console (didn't check very well).  Maybe it isn't possible.  I assume you've looked at ITS and don't see any calls/uuos for this purpose?

larsbrinkhoff:
I haven't found any way to send a message to ITS for printing.

I did try to open T00 for writing.  If T00 is not logged in, an attempt to write results in the message "SYSTEM JOB USING THIS CONSOLE".  If it is logged in the .OPEN call returns a DEVICE NOT AVAILABLE error.

eswenson1:
Yes, Comsat uses this mechanism to write on the console, put the messages aren’t persistent in the log, and as you found out, it can fail if the console is attached to a job.

PR #1818, opened 2020-02-06
Added equivalents for Unix echo and shell comments to doc/DDT.md

jwbrase:
I've added DDT comments to the DDT cheat sheet as an equivalent to the Unix "echo" command and Unix shell comments. I've not submitted many pull requests in the past, please comment on whether I've done everything correctly procedurally.

### DIFF
--- a/doc/DDT.md
+++ b/doc/DDT.md
@@ -43,3 +43,5 @@ The ◊ character represents typing <kbd>ESC</kbd>, or holding down the
 | man 2 open    |                      | :call open               |
 | dmesg         |                      | :sysmsg                  |
 | uptime        |                      | :time                    |
+| echo "Hello World!" | | <kbd>^W</kbd>:◊<kbd>^V</kbd>Hello World! <kbd>^W</kbd>◊<kbd>^V</kbd> *(in an xfile)* |
+| \#Comment     |       | <kbd>^W</kbd>:◊Comment◊<kbd>^V</kbd> *(in an xfile)* |


larsbrinkhoff:
Thank you.  I agree the note was a bit on the long side.

I would suggest that a plain `:◊comment◊` without ^W and ^V is a more idiomatic comment in an xfile.

As pull request procedures go, can you please squash all commits together?  If you don't know git well enough, I can help you do it.  (And if so, maybe [this](https://github.com/larsbrinkhoff/fearless-git) can help.)

jwbrase:
I've merged the commits into one. I notice your link also talks about not adding commits directly to master. Should I close the pull request, figure out how to move the commit to a new branch, and resubmit, or is there a way to move to a new branch without resubmitting, or does it not matter at this point and I should just keep in mind for future reference not to commit to master?

larsbrinkhoff:
No, you don't need to close this pull request.  It's fine as is.

You are working on your own fork, so it's OK to commit to your own master branch.  But it would be messy if you were submitting more than one pull request at a time.  I think it's a good habit to always make a new branch for a new set of changes.

doc/DDT.md
```diff
@@ -43,3 +43,5 @@ The ◊ character represents typing <kbd>ESC</kbd>, or holding down the
 | man 2 open    |                      | :call open               |
 | dmesg         |                      | :sysmsg                  |
 | uptime        |                      | :time                    |
+| echo "Hello World!" | | <kbd>^W</kbd>:◊<kbd>^V</kbd>Hello World! <kbd>^W</kbd>◊<kbd>^V</kbd> *(in an xfile)* |
+| \#Comment     |       | <kbd>^W</kbd>:◊Comment◊<kbd>^V</kbd> *(in an xfile)* |
```
larsbrinkhoff: You have the comment inside a ^W...^V pair which makes the comment invisible in the output.  From reading xfile scripts, I think it's actually more idiomatic to leave the comments visible in the output.  On the other hand, hiding the comment is closer to the Unix behavior.

PR #1825, opened 2020-02-17
Build for FreeBSD on Cirrus CI

larsbrinkhoff:
I'm trying to use Cirrus CI to build for FreeBSD:

https://github.com/PDP-10/its/tree/lars/cirrus

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,16 @@
+env:
+  CIRRUS_CLONE_DEPTH: 1
+  ARCH: amd64
+  BASICS: yes
+
+task:
+  freebsd_instance:
+    image: freebsd-12-0-release-amd64
+  env:
+    matrix:
+      - EMULATOR: simh
+      # EMULATOR: klh10
+      - EMULATOR: pdp10-ka
+      # EMULATOR: pdp10-kl
+  install_script: sh -ex build/dependencies.sh install_freebsd
+  script: gmake all EMULATOR=$EMULATOR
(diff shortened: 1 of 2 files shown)


tvrusso:
To date I have seen no problem with the build mixing and matching gcc and clang.  Some Makefiles explicitly invoke gcc, some $(CC) (thereby getting clang instead).  As far as I can tell it all works fine once everything completes.

larsbrinkhoff:
@tvrusso, do you know if libusb is available for FreeBSD?  And if so, what is it called?

tvrusso:
I don't think you need a package for that.  According to the ports system:

Port:   devel/libusb
Moved:  
Date:   2013-04-30
Reason: Has expired: has been already included into base system

So it should just be there on any recent system since 2013.

larsbrinkhoff:
Getting close to a working build for "simh" (KS10) and "pdp10-ka" (KA10).

KLH10 has a minor problem with libusb which I'm going to fix later.

"pdp10-kl" does not do well.  Ping @rcornwell.

larsbrinkhoff:
KL10 failure looks like this:

```
spawn ./tools/sims/BIN/pdp10-kl build/pdp10-kl/boot
KL-10 simulator V4.0-0 Current        git commit id: 88b26df8
^[lits
 BINU   
The last command timed out.
gmake: *** [Makefile:101: out/pdp10-kl/rp04.1] Error 1
```

Expected is this:

```
spawn ./tools/sims/BIN/pdp10-kl build/pdp10-kl/boot
KL-10 simulator V4.0-0 Current        git commit id: 88b26df8
$L ITS BIN
$$L SALV BIN
```

This is an older version of the KL10 simulator.  I will need to retry with the newest version.

PR #1139, opened 2020-03-24
TVBROT - Mandelbrot set on TV display

larsbrinkhoff:
Or is the math too complex?


atsampson:
Looks good to me! Your inner loop's better than mine - I'd missed that computing z**2 without the multiplication trick saves a register...

PR #1855, opened 2020-04-02
Purify more programs

larsbrinkhoff:
These programs should be purified:
- FILE
- @
- QMAIL


larsbrinkhoff:
No issues.  I think they work fine without purification.  I searched for "purify:" at the beginning of a line.  There were some false positives that need to do self-purification.

PR #1858, opened 2020-04-06
More music from DECUS

larsbrinkhoff:
The .MUS files are the same format as for MUSCOM.  
http://pdp-10.trailing-edge.com/decuslib10-01/

Indeed MCOMP seems to be the same program as MUSCOM.


larsbrinkhoff:
Updating to TIME doesn't help.  I changed the constant TIMEK in H10D (player part of MUSRUN), and that did fix the tempo.  See #1859.

PR #1869, opened 2020-04-14
Update PDP-10 info file with KS10 information.

larsbrinkhoff:


doc/info/pdp-10.16
```diff
@@ -835,16 +835,16 @@ DIVB	C(AC) <- C(AC AC+1) / C(E); C(AC+1) <- remainder;
 
 PDP-10  Node: KL-Only, Previous: Fixed Point, Up: Top, Next: Floating Point
 
-This node describes some instructions that only KL-10's have.  This
-means that the only ITS machine which has them is MC.  You should
-probably not use them even if you expect to run on MC, so that your
-program can be moved.
+This node describes some instructions that only KL-10's and KS10's
```
eswenson1, line 838: This is incorrect English grammar.  The apostrophe signifies the possessive case.  It says that the KL-10s and KS10s have something.  Making something plural, as here, requires no apostrophe, but just the "s".  These should be KL-10s and KS10s.

atsampson:
It's probably worth noting that this INFO file is largely lifted from Ralph Gorin's "Introduction to DECsystem-10 Assembly Language Programming"; I assume the text for it must have been available on SAIL (although I don't think it's public on SAILDART). The [1985 version](http://bitsavers.org/pdf/stanford/lots/Introduction_To_DECSystem-10_Assembly_Lang_Pgmg_Jul85.pdf) has been updated to talk about the KL/KS differences.

larsbrinkhoff:
Thanks @atsampson.  I'll just make this minor update here, for now.

PR #1878, opened 2020-04-30
Imlac SYMFOR

larsbrinkhoff:
ARC: IMLAC; SYMFOR 10  
IMLAC; SYMFOR INFO

Character editor ported to PDS-4.

### DIFF
--- a/build/misc.tcl
+++ b/build/misc.tcl
@@ -1415,6 +1415,14 @@ respond "*" ":imtran\r"
 respond "@" "imlac; crash iml_imlac; crash bin\r"
 respond "@" "\021"
 
+# SYMFORM
+respond "*" ":midas imlac;_imsrc; symfor\r"
+expect ":KILL"
+respond "*" ":imtran\r"
+respond "@" "imlac; symfor iml_imlac; symfor bin\r"
+respond "@" "\021"
+expect ":KILL"
+
 # KLH's Knight TV clock.
 respond "*" ":midas klh; ts tinyw_klh; clock\r"
 respond "=" "1\r"
--- a/doc/programs.md
+++ b/doc/programs.md
@@ -312,6 +312,7 @@
 - SUPARD, draw ARDS graphics to SUPDUP.
 - SUPDUP, Supdup client.
 - SWAR, Imlac Spacewar.
+- SYMFORM, Imlac symbol editor.
 - SYSCHK, check up on system job.
 - SYSLOD, system load histogram.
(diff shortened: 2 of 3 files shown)


larsbrinkhoff:
Ah, the SYMFOR INFO file says what it does.  A character editor by @dlebling618.

```
Identification:

        NEW SYMFOR

        Dave Lebling

        February 13, 1974

Introduction:

     A new version of that old favorite, SYMFOR, has been written.
It contains most of the features of the old one, plus some exciting
new ones, and has the added advantage that (a) it runs on Graphics
IMLACs, and (b) it runs at all.

     WARNING!:  The new SYMFOR should only be used with SSV 30.
     It will kill SSV 29.  You have been warned!

The Display:

     When SYMFOR is loaded, you will see a display consisting of a
large (12*21) dot matrix, with a smaller (6*6) matrix superimposed on
it at the lower left.  This is the Sketchpad.

     Above and to the left of the Sketchpad is a letter B or a letter
D, indicating the state (Bright or Dim) of the "beam" in the Sketchpad.
This is initialized to D, or dim each time the Sketchpad is cleared.
To its right is a number which is continually updated to reflect the
number of Increment mode bytes in the character currently being drawn
on the Sketchpad.

     Below the Beam state are three lines indicating the size of the
characters being edited.

larsbrinkhoff:
SYMFOR overlays on top of SSV in the high part of memory.  But it clobbers locations 1 -4 to install its own interrupt handler, hence the warning.  SSV 30 was updated to restore the interrupt handler when entry points 100 or 101 are called.  SYMFOR will go back to SSV when ^Z or ^Q is typed.

larsbrinkhoff:
Probably same program as @tuban's file http://www.ubanproductions.com/ImlacPTP/40tp_symFormII_1.2C1_8kg.ptp

larsbrinkhoff:
SYMFOR has changes to run on a PDS-4.  Trying to port it back to PDS-1.  Some things are wrong, probably due to the different display resolution.

symfor

larsbrinkhoff:
The paper tape "symFormII" seems like an older version.  Not "NEW SYMFOR" described in the INFO file?  Because the command set and display don't match the description.  But the basics are the same.  I used it to draw a crude "A".

symfor-pt2

larsbrinkhoff:
Needs data switch 0 up, or else it will halt.

larsbrinkhoff:
Awaiting source code from DM tape.

PR #1879, opened 2020-04-30
Fixes for MAZE

larsbrinkhoff:
The LOADER subroutine breaks MAZE because it pushes all code higher, making constants and WALLS cross over 14000 making them inaccessible to the lower 2K.  It's also unnecessary since setting the start address to 101 jumps back to SSV which will accept the next data block sent from the PDP-10.

A possible explanation is that LOADER was added by someone running MAZE without SSV.  No other version of the MAZE source code has LOADER.

I removed LOADER and added a check to ensure the lower part fits in 2K.

### DIFF
--- a/doc/games.md
+++ b/doc/games.md
@@ -34,9 +34,15 @@ board on a Type 340 display.  Type
 <code>FANCY<kbd>TAB</kbd>2<kbd>Enter</kbd></code> to get the fanc
 chess board.
 
-### Maze War
+### Maze
 
-First multi-user first person shooter.  Type `:games;maze` to play.
+First multi-user first person shooter.  When logged in on an Imlac,
+type `:games;maze c` to play.  The `c` is necessary to avoid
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
The MAZE version with LOADER was sent by @mtravers to @novelty:

```
Date: Mon 14 Jul 86 18:20:36-EDT
From: Michael Travers <MT@XX.LCS.MIT.EDU>
Subject: mazewars
To: levitt@MEDIA-LAB.MIT.EDU
Message-Id: <12222708048.22.MT@XX.LCS.MIT.EDU>

TITLE MAZE.3 GREG THOMPSON (GAT) 04/11/74

;                                    MAZE

;       Maze  is  a  experiment  in  3  dimensional  graphics  and  intertask
;   teleconferencing.
```

Full text is sometimes found on @Pul-Ess's machine here:  
http://sv.svensson.org/COMMON%3BMAZE

Michael, I know it was a long time ago, but do you have any idea where you got that version?

Issue #1503, opened 2019-01-23
SST - Imlac monitor, terminal emulator, and graphics


larsbrinkhoff:

larsbrinkhoff:
Required some modifications to be accepted by MIDAS.  The IMTRAN'ed binary runs on sImlac.

larsbrinkhoff:
.INFO.; SST RECENT

```
SST.21                                                  (PDL 9/1/77)

        This version of SST incorporates a major change in the way that
Imlacs interact with ITS.  Specifically, SST.21 implements the "Software
TTY" codes, rather than the idiosyncratic Imlac codes we have been using
in the past.  This buys a considerable increase in speed and also a
saving in CPU overhead with Real-time TECOs.

**Comment by larsbrinkhoff** [comment]
.INFO.; SST INFO

```
SST INFO                                                PDL (3/9/77)

        SST is a new IMLAC terminal program which is designed to
interact with TECO, particularly ^R-mode TECO, more gracefully than does
SSV.  SSV was written during a period in which the normal mode of
editing interaction was to use EDIT mode and IMEDIT.  It was therefore
designed to maximize space for characters at the expense of time.

        Today, the normal mode of editing interaction with the system is
with any one of several ^R-mode TECO macro-packages.  SSV is poorly
adapted to the enormous amount of cursor motion required for such
interaction.  Consequently, SST has been implemented.

        SST is superficially the same as SSV:  It has many of the same
commands, and almost all of the features of SSV, including (limited)
graphics, Keyboard macros, local Muddle structure checking, and so on.
It does not have multiple character sizes, scrolling back of scrolled
off text, and page transmit.

larsbrinkhoff:
SST means "Super-Sonic version for Teco" according to SYS.00.00.

larsbrinkhoff:
New files from the Dynamic Modeling PDP-10: 

- IMSRC; STV21 3
- PDL; STV21 versions 14 to 18

They all claim to be SST 21.

larsbrinkhoff:
Also DM: IMLAC; STV21 IML

@jdersch: this file should be a honest-to-god MIT-modified-PDS-1 program.  And STV21 3 should be the corresponding source code since it has the same timestamp.

larsbrinkhoff:
STV21 screen coordinates:

```
        CSIZE=11.
        LSIZE=21.               ; characters are 8*12
        LEFT=<3*CSIZE>          ; leftmost character position
[...]

Issue #1398, opened 2018-11-15
MSC - Imlac multiple-screen program


larsbrinkhoff:
From *A Dynamic Debugging System for MDL", by @jmberez:

> The terminal most commonly used by users of the Dynamic Modeling System is the IMLAC PDS-1. This is a minicomputer capable of having programs loaded into it from the PDP-10 host computer. One program written for it by Dave Lebling is MSC, a multiple-screen terminal program. Up to four virtual screens (or pages) can be created that will individually operate like the actual screen area of the standard terminal program (SSV). Output may be directed toa ny one of the screens and any screen may be visible or invisible at any instant of time, at the programmer’s option. Selection of screen is controllable either by program from the host or locally by the user.

larsbrinkhoff:
SSV is issue #1001.

larsbrinkhoff:
The INIT program knows about an Imlac program called MSE, but not MSC.

larsbrinkhoff:
.INFO.; SSV RECENT calls this SSVMSE.

larsbrinkhoff:
MSB; SSVMSE 3

```
TITLE SSV 38

; Written for PDS-1D compatible IMLACs, uses DMS display addressing
; modification.  Includes ARDS graphics simulator.

; 38 (04/16/77) aren't Software TTY's wonderful?
; 37 (02/18/77) aren't TV's wonderful?
; 36 (10/26/76) macro keys
; 35 (05/06/76) Bugs in SSVSAI wrt ESC.
; 34 (02/09/76) CSIZE hacking.
; 33 (10/04/75) Really-truly overlays.
; 32 (07/24/75) Audible bell, movable right margin, case shifting.
; 31 (03/31/75) SSVMSE lives again! BUFCK bug fixed.
```

Issue #1001, opened 2018-06-20
SSV - Imlac monitor and text editor


larsbrinkhoff:
IMLAC; IMLAC SSV  
IMLAC; SSV4B IML
IMLAC; SSV4 74  
IMLAC; SSV4B 72  
IMLAC; SSV4LR 71  
IMLAC; SSVTTY 27  
IMSRC; SSV 52  
IMSRC; SSV IML  
.INFO.; SSV ORDER  
.INFO; SSV RECENT  
.INFO; SSV TTY

> SSV is the program sometimes resident in MIT-DMS IMLAC terminals. It provides communication with the DEC-1040, copious local storage for characters and graphics, flexible local editing functions, and ARDS-compatible graphics.

larsbrinkhoff:
This list includes SSV.22:  
https://github.com/livingcomputermuseum/sImlac/blob/master/imlac/notes/software.txt

(Also many other interesting programs.)

larsbrinkhoff:
There is also IMLAC; IMLAC SSV which is an SBLK file; probably output from MIDAS using IMSRC; IMDEFS.  The words are all 16-bit values, and there's a symbol table.

larsbrinkhoff:
Progress running SSV on @livingcomputermuseum's sImlac:
![image](https://user-images.githubusercontent.com/775050/46742638-c7e3cd00-cca7-11e8-8b92-7477c1221a71.png)

larsbrinkhoff:
We now have source code for SSV 52.

     TITLE SCROLL SAVER

It includes SSV CHARS, which looks like display lists for drawing characters.

larsbrinkhoff:
Also something called SST 21.

```
TITLE SST 21

; Written for PDS-1D compatible IMLACs, uses DMS display addressing
; modification.  Includes ARDS graphics simulator.

**Comment by larsbrinkhoff** [comment]
@livingcomputermuseum, see comments above for new Imlac software.

**Comment by larsbrinkhoff** [comment]
SSV 52 required adding closing brackets to literals.  (Setting .MLLIT to 0 or -1 doesn't work.)

The IMTRAN'ed binary runs on sImlac.

**Comment by larsbrinkhoff** [comment]
The .INFO.; IMLOAD RECENT for 12/5/73 says:

```
(b) The standard default program for graphics IMLACs is now
GSV IML, which has ARDS style graphics in it.  It is "SSV.9" in the
upper left of the screen.
```

[...]

PR #1875, opened 2020-05-03
Reconstruct SSV 22 source code

larsbrinkhoff:
Reconstruct SSV 22 source code from the IMTRAN'ed binary.

According to records (see #1001), SSV 22 is from 1974-04-05.  (Roughly contemporary with MAZE, #236, #1330.)  The closest relative seems to be the ML file IMLAC; SSV 52 dated 1973-01-11.  This file may have been edited to run on a PDS-4 (#1598), but the differences are minor.  Other files which can inform the reconstruction are SSVMSE (#1398), SST 21 (#1503), and STV 21.


larsbrinkhoff:
@jdersch, this is source code for SSV 22.

larsbrinkhoff:
The assembled file is an exact binary match against @tuban's SSV file.  MIDAS 324 is used rather than the latest version, because the latter will insert extra zeroes.

PR #1883, opened 2020-05-11
Add a clarification to the ITS build process

wyan:
The `.;ITS BIN` file should be created on `DSK0:` to prevent gotchas later on


larsbrinkhoff:
... when the CI builds go through.

PR #1885, opened 2020-05-13
Split off peripherals from CI build

larsbrinkhoff:
Shorten CI build times by not building all peripheral emulators.


larsbrinkhoff:
Messing with the build scripts revealed a need to also update the run scripts.

larsbrinkhoff:
There's a bug in the KL configuration.  The TTY device should only have 15 DTE lines, not 16.  This is because the CTY is the first DTE line.

PR #1893, opened 2020-05-15
Documentation update

wyan:


larsbrinkhoff:
Thanks!  Would you consider updating this paragraph: https://github.com/PDP-10/its#network-support and move some text to networking.md?

wyan:
@larsbrinkhoff here you go!

Issue #1877, opened 2020-04-30
Imlac CLOCK


larsbrinkhoff:
IMLAC; ICLOCK 5  

PDS-4 program.  Compare against @tuban's http://www.ubanproductions.com/Imlac/clock

larsbrinkhoff:
Binary does match the source code to a high degree.  The program accepts many keyboard commands:

- Keys F/B makes the clock run forward/backward.
- TAB toggles clock on/off.
- ESC (or ^Z in MIT version) jumps to 101, presumably back to SSV or other console/monitor.
- HOME resets clock to 00:00:00.
- R resets something?
- Left/right arrow set the direction for H/M/S.
- H/M/S increments or decrements hour/minute/second.
- Up/down arrows increments or decrements the counter?

PR #1895, opened 2020-06-05
Imlac clock program

larsbrinkhoff:
Add the Imlac clock program.  Displays an analog clock face with hour/minute/second.

See #1877  for source code and keyboard inputs.


larsbrinkhoff:
This version will clobber SSV.

larsbrinkhoff:
Hard from anywhere, I'd expect.  The IML files are just binary files coded as characters in the A to O range.

PR #1897, opened 2020-06-08
Alice's PDP-10

larsbrinkhoff:
Infamous text by @sraustein.  "Original" (i.e. oldest from backups) and latest version.

### DIFF
--- a/doc/humor/alices.pdp10
+++ b/doc/humor/alices.pdp10
@@ -1,0 +1,258 @@
+;;; With thanks (and apologies) to Chris Stacy, Alan Wechsler, Noel
+;;; Chiappa, Larry Allen, and of course Arlo Guthrie, and particularly
+;;; to Ann Marie Finn who is a kind soul and not at all like the
+;;; person portrayed herein.  --sra 3 May 85
+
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
@sraustein, ok to make a copy here?

doc/humor/alices.pdp10
```diff
@@ -171,16 +172,16 @@ hairiest TECO hacker of them all was coming over to me.  And he was
 mean and nasty and horrible and undocumented and all kinds of stuff.
 And he sat down next to me and said:
 
-[1:i*u14<q1&377.f"nir'q1/400.u1>[8
```
atsampson: Are the escape characters here right? (cf AI: DIGEX; MTD ALICE)

larsbrinkhoff: I took the "original" from AI: HUMOR;.  As far as I can see (but then, it's TECO) it matches the "latest" version except control characters have been updated to ^ and a printable character.  The DIGEX copy is different from both.

larsbrinkhoff:
Thanks!

PR #1898, opened 2020-06-08
Add HX ITS

larsbrinkhoff:
Add the HX machine name for [SDF.ORG public access ITS](https://hactrn.org/)


larsbrinkhoff:
For this to be more useful, I added the MCHN environment varable that can be used on the make command line to build a particular ITS.

build/h3text.2016
```diff
@@ -130,6 +131,8 @@ HOST : 10.3.0.54 : MC.MIT.EDU, MIT-MC, MC : PDP-10 : ITS : :
 ;;; DistriBution ITS and generic KA
 HOST : %CHAOS%%IP% : %HOSTNAME%, DB, KA, KL : PDP-10 : ITS : :
 
+HOST : 205.166.94.7 : HACTRN.ORG, HX : PDP-10 : ITS : :
```
larsbrinkhoff, line 134: @SMJ, hactrn.org IP address hardcoded here.

larsbrinkhoff:
About time we can build other ITS machines?  Well, there's some more to do to be able to have other disk pack configurations etc.

eswenson1:
Yes.  About time that we can specify the machine configuration.  And yes, now it would be nice to be able to configure disk configuration too.  If you recall, I had all this at one point and abandoned it (at your suggestion).  In addition to just having a "machine" (MCHN-like) setting that controlled the name, mine also controlled the inclusion and overriding of lots of other aspects of the build (so that you could have different disk configs, for example).  But we determined that bash/TCL were probably not the best ways of doing all the conditional inclusions, and that perhaps some better language was required.  I played with (yuk, barf) M4, and then thought it would be better to use python to "build" the build configuration.  You said you had already started down that path, so I abandoned mine.

PR #1926, opened 2020-08-24
Shorten tvbrot's iter loop

atsampson:
The code is now short enough to run from the registers, which would give a speed improvement on a real KA.

(Since we don't have a real KA, it doesn't actually do so yet, but it should be a little faster in emulation anyway...)

### DIFF
--- a/src/lars/tvbrot.209
+++ b/src/lars/tvbrot.209
@@ -123,19 +123,20 @@ xloop:
 	move i,y
 	movei t,maxiter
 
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Great!  Goes to show you can always squeeze out another instruction.  Almost.

I'm almost tempted to urge @rcornwell to make code run faster from the accumulators.  Shouldn't be too hard right?

Actually, it's slower on a KL10 so it could be inside a assembler conditional, or even more fun, a runtime check.

larsbrinkhoff (approved):

Issue #1916, opened 2020-08-13
Muddle interface to E&D display


larsbrinkhoff:
On ToTS tape 2100085:

MUDSYS; DISPLA MBD052, dated 1977-06-09.

File indicates it was written by @Brucekdan.

Issue #681, opened 2018-02-23
Older MIDAS


larsbrinkhoff:
Some old software, include MIDAS itself, won't assemble with the latest ITS MIDAS 458.

I'd like to have a set of old MIDAS versions all the way to 108 which is the oldest known version.

larsbrinkhoff:
- MIDAS 455 assembles with MIDAS 458 and TSRTNS 234.
- MIDAS 433 assembles with MIDAS 458 and TSRTNS 194.
- MIDAS 421 assembles with MIDAS 458 and TSRTNS 194.
- MIDAS 400 assembles with MIDAS 458 and TSRTNS 194.
- MIDAS 352 assembles with MIDAS 458 and TSRTNS 194.  There's a comment: Why not flush TSSYMS switch?

larsbrinkhoff:
MIDAS 323 doesn't assemble with MIDAS 458.

First, you have to set ITSSW==1.  Then, there's an `.ALSO RET]` which have to be broken into two lines.  `PTR` is expected to be predefined, but it itsn't.

larsbrinkhoff:
MIDAS 108 assembles with MIDAS 458.

Have to set TS==1, PTR==100, LDBI=ILDB, and DPBI=IDPB.

larsbrinkhoff:
Note: MIDAS 394 writes a debug info block (user, timestamp, source file); 353 doesn't.

larsbrinkhoff:
I think there may be a bug in MIDAS 108.  It doesn't seem to be processing the ^M from the command line.

If I force it to go to CMDE as ^M should, it does read a file and write an output file.  The output is an SBLK file, but quite as we know it today.  DDT won't read it, but ITS run it fine.

larsbrinkhoff:
Analysis of a MIDAS 108 binary:

1. Paper tape loader.
2. AOBJN pointer.
3. Data as per the pointer.
4. Checksum.
5. Another block with AOBJN pointer, data, and checksum.
6. JRST to start of program.
7. AOBJN block with symbol table.  Two words per symbol.  No checksum.
8. JRST to start of program.

larsbrinkhoff:
Our copy of MUDDLE; TS MIDAS is version 73.

larsbrinkhoff:
We have a .; @ MIDAS timestamped 1972-12-22.  I don't see much in the way of version numbers in it.

larsbrinkhoff:
Some more from BACKUP:

[...]

PR #1928, opened 2020-08-25
Build Muddle 56 with MIDAS 73

atsampson:
This does two things:

- Add a reconstructed MIDAS 73, and use it to build Muddle. We've had a non-MOBY binary of it for a while, but we need a MOBY version (bigger symbol table) to build Muddle with. Doing the reconstruction reveals why it was kept around - the relocatable output code was completely rewritten in later versions, so this was the last version to support some STINK features. 73 also has different IO code, and if you run it as MIDASD it has a different symbol table for E&S display code.

- Use different versions of the Tenex <mdl.int> sources to build Muddle. Previously we were using the latest version of each file (some of which were from 1983); we're now using the versions dated immediately before mdl106.exe was built in 1981, so they should correspond better to Muddle 106. They seem to have fewer dubious-looking hacks and no use of XMOVEI, at least.

The resulting Muddle binary now builds and initialises with the pure section starting correctly at 700000 (previously I'd hacked it to be 600000 since it ran out of space during initialisation). Zork still runs out of space if you try to load the whole game - but at least it now produces a sensible error rather than crashing.

If non-mangled versions of the <mdl.int> files become available in the future (https://github.com/PDP-10/muddle/issues/5), we could check that the Tenex binaries for MIDAS and Muddle match the sources used here, and add the Tenex code to MIDAS 73. Similarly, our Muddle 56 is still missing DISPL, the display interface that was present in Muddle 54 (#1916); that file could be added in if it becomes available.

(Related to #681 Early MIDAS, #826 MIDAS bugfixes, #830 Muddle.)


larsbrinkhoff:
I have noticed this too: `;XXX CONSTANTS emits two extra 0s`  It makes it harder to make bit-for-bit identical binaries.

larsbrinkhoff:
I believe the DISPLA code was removed because the E&E display was in disrepair by ~1977.  But chances are it will make a return!

src/mudsys/main.351
```diff
@@ -2023,7 +2023,7 @@ GCPDL:	-GCPLNT,,GCPDL
 
 PURE
 
-MUDSTR:	ASCII /MUDDLE /
+MUDSTR:	ASCII /MUDDLE ��
```
larsbrinkhoff: What is this character?

larsbrinkhoff: Oh I understand now, actually it's `/MUDDLE ^?^?^?/`.

larsbrinkhoff (approved):
Excellent work!

larsbrinkhoff:
The 1972 TS MUDDLE prints MUDDLE ^?^?^?^?^? IN OPERATION. Perhaps it's only half baked?

src/mudsys/main.351
ZoBoRf: In the GITHUB repository we will find in https://github.com/PDP-10/its/blob/master/src/mudsys/main.353 in line 2030:
MUDSTR:	ASCII /MUDDLE ¯
STRNG:	-1
	-1
	-1
	ASCIZ / IN OPERATION./
But ITS-TAR makes this in ITS:
MUDSTR:	ASCII /MUDDLE ^?^?^?/ 
(where ^? = 177/DEL)
That was something I stumbled over while analysing the muddle build's patching of the muddle version number.

ZoBoRf:
```
mkdir "mudsav"
respond "*" ":stinkm\r"
respond "STINK." "MMUD56 STINK\033@\033\033"
expect "SETPUR"
respond "\n" "D\033\033"
respond "\n" "strng/\0331\"56^?^?^?\033\r"
respond "\n" ":pdump mudsav; ts mud56\r"
respond "*" ":start\r"
respond "..PERM/ -1" ":pdump mudsav; ts mdl56\r"
respond "*" ":kill\r"
```

ZoBoRf:
> 
> 
> The 1971 TS MUDDLE prints MUDDLE ^?^?^?^?^? IN OPERATION. Perhaps it's only half baked?

Perhaps they patched the version number in after build, like yout build script
(https://github.com/PDP-10/its/pull/1928#discussion_r477122072).

larsbrinkhoff:
I don't see a call to SETUP anywhere in the code, or the build script.  According to a 1972 script, SETUP should be run first, then START.

larsbrinkhoff:
Never mind, the start address is SETUP.

larsbrinkhoff:
The build script no longer works.

```
respond "*" ":start\r"
respond "..PERM/   -1" ":pdump mudsav; ts mdl56\r"
```

doesn't match

```
*:start
MUDDLE INITIALIZATION.
*
..PERM/   BUFRIN   
```

Perhaps rather the somewhat random string ..PERM from DDT internals, check for `MUDDLE INITIALIZATION` and `*`?

atsampson:
That's weird - it was OK in the log of my test build last night:

```
D$$
strng/'MAIN$:   -1   $1"56^?^?^?$
:pdump mudsav; ts mud56
*:start
MUDDLE INITIALIZATION.
*
..PERM/   -1   :pdump mudsav; ts mdl56
*:kill
```

`BUFRIN` is indeed -1, but it's half-killed, so DDT shouldn't be using it for typeout.

It looks for `..PERM/` to check that the somewhat fragile Muddle initialisation process has completed correctly - this is the point at which it's likely to crash or print an error if something isn't right.

PR #1929, opened 2020-08-27
@ MARK

larsbrinkhoff:
Probably a program for marking a microtape. No source.


larsbrinkhoff:
Verified source code against original binary.  They are identical with two exceptions:
- The cruft from 20-37 is not reproduced.
- The original has one more word after the literals.  Maybe left over from patches?

larsbrinkhoff:
Sounds like something for MARK II, sorry MARK 2.

Joking aside, I don't know what this does!  I just work here.

PR #1932, opened 2020-08-28
UTNAM

larsbrinkhoff:
As far as I can see, the only way to set the name of a DECtape is from TECO no later than version 508.  No other program call the .UTNAM UUO.

I propose a new program called UTNAM to do this.  Take the name, three characters, from the JCL.  How to specify the tape unit?


larsbrinkhoff:
This first version gets the tape unit from the DDT :PRINT defaults.

atsampson (approved):
Looks good to me. I guess they might also have just called the UUO directly from DDT, since it's so simple...

Issue #850, opened 2018-04-27
DM security features


larsbrinkhoff:
@Devon7 reminded me DM ITS had special security features the other ITSes didn't have.  He suggested we should search for those in the ITS source code.

We know from Zork history the source code was in a hidden directory called CFS.

I looked in the latest version of ITS, but I didn't find any trace of any security features.

larsbrinkhoff:
What I did:
- Looked in SYSTEM; CONFIG > for special switches in the DM section.
- Looked at all MCOND DM locations in the monitor sources.

Also checked DUMP.  It does have some special code for DM, but nothing that seemed related to security or hiding directories.

jh95468:
ITS was the ultimate "open system" where anybody could do anything.  But you'd have to face your peers...and that was the basic security mechanism in the small community.

At one point, ARPA demanded that the system had to have passwords in order to log in.  So, one of the staff (Mike Brescia) got assigned to make it happen.  Not long after, everyone got a password and you needed to type it in on logging in.   I think that occurred on a Friday.

On the following Monday morning, an anonymous MOTD informed everyone that passwords were now required.  But.... if you forgot your password, you could simply use a new system utility - that let you type something like "psw <username>" to retrieve that user password.  So you could easily retrieve your password....or that of any other user.   I've always wondered who wrote that utility - a classic "weekend hack".

The only "real" security I recall was security through obscurity, e.g., the magic incantation to write enable DDT for SYS debugging, or the "Maze Guncher" that looked for Maze jobs and made them commit suicide without drawing suspicion to the Guncher's existence.

Another obsecurity mechanism was an "Al Imitator".  Al Vezza was our group manager, i.e., the boss.  Someone created a utility that allowed you to connect to someone else's terminal and impersonate Al, using the existing ^_C mechanism.  So on your terminal suddenly "MSG from AV:" would pop up, and your terminal I/O was now linked (allegedly) to Al's.  The hack utility recreated Al's characteristic hunt-and-peck typing behavior, so what came out on the victim's terminal really looked like it was Al typing.  Very useful for getting people to give up disk space, or core memory, or stop running some compute-bound job, etc.  This didn't work much for the group staff members of course, but for the stream of students that came through every semester it was very effective.

There were of course mechanisms in the system code to do things like memory protection so your erranbt code didn't write into someone else's memory space.  But I don't think of those as "security" mechanisms.

/Jack

36bit:
[...]

Issue #832, opened 2018-04-24
Compile 1981 Zork sources


larsbrinkhoff:
Use the MDL interpreter and Zork source code to make a runnable Zork.

larsbrinkhoff:
Random notes:

- I believe there should be a SYS*n*; TS ZORK.
- Maybe the source code should be moved to CFH.
- The default location for the Zork save file in ITS is CFH; MADADV SAVE.
- Zork seems to prefer MDL 104 (and 54?); there could be incompatible differences in 106/56.
- The TOPS-20 build script starts with loading M104UNI.  What's that, maybe a pre-built "universe" populated with libraries?
- Next, it loads LITLPK.GBIN.  What is that?

larsbrinkhoff:
I think we could use some Muddle eyeballs on build.cmd and rooms.394.  @36bit?

larsbrinkhoff:
- ITS-GET-NAME retrieves information from the INQUIR database (or DM equivalent?).  We probably don't have that library.
- Funny symbols: REMARKABLY-DISGUSTING-CODE, EXCRUCIATINGLY-UNTASTEFUL-CODE, etc.

larsbrinkhoff:
- SYS1; TS ZORK has timestamps from 1979-08-07 to 1979-11-24.
- SYS2; TS ZORK has timestamps from 1979-11-24 to 1981-11-27.
- SYS3; TS ZORK has timestamps from 1979-11-24 to 1987-11-13.
- KMP; AR9: ZORK 19 has timestamp 1979-11-24.  "The only MIT machine supporting Zork is MIT-DMS"
- KMP; AR9: ZORK 20 has timestamp 1981-11-27.  "No MIT machine supports Zork any longer."
- KMP; AR9: ZORK 21 has timestamp 1987-11-13.  "No MIT supports DM any longer."

larsbrinkhoff:
@eswenson1, was there ever a playable Zork on MIT-AI or MC?

eswenson1:
I don’t ever recall zork’s being available on MC, AI, or ML — only DM. Can’t rationalize those zork save files.

atsampson:
The SRI-NIC [games directory](https://github.com/PDP-10/sri-nic/tree/master/files/src/games) contains ZORK.EXE, ZORK.DOC, ZORK.LOG and MDL105.EXE. ZORK.EXE contains the strings:

```
SRC:<GAMES>MDLXXX.EXE
[...]

PR #1933, opened 2020-08-29
Move Zork to LCF;

atsampson:
We've previously discussed in #832 whether `LCF` or `CFS` was the right directory for the source; people remember both directories being associated with Zork.

I'm now pretty convinced the source should be in `LCF`, because @larsbrinkhoff found the patch to DM's `@ ITS` (see #850 and #1925) - the hidden directory is `LCF`. Lars also noted: "I imagine the fact that we have no later zork files on the DM tapes indicates the patch protected the files from being backed up."

In the latest Zork source, `CFS` is where `SAVE-IT` puts the `MADADV SAVE` Muddle image (and `build.cmd` shows that the save file went into a separate directory from the source on TOPS-20).


eswenson1:
Look good. iPhone github app won't let me approve. Consider this an approval!

larsbrinkhoff (approved):

PR #1925, opened 2020-08-29
Optionally hide Zork source code

larsbrinkhoff:
Add binary patch as described in #850.

Move Zork sources to LCF or CFS.
EDIT: They moved to LCF, see #1933.


larsbrinkhoff:
@atsampson, @eswenson1 I think for now the patch can just sit idle.  The file name is just my suggestion.

atsampson (approved):

Issue #1005, opened 2018-06-21
MIDAS info duplicated


larsbrinkhoff:
INFO; MIDAS 25 is duplicated by MIDAS; MIDAS MACRO.  The INFO file seems most up to date.

larsbrinkhoff:
Fixed by #1936.

PR #1936, opened 2020-08-30
Remove MIDAS; MIDAS MACRO

atsampson:
This is a slightly older copy of `INFO; MIDAS >`. The differences in the newer version all look like improvements.


larsbrinkhoff:
Thanks for checking!  I noted the similarity in #1005, but I never looked carefully to see which one is the most recent.

larsbrinkhoff (approved):

PR #1938, opened 2020-08-30
Replace MUDSYS; STINK 2 with SYSEN2; STINK 121T

atsampson:
`MUDSYS; STINK 2`, which is used to build Muddle, originally came from the `<mdl.int>` files, and we didn't know which ITS version it corresponded to.

After some comparison, it appears to be `STINK 121T` with TENEX conditionals added. So this gives a cleaner derivation of it, and fixes up some corruption in the `<mdl.int>` copy. It also drops the bad-EXPUNG patch that's no longer needed.


atsampson:
We had two versions of STINK before - there's a regular version (STINK 201) and a Muddle-specific one (STINK "2"). This is replacing the Muddle one.

atsampson:
(Although that reminded me that I hadn't tested whether STINK 201 can handle Muddle now we're using an older MIDAS - unfortunately it still crashes, so we do still need both versions...)

eswenson1:
Oh I see, you are writing to the TS STINK in the MUDSYS directory. Missed that.  Thought you were replacing the one in SYS:.

larsbrinkhoff (approved):
Thanks!

PR #1941, opened 2020-09-03
Document some removed UUOs.

larsbrinkhoff:
Work in progress.  I found the .ORGI and .ORGO UUOs in AI memo 161.  I think it would be fitting to add them and other obsolete UUOs to the documentation.

I also found .MSPACE in ITS 674, and some UUOs have empty descriptions: .POTSET, .VSCAN, and .VSTST.  The robot arm UUOs can also be improved.

### DIFF
--- a/doc/sysdoc/uuos.113
+++ b/doc/sysdoc/uuos.113
@@ -164,17 +164,25 @@ are described in more detail in the file .INFO.;ITS .CALLS.
 	The symbolic call RFPNTR may be used to read the current
 	access pointer for a file.
 
-.ARMOFF
+.ARMOFF					turn AMF arm off
 
-This uuo was for hacking the AMF arm.  The arm no longer exists.
+Turn the ARM arm off.  The arm no longer exists.
 
-.ARMOVE
-
-This uuo was for hacking the AMF arm.  The arm no longer exists.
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I have left some things unspecified, but I think this is good enough.

eswenson1 (approved):
I think it would have been better to have the same annotation "This UUO no longer exists." At the start of each of these historical UUOs. Some of the ones you added have this, but others say something like "The arm no longer exists, begging the question of what happens (now) if you call the UUO.

larsbrinkhoff:
OK, I'll look into that.

larsbrinkhoff:
I checked, and all UUO that *don't* say "this UUO no longer exist" do in fact exist.  I.e. the AMF arm and vidissector UUOs are still there, just inside a conditional is currently off.  In contrast, the code for .MSPACE .ORGI, and .ORGO is no longer there in ITS.

eswenson1:
Great.  Thanks for checking. So when you added new (old) UUOs, you added "this UUO no longer exists" if it cannot be called, and if it can be called, but might not work properly (because of a non-existent device, for example), you also indicated this?  That's exactly what I was hoping for -- wasn't sure whether you had done that or not.  Thanks.

PR #1873, opened 2020-09-06
Muddle function for printing maze

larsbrinkhoff:
Create a file IMLAC; MAZE PRINT with a function PRINT-MAZE.

From MAZE source code:

```
;        Users may specify their own mazes if they are the first player in  a
;   maze by giving a file name after "maze to use: ".  Just a CR will default
;   to the standard maze.  User mazes must have a specific format if they are
;   to be able to work. They must begin with a LOC 10020 followed by the label
;   MAZE:  on  the first of  32.   octal words which  form a bit  map for the
;   maze.   The maze must  end with  LOC 17713, JMP @.+1, 101,  and an  END.
;   After assembling the maze must be imtraned by using the "IMTRAN" command.
;   A  muddle function exists for printing out formated source mazes.   It is
;   initiated by  floading "imlac;maze  print"  in  muddle and  then  issuing
;   <PRINT-MAZE  "input file spec" "output file spec">$ where the output file
;   spec defaults to the TTY.   An example of a formated source maze is given
;   below:

### DIFF
--- a/src/imlac/maze.print
+++ b/src/imlac/maze.print
@@ -1,0 +1,84 @@
+; " -*- muddle -*- "
+; "Print an imtraned maze."
+
+<SETG MAZE <IVECTOR 34>>
+<SETG WEST [!<IVECTOR 13 " "> "W" "E" "S" "T" !<IVECTOR 15 " ">]>
+<SETG EAST [!<IVECTOR 13 ""> "     E" "     A" "     S" "     T"
+	    !<IVECTOR 15 "">]>
+
+<DEFINE GET4 ("AUX" (C <ASCII <READCHR>>))
+  ;"Read 4 bits from an imtraned file."
+  ;"Ignore all characters outside @-O."
+  <COND (<AND <G=? .C 64> <L=? .C 79>> <- .C 64>)
+	(ELSE <GET4>)>>
+
+<DEFINE GET8 ()
+  <+ <* <GET4> 16> <GET4>>>
+
+<DEFINE GET16 ()
+  <+ <* <GET8> 256> <GET8>>>
+
+<DEFINE GET-BLKLDR ()
+  ;"Skip over the block loader."
+  <REPEAT ((N 0))
+    <GET16>
+    <SET N <+ .N 1>>
+    <COND (<=? .N 65> <RETURN>)>>>
+
+<DEFINE READ-MAZE ()
+  ;"Read an imtraned maze."
+  <GET-BLKLDR>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by larsbrinkhoff** [comment]
@atsampson, @ZoBoRf,
You are the top Muddle experts.  Any obvious improvements that can be made here?

**Comment by larsbrinkhoff** [comment]
In particular I'm not so happy about the repeated use of `<SET X <+ .X 1>>`, but I didn't find anything similar to a for loop.

File: src/imlac/maze.print [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+    <REPEAT ()
+      <PUT ,MAZE .I <GET16 .CH>>
+      <SET I <+ .I 1>>
+      <COND (<=? .I .N> <RETURN>)>>
+    ;"Get block checksum."
+    <GET16 .CH>>>
+
+<DEFINE WRITE-LINE (CH S)
+  <PRINC .S .CH>
+  <TERPRI .CH>>
```
ZoBoRf, line 46: You could make CH implicit, then you don't need to put in in each subordinate routine call. 
(cf. "The Mudde Programming Language" Section 24.1.1 Comment 4.)
<DEFINE WRITE-LINE (S OUTCHAN)
    #DECL ((S) ANY (OUTCHAN) <SPECIAL CHANNEL>)
    <PRINC .S>
    <TERPRI>>

ZoBoRf (approved):
Very elegant and straight forward solution.

ZoBoRf:
> In particular I'm not so happy about the repeated use of `<SET X <+ .X 1>>`, but I didn't find anything similar to a for loop.


9.7.1 Local Values versus ENVIRONMENTs
SET, LVAL, VALUE, BOUND?, ASSIGNED?, and UNASSIGN all take a final optional argument which has not previously been mentioned: an ENVIRONMENT (or other TYPEs, see section 20.8). If this argument is given, the SET or LVAL is done in the ENVIRONMENT specified. LVAL cannot be abbreviated by . (period) if it is given an explicit second argument.
This feature is just what is needed to cure the INC bug mentioned in chapter 5. A “correct” INC can be defined as follows:

```
<DEFINE INC ("BIND" OUTER ATM)
   <SET .ATM <+ 1 <LVAL .ATM .OUTER>> .OUTER>>
```

atsampson (approved):
Looks good to me. The Muddle manual writes numerical loops that way; I don't know if there's a better idiom.

PR #1949, opened 2020-09-13
Fix UBAASL

atsampson:
This is the function that allocates page slots in the KS10 Unibus adaptor - I was using it when I was experimenting with KS10 Ethernet a couple of years ago.

It wasn't updating `UBAIFS` owing to a typo (`LSH A,B` for `LSH A,(B)`), so it always returned the first slot number. It's not actually used anywhere in ITS 1650, so this didn't break anything.


eswenson1 (approved):

larsbrinkhoff (approved):
Seems obvious enough.

PR #1950, opened 2020-09-13
Enable the IPC device for all build machines

### LINKED ISSUE #1421: Enable "message slurper" (DM only?)
Set MSPP in ITS to get the IPC device, which Muddle uses.

If #1323 has been solved, only enable this when building DM ITS.

atsampson:
This is a DM feature, and quite a few of the existing Muddle images expect the IPC device to be available. (This doesn't necessarily mean that they actually use it, just that they were saved with an IPC channel open.)

I haven't enabled DM DEMON support - it looks like that's a more intrusive set of modifications.

Fixes #1421.


eswenson1 (approved):

larsbrinkhoff (approved):
I have used ITS with MSPP==1, and as far as I could see everything works fine.

larsbrinkhoff:
Do you have any reason to believe DEMON will be necessary for something in the Muddle ecosystem?  I understand some DM software relied on it being present, but I don't know what exactly or how important it would be.

atsampson:
DEMON enables some extra .CALLs: DEMSIG, RDDMST, STDMST. There are a few programs in the tree that use these: QMAIL (if on DM), NETRFC, COMBAT, FTPS (if on DM), INQUPD (doesn't care if it fails).

Issue #1940, opened 2020-09-01
Muddle executables and save files


atsampson:
Notes on the surviving Muddle interpreter executables and save files, for the purposes of reconstructing the Muddle environment and applications. (Even when we have the source code, it's not always obvious what the resulting artefact should look like!)

This covers both ITS and TOPS-20 Muddle, and ignores anything earlier than version 50. See #1911 for the full history.

A Muddle save file is a dump of the interpreter's memory space minus the interpreter code itself; it can contain both interpreted and compiled code. The interpreter executable can also include preloaded Muddle code. The memory layout in the executable is different between ITS and TOPS-20, and between different versions.

Compiled relocatable functions (RSUBRs) can be generated by the Muddle compiler or assembler (or by hand). RSUBRs need to refer to the interpreter's internal variables and functions, so they aren't inherently portable between Muddle versions or platforms, although there's a "fixups" mechanism that allows a list of patches to be applied when loading into a different version.

ITS files are usually in good shape. Unfortunately most of the TOPS-20 files we have are mangled, e.g. by newline conversion.

"Basics" = the PKG, EDIT, etc. packages described in [The MDL Programming Environment](https://github.com/PDP-10/muddle/blob/master/doc/MDL_Programming_Environment.pdf) in compiled form.

## Interpreters

| Source | Filename | Version | Contents | Notes
| --- | --- | --- | --- | ---
| XX | [\<mdl\>mdl105.exe.93](https://github.com/PDP-10/muddle/blob/master/%3Cmdl%3E/mdl105.exe.93) | 105 | ? | -Wascii, ?
| Panda | [\<games\>mdl105.exe](https://github.com/PDP-10/panda/blob/master/files/games/mdl105.exe) | 105 | ? | -Wascii, good
| This repo | MUDSYS; TS MDL56 | 56 | None | Built from the 106 source
| XX | [\<mdl\>mdl106.exe.52](https://github.com/PDP-10/muddle/blob/master/%3Cmdl%3E/mdl106.exe.52) | 106 | ? | -Wascii, ?
| ? | [\<mdl.int\>mdl106.exe](https://github.com/PDP-10/muddle/blob/master/%3Cmdl.int%3E/mdl106.exe.5).{1-5} | 106 | ? | -Wascii, corrupt
[...]

Issue #1811, opened 2020-01-30
Early Zork


larsbrinkhoff:
Files from December 1977.

LCF; ACT1 38
LCF; NNP NBIN
LCF; NP 93
LCF; NDEFS NBIN
LCF; NP 92
LCF; DEFS 63
LCF; MAKSTR NBIN
LCF; MAKSTR 7
LCF; NROOMS NBIN
LCF; ROOMS 99
LCF; DUNGZ 56
LCF; NACT2 NBIN
LCF; ACT2Z 27
LCF; ROOMS 98
LCF; ACT3 13
LCF; ACT1 37
MADMAN; ZORK LOL
MADMAN; MADADV HELP
MADMAN; MADADV SAVE
MADMAN; OMADAD SAVE

eswenson1:
The sources we have already are later versions than these -- for files that overlap in name.  However, I don't know what NP, NDEFS, and NROOMS are -- my guess is that they were superseded by other files, given that ACT1-3 and ROOMS are older than what we have. Consequently, I suspect that the SAVE files are for an older version.  Still, it would be interesting to try to run them.....

larsbrinkhoff:
I made this as a note for the future.  I found these files some time ago, but forgot where.  I agree the latest version is better for this repository.

taa01776:
The NBIN files are not super-interesting, since they're compiled.
This leaves: NP, which I'd guess is a parser in development; and DUNGZ and ACT2Z, which are interesting because the source files were properly ACT2 and DUNG.
So I'll hypothesize that this shows some work on the parser to better support whatever puzzles were in ACT2.

eswenson1:
Are the DUNGZ and ACT2Z files plain-text or scrambled/encrypted? I ask because the SCRMBL program on ITS takes files with FN1 FOO and produces files with FN1 FOOZ.

taa01776:
Good point. Yeah, that’s plausible; we did keep them encrypted (which led to the whole adventure with Ted Hess’s using DEC 20s running acceptance to find the password for the source files he’d grabbed).
    -ta

----------------------------------------
Ceterum autem censeo Microsoftem esse delendam.

eswenson1:
And the reason I know this is when I stole all the sources off DM and copied them to MIT-MULTICS (when you caught me doing this), I also saved a copy on MIT-MC -- but I used SCRMBL to encrypt them all so as not to leave them lying in the open.

[...]

PR #1951, opened 2020-09-14
Add Muddle 54 and Zork

### LINKED ISSUE #1948: MUDSAV; SAV FILE and MUDSAV; FIXUP FILE
If you load a random Muddle image, and the interpreter exits with

```
:$ FATAL ERROR MAJOR FILE NOT FOUND $
```

... then it's failing to open `MUDSAV; SAV FILE` (or `MUDSAV; FIXUP FILE`). Your image is trying to call a compiled Muddle function from the system's library of pure functions - for example, Zork uses some functions from the `LSRTNS` library to get the user's real name.

There's a rather vague description of `SAV FILE` - the "special large file" - on p165 of the [Muddle 54 primer](https://raw.githubusercontent.com/PDP-10/muddle/master/doc/MDL_Primer_and_Manual.pdf). The code to search the library is in `MUDSYS; MAPPUR`. The code to initialise it and put functions into it is, inevitably, nowhere to be found...

An `RSUBR` object in Muddle can contain code in various forms. If it's a `PCODE`, that's a pointer into this library. The second word of a `PCODE` object is `offset-in-PURVEC,,offset-in-chunk`. When a pure `RSUBR` is called using Muddle's `MCALL` UUO, the UUO handler looks the code up in `PURVEC` and calls `PLOAD` to load it if it's not already mapped into memory.

`PURVEC` is a vector of library code chunks that this program might use, each being described by four words. `offset-in-PURVEC` above is an offset into this vector (so it's always a multiple of 4).
- Word 0: the SIXBIT name of the chunk.
- Word 1: where it's mapped in the current process, or 0 if it isn't.
- Word 2: (something to do with the LRU expiry process for getting rid of unused mappings)
- Word 3: where it can be found in `SAV FILE` - ``length-in-pages,,first-page`` - or 0 if we haven't looked yet.

To load a chunk from `SAV FILE`, you do the following:
- Take the last (non-zero) character in the chunk's name, divide it by the first word of `SAV FILE`, take the remainder and add one.
- That gives you a word number in the first page of `SAV FILE`. Read that word and it gives you a "directory" page number.
- Read the directory page. The first word of that page is the length of the directory in words. The rest is a sorted vector of pairs of words, where the first word is the chunk name you're looking for, and the second contains:
  - top bit - unused?
  - 7 bits - length of the chunk in pages
  - 10 bits - Muddle version number
  - right half - number of the first page in `SAV FILE`
- Binary-search the directory to find the chunk you want, map the pages into memory, and update the `PURVEC` entry.

If a chunk with a matching name and version number isn't found in `SAV FILE`, Muddle will then look for `MUDTMP; (name) SAV(version)` (e.g. `MUDTMP; LSRTNS SAV54`). If *that* isn't found, it'll look at `FIXUP FILE`, which contains patches for pure code that was originally compiled for a different version of Muddle. It has the same structure as `SAV FILE`, but the data is fixup information rather than code. So the final fallback is to load a chunk for a different version from `SAV FILE`, and apply a patch from `FIXUP FILE`.

In order to be able to run Zork images without patching them, we would need to generate a `SAV FILE` that contains a stub version of `LSRTNS` (that just returns `FALSE` from `LSR-ENTRY`), and (I think) a valid but empty `FIXUP FILE`. This shouldn't be particularly difficult...

atsampson:
The overall aim here is to make it possible to run the 1977/8 Zork save files that MIT recently released (#1811) on ITS. These commits:

- Add reconstructed source for the Muddle 54 interpreter, based on the various source versions we've got, and the executables that @larsbrinkhoff and @rmaldersoniii found (#1940). in particular we didn't previously have the GC file for ITS; we do now.
- Fix some problems in Muddle 56 revealed by the above.
- Add a program to generate MUDSAV; SAV FILE and FIXUP FILE, with stubs for the functions Zork uses.
- Add the Zork documentation (from Panda), the latest ITS Zork save file we have, and TAA's Zork launcher.

Fixes #1948. Still doesn't achieve #832 but some of this will help!


eswenson1 (approved):

larsbrinkhoff:
Hats off to @atsampson!  I'm so glad Zork is running now.

MKSVFL is written in the DM style, nice touch.

doc/madman/madadv.help
```diff
@@ -0,0 +1,331 @@
+ZORK User Information
+
+I. Getting a ZORK
+ 
+    a. ITS:  Say ZORK^K to DDT rather than :ZORK to get a zork.
```
larsbrinkhoff, line 5: Any clue what the difference between ^K and a colon command would be?  When I looked in DDT, I didn't see a difference.

larsbrinkhoff:
Commit message "Both of these are from MIT's zork-1978-12 release", should be "zork-1978-01".

larsbrinkhoff (approved):

larsbrinkhoff:
I fixed the "zork-1978-12" commit message.

larsbrinkhoff:
I'm monitoring the CI builds.

larsbrinkhoff:
@atsampson, would you like to write something to its-hackers?

doc/madman/madadv.help
atsampson: If you already have a ZORK job, ZORK^K will get rid of it before starting a new one. I guess they were worried about people accidentally running multiple copies.

atsampson:
Done.

PR #1958, opened 2020-09-18
Update Zork documentation with information about _MSGS_.

larsbrinkhoff:
The Zork save function checks that the user has a directory, and that there is a \_MSGS_ file in it.

### DIFF
--- a/doc/games.md
+++ b/doc/games.md
@@ -91,5 +91,17 @@ Enter a cave and hunt a creature called a wumpus.  To play this, type
 
 ### Zork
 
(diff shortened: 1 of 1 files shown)


eswenson1 (approved):

larsbrinkhoff:
There's the matter of whether to also revert back to the original TS ZORK or not.  This would require the \_MSGS_ file to play at all.  Discuss!

larsbrinkhoff:
No strong opinions were voiced.

atsampson (approved):

PR #275, opened 2020-09-20
File timestamps

larsbrinkhoff:
It would be nice to preserve the timestamps for files we haven't touched.  We lose them because git ignore the file system timestamps.

If we include a DIR.LIST file in the directores, itstar should use it to set the timestamps in the DUMP tape images.


larsbrinkhoff:
Work in progress.

larsbrinkhoff:
Ran all files in src, doc, and bin through a script that compares file names against those from extracted tapes.  When found, the time stamp was added to a DIR.LIST file that itstar supports to add metadata to created tapes.

larsbrinkhoff:
@eswenson1, @atsampson,

I was hoping to use itstar's DIR.LIST feature to add timestamps to the files that go onto the fresh ITS file structure.  DIR.LIST has a list of files and metadata like timestamps and authors.  But it seems this is too cumbersome, because itstar will now include *only* those files that are in DIR.LIST and ignore all others in the same directory.  So every time we add or remove a file, we would have to manually edit DIR.LIST too.  Even if is's a new file that never existed on the historical machines.

Now I think maybe just have a plain text file with one file and timestamp per line.

Suggestions?

atsampson:
Sounds like a reasonable idea. I remember we had to adjust the arguments to touch to make it work across multiple platforms (not sure how portable -h is). Would it be easier to do in Tcl?

larsbrinkhoff:
I use `touch -t` in the script like we did in the Makefile.  MacOS, FreeBSD, and NetBSD seem to have -h.  OpenBSD doesn't, but I would say we currently can't support that platform.  Tcl could well be a good alternative.

ams:
OpenBSD has touch -t.

eswenson1:
Why couldn't you have modified itstar to only apply the metadata from DIR.LIST if it was present, but otherwise, for a file not in DIR.LIST, just add it?

larsbrinkhoff:
Yes, I thought about updating itstar.  But the DIR.LIST format is also
somewhat cumbersome to work with.  For example, itstar doesn't accept
line breaks so it's not a file that is easy to edit manually.

eswenson1:
Isn't DIR.LIST just generated by some lisp code? At least it seems to be an sexp -- a list of lists.  Should be easy to parse (using lisp) and generate.

larsbrinkhoff:
The DIR.LIST files will have to be maintained.  I don't want to add Lisp
to the list of required tools.

larsbrinkhoff:
I made a Tcl version and it looks quite alright, but I haven't found out how to set the mtime of a symlink rather than its target.  The script fails when encountering a dangling link.

```
proc readstamps {list} {
    set f [open $list "r"]
    set lines [read $f]
    close $f
    foreach line [split $lines "\n"] {
        global timestamp
        set data [split $line " "]
        set timestamp([lindex $data 0]) [lindex $data 1]
    }
}

proc timestamps {dir} {
    global timestamp
    cd $dir
    set files [glob */*]
    foreach f $files {
        if [info exists timestamp($f)] {
            file mtime "$f" [clock scan $timestamp($f) -format "%Y%m%d%H%M.%S"]
        }
    }
    cd ..
}

readstamps [lindex $argv 0]
timestamps bin
timestamps doc
timestamps src

larsbrinkhoff:
According to a [stackoverflow answer](https://stackoverflow.com/questions/64005540/how-do-i-set-the-mtime-of-a-symlink-in-tcl), TCL has no way to set symlink mtime, and a call out to `touch -h` was recommended.

larsbrinkhoff:
I'm trying to collect timestamps from the ToTS data.  I'm making a listing with file names and timestamps, and another one with SHA-1 checksums to match against file contents.  From this I will try to make the most complete set of timestamps for the files currently checked into this repository.

larsbrinkhoff:
I think there is around 80-90% coverage now.  Some notable gaps:
- C
- CLIB
- DECUS
- GAMES
- IMSRC
- LLOGO
- MUDS54
- MUDSYS
- SPCWAR
- SYSEN3

eswenson1 (approved):

atsampson (commented):
Looks good to me.

PR #1403, opened 2020-09-22
Muddle intro, hello world

larsbrinkhoff:
@36bit and @atsampson, is it possible to make a small hello world application with our current set of Muddle tools?

Can we make a save file, and run it with something like the TOPS-20 bootstrap?  
https://github.com/PDP-10/its-vault/blob/master/twenex/sys/unsupported/mudsub.mid.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ There are some short introductions for beginners:
 - [TECO survival guide](doc/TECO.md)
 - [DDT debugging newbie guide](doc/debugging.md)
 - [Hello MIDAS](doc/hello-midas.md)
+- [Introduction to Muddle](doc/muddle.md)
 - [DUMP and itstar](doc/DUMP-itstar.md)
 - [Games](doc/games.md)
 - [Printing](doc/printing.md)
--- a/doc/muddle.md
+++ b/doc/muddle.md
@@ -1,0 +1,76 @@
+# Introduction to Muddle
+
+### Background
+
+Muddle, later renamed MDL, was a programming language created by the
+Dynamic Modeling group (also known as the Programming Technology
+Division) at MIT.  The most famous application written in this
+language is the Zork interactive fiction game.  The influental CLU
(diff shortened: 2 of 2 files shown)


atsampson (approved):
Looks fine for now. Real applications would have been compiled before dumping but we don't currently have the compiler...

It'd be nice to have a link to the Primer (and other manuals) from the muddle repo in here, since that's a good introduction to the language...

larsbrinkhoff:
Good idea.  I added links to the four major manuals.  I'm not sure how they all differ from each other, all except "Environment" seem to cover roughly the same subjects.

eswenson1 (approved):
I'm not really sure this is the right perspective, however.  You don't need to create a SAVE file in order to write and use MDL programs.  You can just load muddle source into the interpreter and run it.  The purpose of the SAVE files is really to save out a largish program (consisting of lots of sources) complete with initialization and startup logic.  It is essentially used to save the state of the interpreter after it has been loaded with a bunch of code.  Normally, SAVE files are created with compiled code, too -- not source code.  Normally, you compile muddle source to NBIN files and load them into an interpreter and dump to a SAVE file.  You can also convert the NBIN files to FBIN files, which are shareable.  Then you load the FBIN files into the interpreter and SAVE the whole thing out.  

But if you want to run a simple program your like your HELLO program, you'd normally just either run it interpretered (call <HELLO> after you've loaded it).  Or you'd compile it to an NBIN and load that into a new interpreter and call it.  

One of the values of creating a SAVE file is that you can then write a shell program that opens the SAVE file, determines the MDL version, invokes the correct MDL runtime, loads the SAVE file in, and resumes it.  So your example would be a lot better if you included that part.

larsbrinkhoff:
Thanks @eswenson1.  You make some good points, but I don't want to overhaul the text now.  I'll merge this in, but we can certainly update it later.  Especially if there's a working compiler.

PR #1963, opened 2020-09-29
Increase LLOGO allocations

larsbrinkhoff:
We ported LLOGO to the latest Maclisp, and now it seems very prone to run out of memory.  This triggers a question "you have run out of *something* space. more?"  Increase allocations in LOADER to avoid this.

For example, if I load TVFONT FASL and then type `(usefont '(20fg window dsk henry))` I get this:
llogo-out-of-memory


larsbrinkhoff:
It seems BIBOP was introduced in mid-1973 as an experimental feature, but it was not enabled in the production Maclisp.  By late 1974, it was on by default.  This is roughly the same time period the latest Lisp Logo versions were developed, which would explain why the loader checks whether BIBOP is present or not.

Now that LLOGO has been fast-forwarded into the 1980s without regard for backwards compatibility, maybe the right update is to drop the BIBOP check altogether.

larsbrinkhoff:
There is still a problem with HOW-BIG in LOADER >.

It sets the value of GC-DAEMON to a function which is expected to be called with a list.  The list describes the memory spaces; each item is a list with a symbol naming the space, and four fixnums.  However, LOADER expects the list to be (NAME FIX1 . FIX2) which is an older format.  HOW-BIG needs to be updated from using CDDR to access "FIX2", to using NTH 2.

When this is done, there is one more place with a similar problem in CREATE.

atsampson (approved):
Seems reasonable to me.

eswenson1 (approved):

PR #900, opened 2021-01-05
GT40 Lisp display slave

larsbrinkhoff:
I intended to add the SYSENG; FDITS 43 file in #651, but I forgot.

Maybe it's for the best, because I don't know where to put the built binary.

The PDP-10 340 version is built by answering negative to the PDP6F question, and then negative to the GT40F question.  Toggling on GT40F will make the binary, but I'm not sure where to put it.  The display library tries to load the slave from SYS; ATSIGN 6SLAVE or SYS; ATSIGN 10SLAV, but there's no separate file for the GT40 slave.


larsbrinkhoff:
I turned this into a pull request because @Isysxp wants to use the GT40 display slave.

larsbrinkhoff:
@Isysxp reported a .VALUE trying to run the GT40 Lisp display slave.

It seems the cause is due to a I/O channel not being allocated before LOADing ATSIGN 10SLAV into the DSLAVE job.  It works for the 340 slave because a different call path ends up calling SIXOPN first - allocating an I/O channel in the process - before calling TENOPN.  The GT40 call path calls TENOPN directly.

I pushed a commit to fix this.

larsbrinkhoff:
I compared against a 1974 version of Maclisp.  The GT40 Lisp display support in FDITS seems to have been developed up until then.  At this point the display support was built into Maclisp rather than being a separate FASL file.  The OPNTEN code is:

```
OPNTEN: MOVE T,[6,,(SIXBIT \USR\)]
        .SUSET [.RUNAME,,TT]
        MOVE D,[SIXBIT \DSLAVE\]
IFN NEWIO, PUSHJ P,GTSIXC
        .OPEN SIXC,T
        .VALUE
IFN NEWIO, PUSHJ P,GTDSIC
        .OPEN DSIC,[SIXBIT \  &SYSATSIGN10SLAV\]
        JRST OP6C1
        .CALL LSIXC
        .VALUE
IFN NEWIO, SETZM DSIC+CHTAB
        .CLOSE DSIC,
        MOVE TT,[002400+SIXC,,<400000+<P6/PAGSIZ>_11>]
        .CBLK TT,               ;MAKE PAGE 0 OF SLAVE INTO PAGE OF 10
        .VALUE
        MOVEM F,XARG                    ;0 => 340 SLAVE, "TNM" => GT40 SLAVE
```

This code looks like it should work fine.  So presumably the code was modified for a newer version of Maclisp and moved out to a SLAVE 8 file in 1976.  Apparently some modifications didn't work.  Possibly the GT40 wasn't in use, or used for other things by then.

larsbrinkhoff:
AI memo 307A says to use `STARTDISPLAY 'GT40` to use LLOGO with the GT40.  But I tried this and didn't see anything sent to T34.

larsbrinkhoff:
@Isysxp found out that the `T34` argument to DISINI is lost along the way.  I fixed this.

Now there's a .VALUE in Maclisp's JOBINT subroutine.  I believe there's an interrupt from an inferior job, probably DSLAVE running the LD10 and FDITS code.

larsbrinkhoff:
Also had to change SIMH configuration to set DLO0 mode rather than DLO, or else the keyboard code for ^Z wouldn't be sent.

Finally Lisp Logo displays on the GT40!
llogoworking

larsbrinkhoff:
The 340 turtle no longer works with GT40 support built into the display slave.  This will need to be fixed.

Isysxp:
Dear Lars,

 Absolutely brilliant. Sorry I hadn't got back to you earlier but I had added so many traps to various code
sections that I now have to do a resync and rebuild with your updated code from github.
This is partly why I have not done a formal pull on the code as it involved a lot
of messing about on my part to learn how the whole OS thing works!
 ...(Time passes)
YES. It works. In this case, I loaded up VT07 and did not login.
Used another terminal to run llogo.
Running llogo via the GT40 itself does not seem to work and causes a trap.
I suspect that this is not how people used to use the system. ie VTxx next to GT4x.
Will send Linux build later today as I still have 1 minor problem to fix.

Regards, Ian.


________________________________

larsbrinkhoff:
One more bug fixed.  The display slave called DSTART before GTTY has been set.  Since GTTY is assembled to indicate T34, the 340 display wouldn't be grabbed.  Just removing the first call to DSTART works.  GTTY will be updated, and DSTART is called later.

eswenson1 (approved):

PR #1992, opened 2021-01-15
TT2500 microcode

larsbrinkhoff:
Add TT2500 microcode and assembler.  Assemble the code.


larsbrinkhoff:
CC @picobrian.  This pull request adds some files to the ITS repository, and runs the 2500 assembler.

eswenson1:
I seem to recall disabling backquote for some old liblsp or libdoc files I committed quite some time ago.

Look for a call to SETSYNTAX or something like that. I'm not a computer now, so can't check myself. But I did run into this with sources that used commas for different purposes than the current norm.

larsbrinkhoff:
Thanks, setsyntax did the trick.

eswenson1 (approved):

PR #2002, opened 2021-01-18
Add the SIMH Imlac emulator.

larsbrinkhoff:
Add the SIMH Imlac PDS-1 emulator to the build and start script.

New section in README.md to describe the terminal emulators.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,3 +37,6 @@
 [submodule "tools/simlac"]
 	path = tools/simlac
 	url = https://github.com/jdersch/sImlac
+[submodule "tools/sim-h"]
+	path = tools/sim-h
+	url = https://github.com/simh/simh
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Some important environment variables
-EMULATOR ?= simh
+EMULATOR ?= pdp10-ka
 
 # Sometimes you _really_ need to use a different `touch` or `rm`.
 TOUCH ?= touch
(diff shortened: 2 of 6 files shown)


larsbrinkhoff:
I updated the SIMH submodule to get the latest Imlac emulator.  It seems the KS10 is broken: the build hangs shortly after booting the newly assembled "NITS".

larsbrinkhoff:
See https://github.com/simh/simh/issues/999

larsbrinkhoff:
Since the latest SIMH no longer works for building ITS, I had to add a new submodule to access the Imlac emulator.  The old version of SIMH will still be used to build ITS.

larsbrinkhoff:
I see now that the built file IMLAC; SSV22 IML is wrong.  It's missing the BLKLDR header.  This is due to IMLAC BLKLDR being wrong.

This should be fixed before the pull request is ready.

larsbrinkhoff:
Never mind.

eswenson1 (commented):
This seems very confusing to me. We have submodules "simh" and "sim-h".  No one is going to "get" the distinction.

eswenson1 (approved):

larsbrinkhoff:
I know, it's no fun.  Do you have another suggestion?

I first tried to make the submodule named simh2, but there's a bug in the SIMH makefile which prevents that.  Not that it's much better than sim-h.

eswenson1:
I can't think of anything better.  So is the IMLAC simh repo just a version of the original SIMH repo? And it is just the case that the SIMH emulator (in either repo) won't build ITS any more -- so we're switching to pdp10-ka?

larsbrinkhoff:
The old SIMH repo which is still there builds ITS, so we still have support for SIMH.

The new (sim-h) repo is also SIMH but the latest commit.  It adds an Imlac emulator but the KS10 no longer works for building ITS.  It probably still works for running ITS.

The switch to pdp10-ka is only the default.  I didn't think it was useful to stay with SIMH if it's no longer fully supporting ITS.

PR #762, opened 2021-01-22
SURVEY - Check Arpanet hosts

larsbrinkhoff:
As per RFC308, 523 and 530, MIT-DMS had a program to survey the status of all Arpanet hosts.


larsbrinkhoff:
SURVEY and SURGIV access NETWRK; CSDATA NOW to read/write data.  It's difficult to test this due to the lack of an Arpanet.  Not to mention support for talking NCP between hosts.

At the very least we can ensure there is a NETWKR directory.

larsbrinkhoff:
@brfid, now that I have examined SURSND I think I see 110 of its data files backed up.  Timestamps range from 1977 to 1981.

larsbrinkhoff:
The HISTORY program mentioned above could be MPROG2; HOSTOR USG061

```
;<DEFINE HISTORY-COMMAND()
     <CALICO-COMMAND ,MC-HISTORY
        [,HOST-TABLE
         "of"
         ["name or connection # of host.  If you want the same host as the previous
call just type 0 - this is much faster since the program does not have to search
through the files.  The initial default is DMS (#70). Subsequent default is the
previous host that you called."]
         ["SYM" "FIX"]î
         []
         "at"
         ["the socket #.  At present the only socket #1 is surveyed, so data for
the other socket numbers will not be available.  Default is 1."]
         ["FIX"]
         []
         "# of surveys"
         ["for this many surveys.  I.e. The integer will tell the program the maximum
number of lines of survey to print.  Maximum is 96.  Default is the previous value."]

eswenson1 (approved):

PR #2003, opened 2021-01-22
Deploy built images using rsync.

larsbrinkhoff:
With this addition, disk images built by GitHub Actions will be transferred to hactrn.org.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ defaults:
 
 jobs:
   build:
+    environment: Deploy
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
Rsync transports data over SSH.  They private key is in an environment variable called SECRET.  The key can be updated, but not viewed here: https://github.com/PDP-10/its/settings/secrets/actions

CC @SMJ, the owner of hactrn.org.

larsbrinkhoff:
I don't know what all the messages about "deployed" means exactly.  Images should be uploaded when something is pushed to the master branch.  We'll see how it goes when the branch is merged.

eswenson1 (approved):

PR #749, opened 2021-01-23
HOSTAT - Arpanet host status

larsbrinkhoff:
SYS2; TS HOSTAT  
SYSEN2; HOSTAT 42  
SYSBIN; HOSTS1 >


larsbrinkhoff:
I think this can be added along with #762, even though it can't run (yet).

I noticed that HOSTAB wants a HOSTS2 table, so I added that too.  It's a binary file that will have to do until we can get sources.

eswenson1 (approved):

Issue #2012, opened 2021-01-30
OD runs into GC problems


heasm66:
If I inspect a file with OD it always stops at about the adress 4300 (octal) with a GC error:

`:od madman; madadv dec12`

    004354:  777551055347     -227,,55347    0L9S9M  _]I%KG  ^?vHZs
    004355:  466031342532  -311747,,342532   RZ0LAX  FP9<5:  MAKE-
    004356:  406072444636  -371706,,-333142  7 LVY$  @PZDF>  ACTIO
    004357:  470000000000  -310000,,0        UHAOZV  G       N....
    004360:  440015000000  -337763,,0         0C8ZV  D -     H.h..
    004361:  000010777777       10,,-1         ZXLE    (___  ..G^?^?
    004362:  000000000000        0,,0                        .....
    004363:  000000000000        0,,0                        .....
    004364:  777551055347
    ;BKPT GC-OVERFLOW

larsbrinkhoff:
It's an early attempt at Maclisp programming by me, and it may well be deficient.  I don't see any obvious consing going on.

If you want to dissect the SAVE files, maybe my PDP-10 disassembler may be more useful or robust.  It's here: https://github.com/larsbrinkhoff/pdp10-its-disassembler.  You would typically use it with the `-r` option to disassemble a file without regard to the file format.

heasm66:
I just wanted to extract the text to the "US News & Dungeon Report" from the Zork-files that crashes. For that task it worked by using PRINT.

Is the LISP code available somewhere. I could make a crack at it in MDL...

larsbrinkhoff:
Sure, it's here: https://github.com/PDP-10/its/blob/master/src/libdoc/od.3

heasm66:
I'm a bit of a newbie on LISP, but I experimented a little running OD interactive.

Default program stops with an gc-overflow error at about line 3500:

     003533:  002334777777     2334,,-1       26IJU%   3<___  .Mg^?^?
     003534:  000034000000       34,,0         1XQJV    <     ..`..
     003535:  773446025307    -4332,,25307    Z4QI$M
     ;BKPT GC-OVERFLOW

Rewriting a function, PRINT-ASCII, to use less variables using bitwise operations instead:

    (defun clean-ascii (ch)
      (if (< ch 32.) (setq ch "."))
      ch)

    (defun print-ascii (word)
      (setq word (quotient word 2))
      (let ((result nil))
        (push (clean-ascii (logand word 177)) result)
[...]

PR #2013, opened 2021-02-03
Fix of issue #2012

heasm66:
Exchanged "implode" for "maknam" to avoid choking the GC for larger files.

Filter out ASCII 127 (DEL) from output in ASCII-column.


larsbrinkhoff:
Thanks, looks good!

larsbrinkhoff (approved):

PR #1966, opened 2021-02-03
CROCK deteriorating

larsbrinkhoff:
The CROCK display sometimes loses characters when it has been running for a long time.  Unsure if it's due to CROCK or ITS?

Sample photo:


eswenson1 (approved):
Nice!

PR #2018, opened 2021-02-15
Exclude KLH10 from FreeBSD build on CirrusCI.

larsbrinkhoff:
There is a problem with GNU m4 which is needed to build KLH10, and I haven't found a good fix.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -10,7 +10,7 @@ task:
   env:
     matrix:
       - EMULATOR: simh
-      - EMULATOR: klh10
+      # EMULATOR: klh10
       - EMULATOR: pdp10-ka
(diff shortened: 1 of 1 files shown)


eswenson1 (approved):
I don't understand. What changed that caused this to break? KLH10 is the only reliable network-capable emulator. I sure hope we fix rather than just stop building.

larsbrinkhoff:
I don't know what changed.  One day it worked, the other day it didn't.  I assume something in the build environment.  Maybe some update to the FreeBSD image.

I made some reasonable attempts to find and fix the problem, but I don't want to expend too much time and energy on FreeBSD.  If someone cares deeply, they should step in and provide support.

PR #2010, opened 2021-02-21
Exclude processor-specific things from common build

larsbrinkhoff:
KA10 things should only be built on a KA10: TV11, XGP11, STUFF, SPCWAR...

KL10 specfic things: UCODE, KLDCP, KLFEDR, FEDUMP, KLDDT, KLRUG, 11BOOT, BOOT11, etc.

KS10-specific things: KSHACK, KSFEDR, NSALV, etc.

### DIFF
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -158,9 +158,6 @@ respond "*" ":midas device;jobdev loss_syseng;loss\r"
 expect ":KILL"
 respond "*" ":link device;jobdev los,device;jobdev loss\r"
 
-respond "*" ":midas kshack;ts mtboot_kshack;mtboot\r"
-expect ":KILL"
-
 respond "*" ":midas syshst;_syshst;hosts3\r"
 expect ":KILL"
 respond "*" ":link syshst;ts hosts3,syshst;hosts3 bin\r"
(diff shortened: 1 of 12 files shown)


eswenson1 (commented):
Why is XGP spoiler only on KA10. I thought MC had an XGP spoiler?

eswenson1 (approved):

larsbrinkhoff:
Do you mean XGPSPL?  It's the unspooler which sends data to the XGP-11 through the 10-11 interface.

eswenson1:
I meant SYS2:TS XGPSPL. 

But perhaps when we sent XGP requests on MC, we used the Chaosnet to actually get them spooled from AI.

So you're saying that MC (KL) couldn't have ever had an XGP device connected to it that needed a spooler?

larsbrinkhoff:
I think XQUEUE is the spooler that runs on any ITS machine including MC.  The binary is SYS2; TS XGP which writes files to the AI:.XGPR.; directory, using MLDEV if necessary.  From there XGPSPL picks up the file and sends it over 10-11 to the XGP PDP-11.

larsbrinkhoff:
MC (briefly?) had a Gould printer attached to one of its front end PDP-11s, but it was a plain line printer.

PR #2020, opened 2021-02-22
MMODEM - Xmodem file transfer.

larsbrinkhoff:
GZ; MMODEM 181  
.INFO.; MMODEM ORDER  
SYS3; TS MMODEM  

Xmodem(?) file transfer.


larsbrinkhoff:
Help text when started without JCL:

```
ITS companion to the CP/M 'MODEM' file transfer program.  Usage:
                      :MMODEM xx filename
where xx is one of:
S - send file.  MMODEM figures out whether it is an ascii or a COM file
SA - send ascii file
SL (or SAL) - send ascii file converting CRLFs to LFs
SR (or SAR) - send ascii file converting CRLFs to CRs
SC - send COM file
SB - send random binary file (4 8-bit bytes per PDP-10 word, left-justified)
SH - send HEX file. The file is 'comified' in core before being sent
SM - send MACN80 output file. The file is 'comified' in core before being sent

R - receive file.  MMODEM guesses format (ascii/COM) by file type, or asks you
RA - receive ascii file
RL (or RAL) - receive ascii file converting LFs to CRLFs
RR (or RAR) - receive ascii file converting CRs to CRLFs
RZ (or RAZ) - receive ascii file containing ^Z's. Normally ^Z is treated as end
        of file.  With this option, only trailing ^Zs are ignored.  May be
        combined with the L/R options above (e.g. RAZL)
RC - receive COM file
RB - receive random binary file
RxC - like Rx (where Rx is one of above) but using CRC's instead of checksums.

Note that the non-ascii files on the CPM; directory are *not* random binary
files, they are COM files.  Use SC/RC for those.

Mail bugs, questions, feature requests to BUG-MMODEM@MC.

eswenson1 (approved):

larsbrinkhoff:
Hello @johnsonjh,

You have CP/M running, right?  Would you know how to use (some version of) MODEM to dial in to a mainframe?

johnsonjh:
My primary CP/M is CP/M-68000, and unfortunately, wouldn't work for this, at least not in any historically accurate way. Also, my  simulator has just one serial line supported which is used for the console.

(I'll add playing with this on the TODO list, though.)

However, you *should* be able to get this going, trivially, under CP/M-80, using one of the various emulators out there.

Most of the serial communications programs, such as MODEM/MODEM7/YAM/etc., expect to drive the serial line directly, so CP/M-68K would be of little use even with another port simulated - COM8080 (Intel 8080 emulator for CP/M), at least the version I built for CP/M-68K, doesn't support serial lines - so a "real" CP/M-80 system would be needed.

Somewhat related, I have (what I believe) is the last version of [CP/M YAM](https://github.com/johnsonjh/YAM) in a repository, but I believe an earlier Christensen MODEM(7) would probably be what you'd be looking for, for accuracy.

PR #2022, opened 2021-03-18
Add support for color scope.

larsbrinkhoff:
Update the KA10 emulator to bring in the color scope device.  Add cscope to start script.


eswenson1 (approved):
I wasn't able to tell whether the au module was correctly updated. I assume it was.

PR #2023, opened 2021-04-03
Files for Muddle pretty printing.

larsbrinkhoff:
Courtesy of @Rick-Shiffman.

### DIFF
--- a/src/rrs/muddle.init
+++ b/src/rrs/muddle.init
@@ -1,0 +1,23 @@
+
+;"RRS;MUDDLE INIT ASKS IF YOU WANT TO LOAD PPRINT, Y OR y FOR YES
+  EVERY THING ELSE IOS NO. IT ALSO MAKES SURE THAT ,NULL IS SET
+  CORRECTLY"                                                   
+                                                                
+<DEFINE SETNUL ()                                               
+ <COND (<NOT <GASSIGNED? NULL!->>                                 
+        <SETG NULL!-                                             
+        <INSERT <ATOM <ASCII 127>> <GET ROOT OBLIST>>>)>> 
+        ;"<INSERT <ATOM <ASCII 127>> <GET INITIAL OBLIST>>>)>>"
+                                                                
+                                                                                                                           
+<DEFINE STARTER4 ()                                
+ <RESET .INCHAN>                                   
+ <TERPRI>                                          
+ <PRINC "PPRINT? (Y OR N)">                        
+ <COND (<MEMQ <TYI> "Yy "> <FLOAD "RRS;PPRINT >">)  
+       (T <SETNUL>)>>                              
+                                                                                                            
+                                                                   
+<STARTER4>                                                         
+;<FLOAD "RRS;PPRINT 2">
+
--- a/src/rrs/pprint.2
+++ b/src/rrs/pprint.2
@@ -1,0 +1,404 @@
+ "MUDDLE PRETTY-PRINT, FRAME-SCANNER, AND OTHER ROUTINES"
+<PRINC "/PPRINT/FRAMES">
+<BLOCK (<ROOT>)>
+
+"These atoms are placed in the ROOT oblist to allow general
+	access to their functions" 
+M
+;<COND ( <NOT <GASSIGNED? NULL!->> 
+	<SETG NULL <INTERN <ATOM <ASCII 127>> <GET INITIAL OBLIST>>>)>
+<COND ( <NOT <GASSIGNED? NULL!->> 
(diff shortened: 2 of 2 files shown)


src/rrs/muddle.init
```diff
@@ -0,0 +1,23 @@
+
+;"RRS;MUDDLE4 INIT ASKS IF YOU WANT TO LOAD PPRINT, Y OR y FOR YES
```
eswenson1, line 2: Typo here? MUDDLE INIT rather than MUDDLE4 INIT?

src/rrs/muddle.init
```diff
@@ -0,0 +1,23 @@
...
+        <SETG NULL!-                                             
+        <INSERT <ATOM <ASCII 127>> <GET ROOT OBLIST>>>)>> 
+        ;"<INSERT <ATOM <ASCII 127>> <GET INITIAL OBLIST>>>)>>"
+                                                                
+                                                                                                                           
+<DEFINE STARTER4 ()                                
+ <RESET .INCHAN>                                   
+ <TERPRI>                                          
+ <PRINC "PPRINT? (Y OR N)">                        
+ <COND (<MEMQ <TYI> "Yy"> <FLOAD "RRS;PPRINT 2">)
```
eswenson1, line 17: Should this really reference version 2, or rather ">"?

eswenson1 (approved):

src/rrs/muddle.init
larsbrinkhoff: Thanks!  I made the updates you suggested.

Rick-Shiffman:
Hi Lars, Eric, and everyone,
I have one small change to improve the tastfullnes in my Muddle Init file. Starter4 now
Excepts an ascii space as yes also. The COND was:
+<COND (<MEMQ <TYI> "Yy"> <FLOAD "RRS;PPRINT 2">) 
And is now:
+<COND (<MEMQ <TYI> "Yy "> <FLOAD "RRS;PPRINT 2">) 
String "Yy" has been changed to "Yy ", space is also a yes.

Lars, we should "PPRINT >"  pretty print's file name. At the moment "RRS;PPRINT 2" is my
Most stable version, unfortunately "RRS;PPRINT 3" not reliable yet and not for general use
Yet. I should rename "RRS;PPRINT 3" to "RRS;XPPRINT 3" so "PPRINT >" would work.
Also the FLOAD in STARTER4 should point to wherever PPRINT lives on your system.

Here is my current version of my MUDDLE INIT file:
;"RRS'S MUDDLE INIT"

<DEFINE SETNUL ()                                              
 <COND (<NOT <GASSIGNED? NULL!->>                              
        <SETG NULL!-                                           
        <INSERT <ATOM <ASCII 127>> <GET ROOT OBLIST>>>)>>      
        ;"<INSERT <ATOM <ASCII 127>> <GET INITIAL OBLIST>>>)>>"

                                                      
<DEFINE STARTER4 ()                                   
 <RESET .INCHAN>                                      
 <TERPRI>                                             
 <PRINC "PPRINT? (Y OR N) space is yes too">          
 <COND (<MEMQ <TYI> "Yy "> <FLOAD "RRS;PPRINT 2">)    
       (T <SETNUL>)>>                                 
 

<STARTER4>

eswenson1:
Is your unstable PPRINT 3 checked in (committed to the pod-10/its Git repo?). If not, then it doesn’t matter that the reference is to PPRINT >. When, and if you commit PPRINT 3 to the repo, of course, the changed INIT would use it. 

However, I understand if your local machine has a PPRINT 3 and you pull the updated sources from Git, you’d render your own environment unusable. So locally, you should update your INIT file to reference PPRINT 2.

I suggested the use of “>” because it is more generally how this would be done in the “old days”. And if you later committed a PPRINT 3 (or greater) the INIT would still work without changes.

But in retrospect, the change was to YOUR INIT file and not some system library file, so perhaps the suggested and implemented change was not justified. 

Also, there were cases in the old days, where explicit version numbers were specified — in cases where known breakages (incompatibilities in later versions) were intended to be prevented.

-- Eric

eswenson1:
Feel free to commit your updated INIT file with the explicit version if you wish (and your updates) and this time I won’t, in my review, object to the use of explicit version numbers (because it is simply your personal MDL init file.

-- Eric

larsbrinkhoff:
There is still time to update this pull request, because it hasn't been merged yet.

eswenson1:
Good point! :-)

larsbrinkhoff:
I made the updates Rick wanted.
1. Accept a space character in STARTER4.
2. Got the latest copy of PPRINT 2 from TT.

PR #2031, opened 2021-04-29
Fix broken links in CHANNA causing TV-11 and NAMDRG to fail.

larsbrinkhoff:
@techfury90 noted the TV-11 would no longer work after booting up ITS.  The cause is a mistake in #2010.  The links to RAKASH NAMDRG and TVFIX were wrong due to an added semicolon at the end of the :LINK command.  This would cause the automatic TV-11 stuffing to fail, and also not start the name dragon.


eswenson1 (approved):
Wow! How long has that been there? :-)

eswenson1:
Never mind. Saw your email with PR where it got broken. Fire your reviewer (assuming it was me who approved that PR)!

larsbrinkhoff:
Two months!

PR #2038, opened 2021-06-22
Use the CH11 Chaosnet address

bictorv:
Use the Chaosnet address reported by the CH11 device at boot time (e.g. as configured in klh10), rather than the statically defined one. 

This makes it easier to change a system's address (e.g. for joining the Global Chaosnet).  Rather than recompiling ITS, copying it to the host, converting the format, and dumping a new @.its (as described in distrib.its), you just need to change the klh10 init file (the myaddr parameter of the ch11 devdef). 

I have only tried this in klh10, but if the CH11 is properly implemented it should also work in other systems.

This patch only handles the CH11 device, but it seems the CH10 device also supports reading its address, so someone could fix that. 

(This also may open the possibility of using dynamically allocated Chaosnet addresses in the future.)

The H3TEXT file will still need changes, but it is easily editable, and there is DNS to the rescue...


larsbrinkhoff:
Bumped version numbers.

larsbrinkhoff (approved):

eswenson1 (approved):
Very nice!

PR #2042, opened 2021-06-22
Default to a new century in PDTIM

bictorv:
Boldly moving into the 21st century


eswenson1 (approved):
It's about time!  Thanks!

PR #2043, opened 2021-06-22
Avoid binary patch of literal by instead editing source

bictorv:
Come on?


eswenson1 (approved):

atsampson:
The version number should have been bumped to prufd.11 at some point (as a previous commit edited it from the original) - might be worth doing that here.

larsbrinkhoff:
I did the bump.

PR #2046, opened 2021-06-25
Avoid hardcoding ITS machine names

bictorv:
Instead, use ITSNMS table.


eswenson1 (commented):
Wow! This was a lot of work. Did you do this a while back, and have you been running with these changes for a while?

larsbrinkhoff:
Thanks!  Was this work powered by pickled herrings?

eswenson1:
That reminds me of the time my wife was hosting the Borland Scandinavian product managers at headquarters in the US (she was in product management there at the time). She thought she was doing something that they would appreciate by having an assortment of picked herring as the centerpiece of the snacks. She very quickly learned that many Scandinavians actually don't like pickled herring. She was disappointed. What's with you guys, pickled herring is great?

bictorv:
No I just did it after https://github.com/PDP-10/its/issues/441#issuecomment-866799912 - it's very much the same thing over and over, but with some interesting quirks.

The test build fails since the input to MAGFRM is "KA" which apparently is neither the local host or one of the ITSNMS. I don't really see the point of checking for known hosts except possibly to avoid spelling errors, but the original did the corresponding thing for a fixed list of machine names. Either fix the input to the test build, or we change the behaviour of the program?

Pickled herrings are just about to start. Lars, don't forget about the mandatory herring for easter! And in Uppsala, also on the last day of April! :-o
(And some snaps, of course...)

bictorv:
Oh dear, someone might want to update the version numbers of files.

eswenson1 (commented):
Yes, let's please always bump up the versions of changed files. It makes it so difficult when we can't tell by FN2 that a file is different -- especially when you are trying to maintain multiple systems and keep them up-to-date with latest DB builds.

larsbrinkhoff:
I can bump the versions.  I'd also like to fold the "typo" and "sigh" commits back in their proper place.

eswenson1:
Which were those? What about MOON's comment regarding "This sucks totally?". I guess BV fixed the suckage Dave was referring to, but we're losing the comment! But I didn't see "sigh" and "typo" comments removed.

larsbrinkhoff:
Here are two commits titled "typo" and "sigh."  Where is the Moon comment?
sigh

bictorv:
It was in DIRDEV. Perhaps-a it could be added to the general problem description? :-)

eswenson1:
I completely misread your comment Lars. I read "comment" rather than "commit". I thought you were saying that there were comments (with those words, and presumably more) that BV's changes removed (due to replaced code) that you wanted to fold back in (because they were historically interesting). And when I reviewed the PR, I saw one of Moon's comments was "removed" (because of BV's better code). But in fact, presumably Moon's comment was saying that the burned in table was yucky, and BV's fix hot rod of it, so the comment was justifiably removed. So never mind!

src/syseng/magfrm.19
```diff
@@ -86,8 +86,11 @@ WHMCHN:	TYP FOR WHICH MACHINE?
 		  .VALUE
 		 JRST .+1 ]
 	MOVSI A,-LMCHNTB	
+ifn 0,[	;; BV: The original program checks to see if the user typed a known machine,
+    	;; but I argue there is no point in doing that.
```
larsbrinkhoff, line 90: I think I may agree with you on that.  And if so, MCHNTB isn't needed either.

There is some value in checking that there is a corresponding MADGMP BIN file for the target ITS but the .OPEN will fail if there isn't.

larsbrinkhoff:
I squashed together those commits that changed the same files.  I wonder if perhaps it should be one big commit for everything, or one for each program?

I also added another update to MAGFRM.

Version numbers not bumped yet.  I'm waiting until the dust settles.

bictorv:
I agree on the MAGFRM change! And go for one big commit for everything, I'd say.

larsbrinkhoff:
Squashed together and bumped version numbers.  This should be ready now, barring any objections.

larsbrinkhoff:
The build now breaks in the Macsyma part: 
```
YESNOP 44^V SYS: SYS; PURQIO 2155 - FILE NOT FOUND
```

Is this due to the new QIO version?

larsbrinkhoff:
I updated the PURQIO version number and that fixed the build problem.

src/sysen1/pword.2664
```diff
@@ -9826,7 +9850,11 @@ bltspc:	syscal sstatu,[val x ? val x ? val x ? val x ? val x
 	       blt x,spcend-1	   ;  all of them to the bitter end
 	       setom dm.flg
 	       ret ]
-	.lose		           ; Unknown machine!
+	;; unknown machine, default to MD
```
eswenson1, line 9853: Why default to MD?

eswenson1 (approved):

src/sysen1/pword.2664
bictorv: As good as any? If you want specific settings, you will have to edit the code anyway, but if you don't, then MD is OK? The main goal is to be able to set a new machine name in CONFIG with a minimum of other edits (TTYTYP and H3TEXT seems to suffice).

larsbrinkhoff: Historical reasons.  DB used MD as the default (maybe Bawden's choice?), so I kept that for new machines.

PR #2048, opened 2021-06-26
Increase KA ITS number of directories to 500.

larsbrinkhoff:
Requires an update to the build script to accommodate a bootstrap ITS with a disk format different from the target ITS.

PR #2052, opened 2021-07-05
Add support for building more machine configurations

larsbrinkhoff:
With #2048 in place it's possible to start with any ITS and disk format and use that to bootstrap a new ITS with another disk format or controller.  Use this to expand the ability to say MCHN= on the make command line.  Individual machine configurations could sit in e.g. `build/mchn/$MCHN`.


larsbrinkhoff:
@eswenson1, this is only implemented for KA10 so far.  Do you want to build ES and/or EX like this?

eswenson1:
I don't think I'd use it for ES. That system is based in PI and I just overlay selected source changes and rebuild on ES.

EX is actually a DB, so I would like to be able to build that KLH10-based system from source. And it would be nice to have it claim it is EX when built. 

So, yes, for EX, I'd like to be able to build a KLH10-based system with the name EX, and possibly some customizations (disks).

bictorv:
Interesting, I'm working on something along the same lines (for build/klh10 for now), but probably won't have time to compare and look into until next week.

larsbrinkhoff:
@bictorv, I'm updating this pull request (after all, it's still a draft) to also include klh10 and simh (all KS10 emulators, really).

@eswenson1, I'm *not* editing files to add EX ITS here.  But if someone wants to add a new KS10 ITS, the road has been cleared.

eswenson1:
> @eswenson1, I'm _not_ editing files to add EX ITS here. But if someone wants to add a new KS10 ITS, the road has been cleared.

Great!  Thanks.  I'll give it all a whirl (providing I'm able to run the build (timeouts)), and if I can build an EX successfully, I'll update the repo with required changes.

bictorv:
Still haven't had time to look at it. My version essentially substitutes in CONFIG similar to how %IP%,  %NETMASK%, %CHAOSP% and %CHAOSA% are done, but also for %MCHN%, %TZONE% and %DSTEU. Additionally %MCHN% for TTYTYP. In both TTYTYP and CONFIG (in build/klh10), %MCHN% is where DB used to be, if you see what I mean? This makes it easy to create an ITS for a new machine (but only with default values for things not substituted, of course).

I also made an Expect hack to only copy CONFIG, TTYTYP and H3TEXT (from out/klh10/...) to an existing ITS, and another to recompile ITS and dump it to @.itsbin-u. 

Back online in about a week.

eswenson1:
Wanna share your expect hack?

bictorv:
> Wanna share your expect hack?

Of course, when I'm back online in a week!

eswenson1 (approved):

src/l/struct.658
```diff
@@ -715,7 +715,7 @@ PG$		MRA PAGING
 		MRA PPN
 20$ 		MRA PS
 IFN ITS,[
-		MRA [ITS,AI,ML,MC,MD,MX,DB,KA,KL,HX]
+		MRA [ITS,AI,ML,MC,MD,MX,DB,KA,KL,HX,TT]
```
larsbrinkhoff, line 718: @bictorv, did you consider this list of machine names for replacement with something else?

bictorv: That code just creates the atoms (MRA=Make Random Atom) which were needed by the old code (in macro expansion), but the new code creates the atom for the local host using SIXATM, so no need to create them beforehand.

bictorv: Actually the list could probably/better be removed in STRUCT (possibly retaining just ITS).

PR #2054, opened 2021-07-13
Update and add MIGS sample code

larsbrinkhoff:
- Fix parsing of floating point number.
- More MIGS examples.


Rick-Shiffman (approved):
Thanks Lars, I think I may
Have found a error / feature of Muddle 56's parsing of floating point numbers. Work on finding the full extent of the problem. I will post my finding a little later.

Rick-Shiffman (approved):

PR #2009, opened 2021-07-16
TV11 doesn't work without PDP6

larsbrinkhoff:
If ITS is assembled with TEN11P on but PDP6P off, STUFF TV doesn't work.

Also, the expression `IFE <1-PDP6P>\<<T11CPA_-18.>-<PDP6BM_-18.>>` warns about PDP6BM being undefined.  I'm not sure, but I don't think this is the source of the problem because the conditional still evaluates to false as it should.


larsbrinkhoff:
Well, I fixed the conditional but that didn't fix the problem.  Turns out the code inside really is needed even if there is no PDP-6.  So maybe @cstacy was right and there is some implicit assumption somewhere that breaks when the PDP6P code is removed.

larsbrinkhoff:
I was wrong about being wrong.  I forgot that we first tested the TV-11 without any PDP-6 and then it worked fine.  Also, AI back in the day had the TV-11 up and running after the PDP-6 was decommissioned.  The problem was that the KA10 emulator hard coded 3004000 as the base address for the 10-11 address range.  This is fine if there is PDP-6 memory at 3000000-3003777, but if there is no PDP-6 we'd like 10-11 memory to start at 3000000.

I submitted this pull request to fix the emulator: https://github.com/rcornwell/sims/pull/260

eswenson1 (approved):

PR #2059, opened 2021-07-19
Updates for the hello world directory.

larsbrinkhoff:


eswenson1 (approved):
What did an AC value of 2 do in .logout, anyway?

larsbrinkhoff:
I don't know, that's why I think it was a typo.  I checked the UUOS file, but it just refers to any non-zero AC.

ITS source code says

```
ALOGOU: SKIPL SUPPRO(U)
         JUMPN J,ABREAK ;.LOGOUT N, TURNS INTO .BREAK N,33 IF NOT TOP LEVEL
```

I don't understand the `.BREAK N,33` comment.  Anyway, a signal will be raised and handled by DDT:

```
JBIBRK: MOVE I3,I40             ;HERE IF INFERIOR GOT A .BREAK INTERRUPT.
        TLZ I3,(0 17,)          ;THAT CAN BE DUE TO A .BREAK, OR TO A .LOGOUT <NONZERO>,
        MOVE A,[.BREAK 16,160000]
        CAMN I3,[.LOGOUT]       ;WHICH SHOULD BE TREATED AS A .BREAK 16,160000 (SUICIDE).
```

So I think `2,` is exactly the same as `1,` except not idiomatic.

eswenson1:
Thanks. I agree.

PR #2061, opened 2021-08-03
Read the IP configuration from NOP from IMP

bictorv:
With support in KLH10 (dpimp version 1.1.4 or later, in a pull request for klh10), read the IP configuration from the NOP sent. The IP address is in the network, source host, and source IMP fields, and the subnet mask size is in the "handling type" field. The first and last of these are very non-standard, but hey... it works.


larsbrinkhoff:
Looks like it should work with an old/real IMP too, so that's nice.

A corresponding change could be made to IMPOLD WTHNCP, but I'm not going to touch that now because I'm not set up to test it.

eswenson1 (approved):

PR #2062, opened 2021-08-11
Allow unnamed networks

bictorv:
Don't fail to compile H3TEXT if a NET corresponding to a HOST address is missing, just whine.
(As a comment in SYSHST;HOSTS3 says, the network table is almost useless anyway.)

Also, remove many historic NETs (and some HOSTs) from H3TEXT. With the fix to HOSTS3, you don't need the NETs anyway, and they get in the way reading the file.

Finally (and this is where you might disagree I suppose) remove some redundant aliases, and the entry for TT.


larsbrinkhoff:
I updated the branch with my suggested changes, to be folded into previous commits.

bictorv:
Looks good!

larsbrinkhoff:
I squashed my changes and bumped file versions.  This should be ready to go now.

larsbrinkhoff (approved):

PR #2068, opened 2021-08-31
Update networking docs for KLH10

bictorv:
But I think it can be improved wrt the new build/mchn setup - @larsbrinkhoff ?

### DIFF
--- a/doc/networking.md
+++ b/doc/networking.md
@@ -12,7 +12,8 @@ SMTP mail inbound and outbound is included,
 as well as local mail delivery.
 
 ## IP address
-To change the machine's IP address you need to [rebuild ITS](NITS.md).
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
The use of MCNH is an advanced topic perhaps better explained in another document.  I think the parameter will only affect which section of CONFIG is used.  People with their own ITS name probably understands this.

bictorv:
I agree it's for another document. But I think the parameter could be used in more ways - again a separate discussion. :-)

larsbrinkhoff:
Ok to remove that part of the commit message?

bictorv:
Yes, of course.

eswenson1 (approved):

PR #1361, opened 2021-09-03
SEARCH

larsbrinkhoff:
SYS2; TS SEARCH

This program searches files for a fixed string.  There can be several file names on the command line, and they can use `*` as a wild card.


larsbrinkhoff:
This program searches files for a fixed string.  There can be several file names on the command line, and they can use `*` as a wildcard.

atsampson (approved):
Looks fine.

PR #1327, opened 2021-09-03
SHELL - Unix-like shell

larsbrinkhoff:
SYS2; TS SHELL *source missing*


eswenson1 (approved):

larsbrinkhoff:
The shell has a special feature regarding programs.  In addition to the conventional TS naming, it will also recognize programs having the first file name as ZZ and the second file name adjusted to the right.  Many of the "C programs" (#1764) use this naming.

PR #2071, opened 2021-09-06
Have GitHub Actions finish jobs even if one of them failed.

larsbrinkhoff:
Currently the pdp10-ka job is prone to failing randomly, causing the other jobs to be cancelled.  This change will let them finish, and the failed pdp10-ka job can be restarted.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ jobs:
     environment: Deploy
     runs-on: ${{matrix.os}}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Well, it was a nice idea but apparently there's no way to restart a single job.  You have to restart the entire workflow including all jobs.

eswenson1 (approved):
I don't know enough about this stuff (GitHub actions), but I'll approve under the assumption that Lars knows what he's doing!

larsbrinkhoff:
Thanks!  Unfortunately it didn't have the intended effect, so I'll abandon this.

PR #2073, opened 2021-09-07
DQDEV support (DNS in ITS)

bictorv:
This is a cleanup of the stuff I did back in 2005, previously "documented" at http://its.victor.se/wiki/dqdev. I added support in SUPDUP, and changed the default DNS server to 1.1.1.1.

To make this work, I think the only thing you might need is

```
iptables -I PREROUTING -t nat -s $YOUR_KLH10_ITS_IP -p udp --dport 53 -j DNAT --to-destination $YOUR_DNS_RESOLVER
```

It might not be needed if your ITS has a "real" IP address (rather than a tun address). For Chaosnet support, you need a DNS server which knows about DNS.Chaosnet.NET and allows recursion (for your requests).

Before merging, it needs some more/better documentation, and perhaps some testing. (Some of the fixes in DQDEV I admit I haven't understood now - there are some odd comments, but "it seems to work".)

How do I make this a "draft" pull request, or is that not necessary?

### DIFF
--- a/src/sysen2/name.558
+++ b/src/sysen2/name.558
@@ -264,6 +264,15 @@ $$ANALYZ==1
 .INSRT SYSTEM;CHSDEF
 .INSRT SYSENG;NETWRK
 
+DODQ==:1
+ifn DODQ,[
+RESOLV"$$DQCH==:1
+.INSRT SYSNET;RESOLV
+]
+
+;; Whether reagan.ai.mit.edu is used or not
+ifndef $$dorgn, $$dorgn==0
+
 putchr:	.iot tyoc,t
 	popj p,
 
@@ -804,10 +813,20 @@ fjcl55:	movem a,hstcur		;store # as current host.
 	typi "[
 	push p,b
 	move b,a
+ifn DODQ,[
+	move a,[440700,,hanlst]
+	pushj p,resolv"hstsrc
+	 jrst [	move a,hstcur
+		pushj p,typehn
+		jrst fjcl56 ]
+	movei a,hanlst
+]
+.else,[
 	pushj p,netwrk"hstsrc
 	 jrst [	move a,hstcur
 		pushj p,typehn
 		jrst fjcl56 ]
+]
 	typz (a)		;type official name of site.
 fjcl56:	pop p,b
 	typi "]
@@ -1090,7 +1109,13 @@ hstlook:
 	movei a,(p)		;B may contain host name in asciz,
 	tlnn b,-1		;or if lh is 0, it is address of asciz string.
 	 move a,b
+ifn DODQ,[
+	hrli a,440700
+	pushj p,resolv"hstadr
+]
+.else,[
 	pushj p,netwrk"hstlook	;convert to host number.
+]
 	 setz a,
 	move b,a		;return that in B.
 	sub p,[1,,1]
@@ -1293,6 +1318,7 @@ ncvt63:	sojge c,ncvt62
 ;Look through the host table to find all the Lisp machines,
 ;or all Lisp machines associated with the machine we are on.
 ;Contact them eight at a time using all 16 I/O channels.
+;#### More efficiently, use a Chaosnet broadcast (BRD) and collect answers?
 lmfing:	pushae p,[a,b,c,d,e,t,tt]
 	.iopush ntsi,
 	.iopush tyoc,
@@ -1304,7 +1330,7 @@ lmfing:	pushae p,[a,b,c,d,e,t,tt]
 	movem c,lmintb
 	.suset [.smsk2,,c]
 	setz c,			;C has index into LMADRS.
-ife 1,[				;now we have reagan this works differently...
+ifn $$dorgn,[				;now we have reagan this works differently...
 	trne f,%alllm
 	 jrst [	movei b,hslm	;Find the table of all lisp machines, or
 		jrst lmfin2]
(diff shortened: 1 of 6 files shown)


larsbrinkhoff:
Thanks!

It's not really necessary to convert to a draft but it could be nice to signal the intent.  There's a small line saying "Still in progress? Convert to draft" in the right sidebar on this page.

src/sysen2/name.558
```diff
@@ -264,6 +264,17 @@ $$ANALYZ==1
 .INSRT SYSTEM;CHSDEF
 .INSRT SYSENG;NETWRK
 
+;; default off. Compile with /T to input DODNS==:1
+ifndef DODNS,DODNS==:0
```
larsbrinkhoff, line 268: Minor tiny silly comment, but I'd say DNSP would be more idiomatic.

larsbrinkhoff:
I think it's polite to ping @cstacy and explain the DQDEV code is being hacked here.

larsbrinkhoff:
Reviewing the changes to NAME took me on a tour in time to see the ups and downs of ITS. 

```
SYSEN2; NAME 482        1983-02-07
        ; List of ITS hosts
hsits:  irps x,,ai ml mc dm

SYSEN2; NAME 493        1983-03-02
        ; List of ITS hosts
hsits:  irps x,,ml mc dm

SYSEN2; NAME 516        1983-12-01
        ; Too-short list of ITS hosts
hsits:  irps x,,ml mc

SYSEN2; NAME 522        1984-04-01
        ; Too-short list of ITS hosts
hsits:  irps x,,mc

SYSEN2; NAME 551        1986-03-01
        ; Too-short list of ITS hosts
hsits:  irps x,,[mc,ai]

SYSEN2; NAME 554        1986-04-02
hsits:  irps x,,[AI,MX,MC]
        sixbit /x/

File: doc/networking.md [hunk 2]
```diff
@@ -52,7 +52,44 @@ If you use Chaosnet, you may be interested in joining the Global
...
+	```
+5. Make sure your ITS system can reach a DNS resolver which allows recursive queries.
+ 	If you don't use Chaosnet, the default resolver in DQDEV, 1.1.1.1, should work fine as long as packets from ITS reach it.
+	You might find the `iptables` incantation below useful:
+	```
+	iptables -I PREROUTING -t nat -s $YOUR_KLH10_ITS_IP -p udp --dport 53 -j DNAT --to-destination $YOUR_DNS_RESOLVER
+	```
+
+	If you use Chaosnet, you need a DNS resolver which knows how to find Chaosnet data, e.g. from the server at DNS.Chaosnet.NET (which does NOT allow recursion).
+	Get in touch and I'll help you (@bictorv)!
```
larsbrinkhoff, line 92: Maybe mention ..NEW. (DAT) here too?

src/sysnet/resolv.35
```diff
@@ -146,13 +175,20 @@ CVH3NA:	PUSH P,B
 	CAIL B,1000		; If any of high 3 bits were set,
 	 JRST CVH3N3		;  it must be a HOSTS3 strange-fmt addr.
 	JUMPN B,CVH3N2		; If not zero, then must assume HOSTS2 fmt.
+;; Noone uses Arpanet anymore - but some use Chaosnet.
```
larsbrinkhoff, line 178: I WANT TO USE ARPANET!

doc/networking.md
```diff
@@ -52,7 +52,44 @@ If you use Chaosnet, you may be interested in joining the Global
 Chaosnet: read more about it at https://chaosnet.net.
 
 ## DNS
-Check out this [external guide](https://its.victor.se/wiki/dqdev)
+To make ITS use DNS like a modern netizen, you need to do the following:
+
+1. Compile the handler for the DOMAIN: device, which interfaces to DNS.
```
larsbrinkhoff, line 57: Explain about DQ, DQX, and DOMAIN?

src/sysnet/supdup.325
```diff
@@ -273,9 +281,24 @@ SUPDU1:	MOVEI A,HSTPAG
 	 .VALUE
 	MOVEI A,JCLBUF
 	PUSHJ P,NETWRK"HSTLOOK	;GET HOST NUMBER INTO A, NETWORK NUMBER INTO TT
+ifn DODNS,[
+	 ;; NETWRK failed, try DNS
```
larsbrinkhoff, line 285: Why can't NETWRK"HSTLOOK do DNS under the covers?

doc/networking.md
bictorv: Of course!

doc/networking.md
bictorv: Yes, somewhere. I'm not sure where to document what really, this file seems to be "user friendly" while explaining DQ/DQXDEV and DOMAIN/DQDEV might be more low-level info?

src/sysnet/supdup.325
bictorv: It would make sense, except it would (under the hood) make many programs larger and more complex/slower. Consider e.g. PEEK which is in a sense a real-time program, and might need to resolve addresses quite a few times per time unit in some circumstances. 

The other way around might be more reasonable: let RESOLV use NETWRK under the hood? 

But it's really unsustainable to have "all these programs" do both, in a manual fashion, reproducing the same code.

doc/networking.md
larsbrinkhoff: Info can go in INFO. :-)

src/sysen2/name.558
bictorv: Does it make sense to require /T, or would .TTYMAC be better? I'm assuming most people would not want the DNS complication (but instead the H3TEXT complication), but I'm not sure.

src/sysnet/supdup.325
larsbrinkhoff: I see your point about size/speed considerations.  I don't know what PEEK does with hostnames.  Would it make sense for programs to say $$DNS==0 or 1 before .INSRTing NETWRK to control whether to use DNS or not?  It seems to me HSTLOOK is the natural entry point for any program that want to look up a host name, and it might be possible to extend this API in a way to make all callers happy.

bictorv: And then again, NETWRK is not only about name/address resolution, and RESOLV is not only about name<->address but also other DNS things. Hmm.

larsbrinkhoff:
/T would be ok.  Or people can also just edit the file.

src/sysnet/supdup.325
bictorv: Ho hum, reading the code is a way to learn. DQDEV already uses NETWRK if asked for it (e.g. domain:HOSTS3;ch;a;no.nocrew.org rather than domain:QUERY;ch;a;no.nocrew.org), so making RESOLV first ask for HOSTS3 and then DNS data should be easy - not efficient, but easy. ;-)

bictorv:
I'm still working on this - there is plenty still to do - but I find it a bit convenient to have the sum of all the commits visible in a PR. (Maybe there is another way?)

larsbrinkhoff:
I think it's perfectly fine to have an evolving work in progress as a draft pull request.

larsbrinkhoff:
Is this still in (intermittent) progress?

bictorv:
It is, indeed, but as you say, intermittently. Will find the time!

Issue #2051, opened 2021-07-02
Fix timestamps


larsbrinkhoff:
Update the file build/timestamps.txt and be careful to ensure they are correct.

Also look into using a version of touch or gtouch that handles symlinks well.

larsbrinkhoff:
See comments here: https://github.com/PDP-10/its/issues/2044#issuecomment-870858093

cstacy:
Please see this file 2016-0702-touch.txt (comments and testing).

[2016-0702-touch.txt](https://github.com/PDP-10/its/files/6757445/2016-0702-touch.txt)


Also attached is a build illustrating the issue.

[2021-0702-build.txt](https://github.com/PDP-10/its/files/6757447/2021-0702-build.txt)

larsbrinkhoff:
Thanks, fixing touch to work right with softlinks is also part of this.  Added to the top description.

larsbrinkhoff:
@cstacy, you ask:

```
QUESTIONS:

* Are some people NOT getting the error messages?

* I'm not sure what is intended.
  Are there supposed to be files there at this point?
  Or is it maybe trying to create artifacts that are
  to go onto a virtual tape? With the -h flag it seems
  that symlink are supposed to be involved...
```

I don't get the error messages running on Ubuntu 16:

```
Synchronizing submodule url for 'tools/pdp6'
git submodule update --recursive --init `dirname tools/pdp6/.gitignore`
build/stamp.sh build/timestamps.txt
mkdir -p out/klh10/stamp
touch out/klh10/stamp/touch
mkdir -p out/klh10/system
```

The intent is that `touch -h` should create a symlink and set its timestamp.  Maybe the touch you ran doesn't create symlinks?  If so, the script could be updated to first make the symlink, then use touch -h to set its timestamp.

larsbrinkhoff:
@cstacy and @eswenson1, what happens on a Mac for various permutations of touch, gtouch, /usr/bin/touch, etc using -h and a filename that either doesn't exist, or exists and is a symlink?

cstacy:
See my notes!


:)

cstacy:
"touch -h" on all platforms is documented NOT to ever create any kind of file. It does not create symbolic links. This is POSIX.

In addition to the test in my notes above, I also tested it just now on a Linux machine. It works correctly there, too: no file is created.  It gives the "No such file" error message.

The only difference between platforms is that on some of them, it fails silently rather than giving the error message.

[...]

PR #2074, opened 2021-09-14
Correct file timestamps

lashtear:
Several files in `EMACS1;` have a timestamp in the gap caused by
EST->EDT timezone shift of 1989, likely as a result of
daylight-savings handling limitations in ITS at that time.

See https://github.com/PDP-10/its/issues/2044#issuecomment-870858093
and the later comments in #2051 .

The solution is to advance the three files in question one hour, which
is what they would have had if the timezone flipped at 2AM.


larsbrinkhoff:
Thank you.  I have been suspecting the timestamps have been mangled worse than just by one hour.  I'm not sure exactly what happens to the timestamp on the tape as it passes through itstar, then goes to disk, and finally is printed by ls.

Take the VT52 4 file for example.  The bits on the tape say "1989-04-01 19:11:39", but what I wrote in timestamps.txt was "198904020211.39".  So that's off by seven hours.

If I set TZ=EST, I do get the right timestamps from the extracted files on my disk.

larsbrinkhoff:
I made a very slow shell script which should generate updated timestamps.  I will update this pull request when it has finished, in a few hours or days.

larsbrinkhoff:
I have now carefully checked the timestamp bits on tape, and compared against those extracted to disk.  They seem to be the same except the first are (implicitly) local Massachusetts time - maybe with or without daylight saving, and the disk timestamps are in GMT.  My confusion came from running `ls` or `stat` and relying on the result, but that's converted to *my* local timezone which is off by around six hours.

I have now regenerated timestamps with TZ=EST.  Maybe this will be off by one hour during summers, but I'm willing to live with that for now.

eswenson1 (approved):

lashtear:
I'd been thinking about shifting everything to UTC for clarity, but using EST (or EST5EDT maybe) does make sense for keeping things matching original data.

There are a couple places where current time is set (e.g. for the running machine) from localtime, which probably should get the same TZ treatment.

Thanks for looking!

lashtear:
|Error|Line|
|:---|---:|
`touch: invalid date format ‘19780201141707’`| scheme/docume.89 19780201141707
`touch: invalid date format ‘(198604220323.02’`| inquir/inquir.(dump) (198604220323.02
`touch: invalid date format ‘(198106271729.00’`| maxtul/mcldmp.(init) (198106271729.00
`touch: invalid date format ‘(198001192010.37’`| scheme/nschsy.(init) (198001192010.37

Not sure what is causing the `SCHEME;` one to not have a dot; the `.00` case shows up on other lines at least.  The other three are an extra paren-- which tool is generating the list?  (May be reasonable to add it if it can do this from old tape images in an arbitrary way)

larsbrinkhoff:
Thanks for checking.  I'm fixing those errors.

There is no nice tool for generating the timestamps.  It's a mess of scripts I make up on the fly pulling data from various places, plus lots of manual editing.

(I think that would be an appropriate collective noun.  You know gaggle of geese, parliament  of owls, mess of scripts.)

Issue #2, opened 2016-11-18
EMACS


larsbrinkhoff:
I'd like to build EMACS from source code, but that seems to be too much effor for too little gain.

I'm going to try to find a small number of binary files which are sufficient to build a new EMACS.

larsbrinkhoff:
I have now installed a minimal set of binaries for just running EMACS.  No source code.

Another set of binaries will have to be added to rebuild EMACS.

larsbrinkhoff:
Don't forget to see if Rich Alderson's EMACS 170 has some interesting updates.

larsbrinkhoff:
I think we now have a good selection of EMACS libraries installed.  I'd like to do future additions in small batches.  We can't install everything, because that will overflow the EMACS1 directory.

See #250 for a long list of libraries to tick off.

eswenson1:
You mentioned not being able to install all EMACS libraries due to overflowing EMACS1. I'm not sure I understand. The source to EMACS libraries is in EMACS1.  Clearly, you can include all the EMACS library source that AI/MC had in EMACS1.  Are you talking about additional sources that lived elsewhere (e.g. in TOPS20 tapes)?  If so, I think the additional sources should go into a new EMACS2.  Or are you worried about not being able to have all the :EJ files in EMACS?  Perhaps we'll have to create an EMACLB (emacs library) directory or some other directory (EMACSn) for the additional :EJ files and update the autoload properties (if necessary) for those to support loading :EJ files from both EMACS; and the new directory.  

Our initial goal, however, should be to include all libraries we can (assuming source can be found) for which there were :EJ files (or \[...\] files) in EMACS; This, by definition, can't overflow EMACS, if these binaries were already there in the original distribution.

larsbrinkhoff:
The set of AI and MC libraries are not identical.  AI has some that MC don't, and vice versa.  When I tried to combine them, the result didn't fit in EMACS1.

I was thinking EMACS2 for the additional libraries, and EMACS0 for :EJ files if needed.

eswenson1:
Ah. Makes sense. I agree with your proposal.  And EMACS0 makes good sense for the :DJ files.  I was "bothered" by the notion of EMACS1 and EMACS2 for sources, and then what name to use for binaries -- EMACS0 is a good choice.  We could also care a XXXLIB directory, which parallels what was done for LISP (LSPLIB).  But what to use for XXX (EMA, EMC, EMS) or EMCSLB?  :-)

PR #2082, opened 2021-09-28
Various EMACS additions.

larsbrinkhoff:
Add documentation for EMACS modes.  Add Fortran mode.  Add Scribe source for EMACS manual.  Fix some version numbers.


eswenson1 (commented):
Are you really able to build with the new files in emacs and emacs1?  These directories are both full on ES-ITS and I have to keep moving files to archives and other directories when we add more to these....

larsbrinkhoff:
It seems to build ok.  There's only one new EMACS1 file.  Most of the files are INFO documentation.

We noted the overflowing EMACS directories already back in issue 2: https://github.com/PDP-10/its/issues/2#issuecomment-268054975  Maybe it's time to implement that soon.

eswenson1 (approved):

larsbrinkhoff:
Even with all the COMPRS files lying around in EMACS1, I was able to create 39 more files there.

larsbrinkhoff:
Eric, you mentioned archives.  Maybe another option besides splitting off EMACS0 from EMACS would be to add code to search archives in EMACS.

eswenson1:
It is probably because I have a lot more files in EMACS and EMACS1 on ES than we have in DB then.

larsbrinkhoff:
Added autoloads for the programming language modes.

PR #2100, opened 2021-12-13
New text functions added to MIGS.

larsbrinkhoff:
New Muddle functions `SWTEXT`, `SWTXT2`, and `SWTXT3` for plotting text.


eswenson1 (approved):

larsbrinkhoff:
Looks like Circle CI has imposed a 60 minute limit on their builds, so we will not be able to use them any more.

PR #2101, opened 2021-12-16
New IP for UP

bictorv:


larsbrinkhoff:
The first commit is ok, but the second results in this:

```
*$l  sys;ts name
*debug/'NAME$:   -1   0
$g
New sys, must initialize--
Take paws off keys and wait
$^K
*:pdump sys;dsk:ts name
*$p
Init done
:$ Urk! AAAAAAIIIIIEEEEeeeeeeeeee... $
```

This typically happens when there's something wrong with the host table.  Maybe if you run `:syshst;hosts3` manually with the updated table, it will tell you if there's something it doesn't like.

bictorv:
Sigh. syshst;hosts3 is too clever for its own good, and rejects two hosts with the same address. I'll try to uncommit the second commit.

larsbrinkhoff:
Depending on your branching model, you might typically remove the latest commit by doing `git reset --hard HEAD~` rather than revert.  I did this now and force pushed the branch.

bictorv:
Thanks! While I have a "pro" account, I'm not really a git pro. ;-)

larsbrinkhoff (approved):

PR #2105, opened 2022-01-12
New domain for UP

bictorv:
and moved Router.Chaosnet.NET to MX12, including its address. Removed MD.Update.UU.SE which hasn't run for 15 years, and Greek which will probably never run again, and a demo LambdaDelta.


eswenson1 (commented):
Why the weird capitalizations (dFUPDATE)?  And why did you leave UP.UPDATE.UU.SE as an alternative hostname for UP?

bictorv:
Because it's a weird abbreviation for "datorföreningen Update". And because there might be references to the old domain somewhere.

PR #2108, opened 2022-02-05
Added new MIGSPK from RRS.

eswenson1:


Rick-Shiffman (approved):
I t looks good to me #2108

larsbrinkhoff:
Also removed previous MIGSPK URS107 version.

larsbrinkhoff (approved):

eswenson1:
I actually didn't delete the previous version on purpose because I thought it would be good to allow RRS/others to test out the new version and revert locally to compare behavior. 

But I realize that this practice can lead to leaving older versions in place and forgetting to clean them up. But it was the practice "in the old days". You'd add a new version (keeping the existing one) and then as time passed, the older versions would get reaped/deleted.

PR #1907, opened 2022-03-16
BKPH display hacks for TV

larsbrinkhoff:
BKPH; TS CIRCLE  
BKPH; TS BULL  

Display hacks.

As seen on film.  http://projects.csail.mit.edu/video/history/aifilms/x05-graphics.mp4


larsbrinkhoff:
We now have some source files to add here, courtesy of Berthold Horn.

I found one more program, GRADE, which appears in the film (http://projects.csail.mit.edu/video/history/aifilms/x05-graphics.mp4), and RANDY which does not.

Unfortunately I did not find these:

 A program that draws moiré patterns in one quadrant and then copies it to the rest of the screen.  It looks like it *progbably* just draw XOR'ed lines from the center to the perimeter.
Screenshot from 2022-03-16 17-49-07

Gradients in a 2D field, maybe?
Screenshot from 2022-03-16 17-49-45

Nice swirls, nonlinear system?
Screenshot from 2022-03-16 17-50-08

eswenson1 (approved):

Issue #2099, opened 2021-12-13
Non-existent device


arpruss:
I followed the direction as best I could by running make `EMULATOR=simh`, and then `./start`, on my Raspberry PI 3. Here's what I got:
```
PDP-10 simulator V4.0-0 Current        git commit id: e1db7dec
out/simh/boot-6> at dz0 10004
Listening on port 10004
out/simh/boot-8> at dz0 line=7,10018
Line 7 Listening on port 10018
out/simh/boot-10> at tty line=6,10019
Non-existent device

 DSKDMP
```
I then typed `its` and got `DIRNF`.

Is there a step I missed?

larsbrinkhoff:
The line that says `at tty line=6,10019` has a mistake.  The `tty` should be `dz0`.  But that's not fatal.

Getting `DIRNF` is a problem.  I will try to reproduce it myself.

Some things you can try:
- Rebuild using `make clean all EMULATOR=simh`.
- Get disk images directly from here: http://hactrn.org/images/
- Build using another emulator.  Do `make clean` first.

jeffhhk:
The title of this issue and the repro have two different root causes.  I'm going to comment based on the title of the issue.  I will follow up with comments on the higher-impact repro on a separate case.

When I build with:

```
make clean all EMULATOR=simh
```

The following code removes the (low impact) "Non-existent device" message:

```
diff --git a/build/mchn/DB/boot b/build/mchn/DB/boot
index e9d1d845..dd22deb1 100644
--- a/build/mchn/DB/boot
+++ b/build/mchn/DB/boot
@@ -7,7 +7,7 @@ at dz0 10004
 # VT52
 at dz0 line=7,10018
 # GT40
-at tty line=6,10019
+#at tty line=6,10019
 set rp0 rp06
 at rp0 out/simh/rp0.dsk
 b rp0
diff --git a/build/simh/boot b/build/simh/boot
index 08d2e656..207aa7ed 100644
--- a/build/simh/boot
+++ b/build/simh/boot
@@ -7,7 +7,7 @@ at dz0 10004
 # VT52
 at dz0 line=7,10018
 # GT40
-at dz0 line=6,10019
+#at dz0 line=6,10019
 # Dial-up
 at dz0 line=5,10015
 set rp0 rp06
```

larsbrinkhoff:
Thanks for checking, @jeffhhk.  The mistake is that the simh configuration refers to `tty` instead of `dz0`.

[...]

PR #2114, opened 2022-04-20
Update SIMH configuration

larsbrinkhoff:
- Replace `at tty` with `attach dz` for simh, see #2099.
- `set mta enabled` for pdp10-kl.


rcornwell:
Yes, for pdp10-ks, imp and ch11 should be enabled and configured into system.

eswenson1 (approved):

Issue #2123, opened 2022-10-12
H3MAKE has hardcoded list of email recipients for error emails


eswenson1:
H3MAKE has a hardcoded list of email addresses to use when sending emails due to errors converting the h3text file.  That list is currently:

{{{
;; People who should get our warning messages
kvtusr: irp who,,[SRA,Alan,JTW,cent,MAP]
                sixbit ∫who∫
        termin
        repeat  10,[ ? 0 ]              ; are there this many ITS users left?
}}}

Recommend we change this to a more general recipient like others in .MAIL.;NAMES >.  This would allow system maintainers to set appropriate recipients for this mailing list.

PR #2124, opened 2022-10-12
Updated h3make to use h3make mail recipient for error emails, rather

eswenson1:
than hardcoded list of non-existent users.

Updated .mail.;names > to define entry for h3make mail recipient.

Resolves issue #2123.


larsbrinkhoff (approved):
Looks good.

PR #1158, opened 2022-11-03
LODER, transfer programs to a PDP-11

larsbrinkhoff:
SYS; TS LODER  
SYS3; TS LODER  
RUG; TS LODER  
RUG; TS MLODER  
PDP11; TS LODER  
SYSEN2; LODER 20

Load a PDP-11?  Maybe trough 10-11?  Something special if named MLODER?  Something about RUG?


eswenson1 (approved):
Do these actually work with an 11 simulator connected?

larsbrinkhoff:
That's the idea.  I tested the "ITS" program, and it seems not to work great.  I have not tested ITSCOM, so I assume it works perfectly.  Same with RUG.

A PDP-11 (emulated or real) needs a serial line to ITS to use these programs.  Some of them can optionally use the 10-11 shared memory to boost transfer speed.

CC @kb1dds FYI.

PR #2131, opened 2022-11-22
Fix for DDT.md.

larsbrinkhoff:
$$^S is like :CWD, not $^S..  Bug discovered by @PiDP.


eswenson1 (approved):
It's funny, but when I read that originally in this doc, I knew it was wrong -- I always use $$^S -- but I forgot to say anything about it.

PR #2135, opened 2022-11-28
Use correct file timestamp format for all entries.

sebras:
The timestamps for four entries were malformed. Assume that the entries are missing a seconds entry and that it should be zero. This might not be accurate, but can be corrected in a later commit. At least this way a warning from touch can be avoided.


eswenson1 (approved):

larsbrinkhoff:
I checked what the correct timestamps are and updated this pull request.

larsbrinkhoff (approved):

sebras:
> I checked what the correct timestamps are and updated this pull request.

Thank you! I didn't know how to do that, but I knew how to get rid of the error message at least. :)

PR #2141, opened 2023-01-12
Re-run the new DSKDMP to ensure it's swapped out to disk.

### LINKED ISSUE #2099: Non-existent device

larsbrinkhoff:
The last time DSKDMP is booted off tape, the newer on-disk DSKDMP must be run to ensure that version is written to disk.

Fixes #2099.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,9 +4,10 @@ env:
   BASICS: yes
   CONFFLAGS_USR: -DKLH10_CTYIO_INT=0
 
+freebsd_instance:
+  image_family: freebsd-13-0
+
 task:
-  freebsd_instance:
-    image: freebsd-12-1-release-amd64
   env:
     matrix:
       - EMULATOR: simh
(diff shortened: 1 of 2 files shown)


eswenson1 (approved):
I think we also need to update the dskdmp tape to make sure we have the latest dskdmp (I think we do this for pdp10-ks) and we should export the dskdmp 217bin (or whatever) to the host and populate the klh10 directory with it so that we don't use an old version of dskdmp with klh10.

PR #2142, opened 2023-01-15
Fix new directory pack allocation, and distribute files more evenly

### LINKED ISSUE #2049: File system not balanced between disk packs.
`^F` in a recent build display this:

```
FREE BLOCKS #2=280 #3=1009 #0=1277 #1=8419 
```

Disk pack 2 is almost full, and 1 is almost unused.

larsbrinkhoff:
This makes one bug fix, and one enhancement:

1. When creating a new directory, the `QSALLO` field was not updated.  This results in a random value being left, which confuses the file disk pack allocation later.  -1 means the directory has no allocation, so files can be put on any pack.
2. Update the file disk pack allocation to cycle through packs, which gives a more even distribution.

Fixes #2049.


eswenson1 (approved):
Looks good!

PR #2150, opened 2023-02-21
Added build of Zork (616 point version + 100 point endgame) that uses MDL 55

eswenson1:


larsbrinkhoff:
Fantastic!  A long-time dream coming true.

Good idea putting the zork build script in its own file.

src/lcf/rooms.395
```diff
@@ -7,7 +6,7 @@
 
 <GDECL (MUDDLE) FIX (TENEX?) <OR ATOM FALSE> (VERS DEV SNM SCRATCH-STR) STRING>
 
-<DEFINE SAVE-IT ("OPTIONAL" (FN <COND (<L? ,MUDDLE 100> "CFS;MADADV SAVE")
+<DEFINE SAVE-IT ("OPTIONAL" (FN <COND (<L? ,MUDDLE 100> "LCF;MADADV SAVE")
```
larsbrinkhoff, line 9: Why did you change this?  We discussed having sources in LCF and the save file in CFS.

larsbrinkhoff:
CC @heasm66

larsbrinkhoff:
I would like to suggest rearrange things a little bit.  For example, there's a Y2K fix that's logically separate from the rest.  But I know it's painful to rewrite git history, so I can do it myself if that's ok?

eswenson1:
Sure! Go for it!

src/lcf/rooms.395
eswenson1: Because the SAVE file should end up in MADMAN, not CFS.

eswenson1: The implementers used the script to produce the SAVE file, which was then tested and moved (manually) to the MADMAN directory. It seemed more logical to me to put the NBINs in the LCF directory (the RECORD files will end up there anyway — even if you tell the compiler to put the NBIN in a separate directory).

larsbrinkhoff: I checked backup records, and I see MADADV SAVE resided in the MADMAN directory until February 1979.  After that it was moved to CFS.  There was also a CFS; MADADV XXFILE, so your hunch was right!  The DOC, HELP, and INFO files were also moved to CFS.

larsbrinkhoff:
Eric, is this XXFILE useful here? https://github.com/PDP-10/its/issues/832#issuecomment-688485007

eswenson1:
I wasn't able to get that XXFILE to work. First of all, there is no ITS version of LITLPK available (that I can find).  The version that I can find is for TOPS-20 (it references <MDLLIB>).  And since this mini package manager only looks in ONE directory for an FBIN, and the XXFILE does a <USE "CLEAN" "PURITY"> and nowhere can I find evidence that those two packages lived in the SAME (one) directory, that <USE> can never be satisfied by the implementation in LITLPK.  

RRS has hacked a version of LITLPK -- but his, too, only looks in one directory and only for ">" files, and not FBINs or NBINs, it isn't immediately useful as a substitute.

Finally, this XXFILE uses a saved image (MUDDLE;M55UNI SAVE), which I can find in ToTS, but for which I can't find a means to recreate.

The XXFILE that I conjured up works -- albeit without all the optimizations that CFS;MADADV XXFILE has.  However, with our ITS systems' larger memory (than DM), they appear not to be necessary.  Also, we're not as concerned about sharing pure images.

larsbrinkhoff:
I think it's clear that historically, MADADV SAVE was first in MADMAN and later in CFS.  The switch-over was around February 1979.  I think this would allow us to put the older 1977-78 binaries in MADMAN, and the freshly built save file in CFS.  So the new file name wouldn't to collide with the old.

Checking the records, I see that NBIN files were a little all over the place.  Early on in a user directory or MUDBUG, then in LCF, and later in CFS.

This information is from the .TAPE files, which has more data than the few tape images we have.

larsbrinkhoff:
The last "real" TS ZORK from January 1979 loads MADMAN; MADADV SAVE.  It doesn't have any assembly info.  There's a newer TS RBYE (previously discussed: https://github.com/PDP-10/its/issues/832#issuecomment-457338954) loads CFS; MADADV SAVE.  It was assembled by MARC from AR1: CFS; ZORK 2.

src/lcf/rooms.395
```diff
@@ -55,7 +54,7 @@
 		      <TELL ,FLUSHSTR1>
 		      <QUIT>)>
 	       <COND (<AND <TYPE? <SET STV <STARTER>> FIX>
-			   <G=? .STV 3>>
+			   <G=? .STV 6>>
```
larsbrinkhoff, line 58: What is this?

src/lcf/rooms.395
```diff
@@ -310,7 +309,8 @@
 	<SETG INTIME <DSKDATE>>
 	<COND (<L? .MUDDLE 100>
 	       <AND <G? <LENGTH .XUNM> 2> <=? <SUBSTRUC .XUNM 0 3> "___"> <QUIT>>
-	       <SET FN <ITS-GET-NAME .XUNM>>)
+"	       <SET FN <ITS-GET-NAME .XUNM>>)"
+	       <SET FN .XUNM>)
```
larsbrinkhoff, line 314: And this?

src/lcf/rooms.395
eswenson1: This basically neuters the calculation in STARTER. This function is supposed to return a small number if the user is allowed to run Zork and a larger number if not. There are various checks. But the last check appears to be trying to perform a calculation on pure memory pages (I think) that was always preventing the running of Zork (returning 5). I couldn’t figure out why, so I raised the limit from 3 to “greater than 5”.

My thinking was that I’d try to understand and tackle that problem later.

src/lcf/rooms.395
eswenson1: The second hack replaces the use of GET-ITS-NAME with the use of simply the XUNAME. I did this so long ago that I’ve forgotten why I had to, but the original code crashed Muddle and this Zork. Again, I’ll look at this again and see about getting GET-ITS-NAME to work.

eswenson1: But maybe we can get this merged first and open tickets for these two issues?

larsbrinkhoff: Thanks for the info, I will add that.  Yes, let's keep it as is for now.  I'm running a test build to ensure things come out right.

larsbrinkhoff (approved):

larsbrinkhoff:
Just a tiny rewrite, all done now.  Running yet another build locally to make sure Zork still runs fine.

eswenson1:
Thanks Lars. I was still waiting because “some checks haven’t completed yet”. Do we care?

larsbrinkhoff:
They were completed now.  Sometimes you have to reload the page to see the updated check statuses.

eswenson1:
Ah, ok.  Thanks for merging.

-- Eric (KC6EJS)


From: Lars Brinkhoff ***@***.***>
Reply-To: PDP-10/its ***@***.***>
Date: Wednesday, February 22, 2023 at 10:33 PM
To: PDP-10/its ***@***.***>
Cc: Eric Swenson ***@***.***>, Author ***@***.***>
Subject: Re: [PDP-10/its] Added build of Zork (616 point version + 100 point endgame) that uses MDL 55 (PR #2150)


They were completed now. Sometimes you have to reload the page to see the updated check statuses.

larsbrinkhoff:
@heasm66, I don't remember exactly but you would probably know: is there any Zork source code file set in the https://github.com/MITDDC/ repositories that seems to be close enough to compile and run earlier versions?  E.g. the April 1978 GSB files seem promising.

heasm66:
My main goal was to make them work under Confusion. For the 1977-12 and 1978-01, I had to patch a lot with later files and change the syntax a bit because Confusion isn't fully implemented MDL. 

1978-04 Is added a bit later after I worked with Zork, but glancing at it now it seems to be a pretty complete set of files. Getting it to work under Confusion would be a little work but now that there's a toolchain for MDL55, maybe it's worth a try?

I tried documenting where I picked and choosed files from when I patched the different versions at https://github.com/heasm66/mdlzork, but that was a while ago...

larsbrinkhoff:
Thanks, @heasm66.  If they work under Confusion, I believe they should work (or be made to work) under MDL 55.

Nothing to jump on now, but could be a nice project for a rainy day.

PR #2155, opened 2023-02-24
Update networking.md

oilcan-productions:
Updating with more detailed steps for KA10/KL10

### DIFF
--- a/doc/networking.md
+++ b/doc/networking.md
@@ -16,7 +16,8 @@ Unless you are running the current ITS on a current version of KLH10 (see [below
 you need to [rebuild ITS](NITS.md) to change the machine's IP address.
 
 ### SIMH KA10 / KL10
-To get the `pdp10-ka` or `pdp10-kl` online with reasonably low effort, use the included SIMH NAT interface via DHCP.
+To get the `pdp10-ka` online with reasonably low effort, use the included SIMH NAT interface via DHCP.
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Thank you!  Wow, it looks like an ordeal and a half.  No wonder I didn't get it to work.

@eswenson1 and @rcornwell, I'd like you to take a peek at the update, please.

eswenson1 (commented):
Have you actually done this and verified that it worked? I can’t (I need different approach).

oilcan-productions:
> Have you actually done this and verified that it worked? I can’t (I need different approach).

Yes, this is how I configured KA10 on my raspberry pi 400. I am re-building as KL10 now to make sure it works the same way

Mike

rcornwell:
So far looks good. You might want to add in about manually setting IP address and how to set up NAT interface like Eric uses.

rcornwell:
Note these configurations will also work for pdp10-ks, which supports the same IMP interface. SIMP needs to be defined in the configuration file of ITS to enable it.

eswenson1:
I *think* I saw in networking.md something to the effect that pdp10-ks didn't support networking.  If so, we should update that.

eswenson1 (approved):

larsbrinkhoff:
Approved, so let's merge this now.

PR #2157, opened 2023-02-25
Change location of MDL 54 pure library to allow old Zork to run.

eswenson1:
- Update MDL 54 interpreter to look in MUDSAV;SAV 54FILE and MUDSAV;FIXUP 54FILE for pure library.
- Update MKSVFL utility to create pure library in new location.
- Update muddle.tcl to run MKSVFL to create dummy pure library.


eswenson1:
Please don't merge yet.  I'm still building this to test to make sure it works.  I don't see why it wouldn't, but I've been known to do stupid things in the past....

eswenson1:
I ran this build and verified that I can now play the old zork by running OZORK^k (and I can run new zork by running ZORK^k). So this is ready to merge if you approve.

larsbrinkhoff (approved):

larsbrinkhoff:
Thanks!

If I understand your emails correctly, a fancier solution would be to have the necessary libraries in the common SAV/FIXUP FILE, right?  But we can keep that on the to-do list for a rainy day.

eswenson1:
Yes, but the way you create those SAV54/FIX54 files is by using PDUMP (we have the source for this, so ostensibly we could run it in a MDL 54).  But as input to PDUMP, you need to have loaded the code from NBINs first.  We don't have a compiler for MDL 54 (we have one for MDL 53, but no interpreter for MDL 53), and therefore we cannot compile updated/hacked sources to get the NBINs for MDL 54.  We might be able to patch the SAV54/FIX54 files we have though, and then load them into the pure library.

PR #2160, opened 2023-02-28
Added files and directories that support the MDL 55 runtime.

eswenson1:


larsbrinkhoff (approved):

larsbrinkhoff:
Oops, forgot about timestamps.

eswenson1:
Do you have some way to capture all the timestamps, from, say an itstar extract, so I don’t have to do this manually?

larsbrinkhoff:
I have no ready to go solution, but what I will usually do is to get it out of a tree of extracted tapes.  Matching on file names and maybe sha1sum if names are not unique.  You have to be careful to set TZ right when listing the files to not get your local time, but "MIT time".

eswenson1:
When I extract the files, I don't see the times for the files that are in previous years.  I see output like:

```
 15478160     12 -rw-rw-r--   1 eswenson eswenson     8594 Jun  7  1974 ./mprog/mudcom.utt045
 15478141      4 -rw-rw-r--   1 eswenson eswenson     1375 Aug 22  1979 ./mprog/fmacro.ubkd10
 15478115      4 -rw-rw-r--   1 eswenson eswenson      450 Apr 11  1976 ./mprog/blt.1
 15478186      4 -rw-rw-r--   1 eswenson eswenson     1239 Jun 17  1976 ./mprog/unload.ubkd08
 ```
 
 How do I get the times?

eswenson1:
Never mind, I see that `ls` has a `--full-time` option.

eswenson1:
I’m writing a program to generate the appropriate entries for timestamps.txt.

eswenson1:
I wrote a program generating timestamps for a directory with ITS directories as children. For all the ITS files that were written in EST, the date/times are correct, and for all those in EST, they are (predictably) 1 hour off. 

So I need to first determine which time zone (EST or EDT) would have been in effect and then adjust all the EDT times by an hour. Should be simple. As soon as I’m convinced the program works correctly, I’ll update all the timestamps in build/timestamps.txt with the MDL library files.

eswenson1:
Fixed program. Ran it and generated a timestamps.txt.  Merged with build/timestamps.txt and successfully ran `make EMULATOR=simh out/simh/stamp/touch`.  Will commit shortly and generate a PR.

PR #2162, opened 2023-03-01
Update networking.md

oilcan-productions:
Simplified the TAP setup script and made it more configurable for additional TAPs


eswenson1 (commented):
None of this works for me — it requires having multiple hardwired (Ethernet) NICs and multiple IP addresses. On my VPS, I only have one NIC and IP address, and have to use the NAT configuration rather than the TAP configuration. Same is true on my laptop, since I have only WIFI networking. 

So I don’t feel I can review this PR.

larsbrinkhoff (approved):
This is a parametrization of the previous document.

PR #2165, opened 2023-03-04
Add MUDDLE pure library maintenance daemon (MUDSYS;TS DEM).

eswenson1:
This PR adds the pure library maintenance daemon used by Muddle.  It looks for work to do based on files in the MUDTMP directory and updates the pure library (MUDSAV;SAV FILE and MUDSAV;FIXUP FILE) accordingly.  

Note: this was a daemon that was run periodically by the MDL BATCH facility and not by entries in the DEMSTR program (which requires the DM daemon (demon, sic) support in ITS.

Until we have the BATCH system in DB and a define task to run DEM at specified intervals, this daemon can be run manually (:mudsys;dem) when the user has updated MUDTMP.


larsbrinkhoff (changes requested):
Please add timestamps.

eswenson1:
Is that really necessary? I modified both source files. I only copied the lock file, but that is just a text file that serves as a lock file. Its contents are irrelevant. I could have created one with “foo” as
Its contents.

Now my other already merged PR with all the Muddle libraries, on the other hand, should have time stamps. Is there some automated way to capitulee them. There are too many files in that PR to do it manually.

larsbrinkhoff:
Oh, I thought they were original files.  In that case, approved.

larsbrinkhoff (approved):

PR #2167, opened 2023-03-06
Added timestamps for all MDL 55 library files added in #2160.

eswenson1:


larsbrinkhoff:
The commit moved many lines around.  I moved them back so it's easier to see what changed.  (We can move things around, but let's do it in a separate commit if so.)

I have two questions:
1. I only see added timestamps for MUDBUG.  Aren't there also LIBMUD, LIBRM1, etc?
2. I see you updated two timestamps in FONTS1, was that intentional?

larsbrinkhoff:
Oh, one more question.  Can you put your program in here somewhere?  It seems useful for the future, especially for getting EST/EDT right.  I'm not sure I did that for every file.

eswenson1:
I’m a bit confused and will need to look into this. My program generated ONLY timestamps for a subdirectory on my local machine with the new files I added in that original PR. I then added the block of timestamps to the end of build/timestamps.txt, and since the file looked sorted, I sorted it.

i didn’t have the FONTS1 directory in my set of files so I don’t see how/why the timestamps for it were touched. My set of files should have included all those I added. That all timestamps were not added suggests something strange — I’ll look into it.

eswenson1:
> Oh, one more question. Can you put your program in here somewhere? It seems useful for the future, especially for getting EST/EDT right. I'm not sure I did that for every file.

Well, I can, but it is a scala project with a JAR artifact that requires the JRE to run….

eswenson1:
I have no idea how the fonts files changed. I understand why only a subset of the files in the MDL55 libraries got timestamps -- when testing out my tool, I temporarily used ONLY the mudbug directory (so that I could check the results on fewer files).  Then, I switched back to all directories, but at the last moment, when I created the final list, I accidentally yanked back a command line that restricted the files to the mudbug directory.  

However, I have NO clue why anything other than mudbug changed.  I have the .txt file that I merged with the timestamps.txt in the repo, and that ONLY has mudbug files in it. So I would ONLY have added mudbug files to the list and then sorted it.  

I'll run my tool again -- this time excluding mudbug directories, and add the results to the END of the current timestamps.txt in this PR.

eswenson1:
I added all the non-mudbug directories. I didn't sort the full list -- just my new entries.  Please review.

larsbrinkhoff (approved):
I reverted the FONTS1 timestamps, and squashed your two commits.  Approved!

eswenson1:
> Oh, one more question. Can you put your program in here somewhere? It seems useful for the future, especially for getting EST/EDT right. I'm not sure I did that for every file.

I've committed the tool here:  https://github.com/PDP-10/its/pull/2171

PR #2170, opened 2023-03-07
Updated zork documentation for old and new zorks using heasm66's recommendations

eswenson1:
Note that now, the old zork (MADMAN) documentation is different from the new work (CFS) documentation.

@heasm66 Please make sure I got the right ones.


larsbrinkhoff (approved):
Looks ok to me with regards to matching Zork versions with timestamps, but I'll let @heasm66 have the final say.

heasm66 (approved):

PR #2171, opened 2023-03-07
Added tool to create timestamps.txt fragment from directory.

eswenson1:
This is intended to be used on a directory containing ITS directories. It recursively walks the tree and generates timestamps entries for each file based on the file creation date/times.


larsbrinkhoff (approved):
This is quite small and nice!

PR #2180, opened 2023-03-24
Test for the 340 interrupt "hack hack".

larsbrinkhoff:
Adding an automated test that runs standalone to check that the KA10 emulator (currently unknown what other processors should do) supports the "hack hack" used by ITS to handle overflowing 340 display BLKO interrupts.  For now, it's just a draft because the KA10 emulator has regressed in this regard.  The problem was previously found and solved in 2018.

Reported here: https://github.com/rcornwell/sims/issues/297

```
;This tests the ITS "hack hack" for handling Type 340 display
;interrupts.  Data is output to the display through a BLKO from a low
;priority data channel, whereas other conditions are handled through a
;higher priority special channel.  The "hack hack" is that when the
;BLKO overflows, ITS triggers an interrupt on the special channel.
;This interrupt is handled still in the overflow state, so the second
;interrupt location should be used.  This hack goes back to at least
;ITS 672.
```

For more detailed analysis, see @aap's description here: http://pdp-6.net/asmhacks.html#pisystem

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ DUMP=$(shell cd src; ls syseng/dump.* sysnet/netwrk.*)
 SMF:=$(addprefix tools/,$(addsuffix /.gitignore,$(SUBMODULES)))
 OUT=out/$(EMULATOR)
 
-all: its $(OUT)/stamp/emulators tools/supdup/supdup
+all: its $(OUT)/stamp/test $(OUT)/stamp/emulators tools/supdup/supdup
 
 its: $(SMF) $(OUT)/stamp/its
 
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
Note that the test currently fails with PC = 115.  A passing test should have PC = 201, which the script tests for.

larsbrinkhoff:
@rcornwell has corrected this, so I'll update this pull request to bring in the fix.

larsbrinkhoff:
Tested running Spacewar, and it's fine now.  Also, the build script now checks this to ensure it will keep working in the future.

eswenson1 (approved):

PR #2185, opened 2023-04-05
Create nat-networking.md

oilcan-productions:
Writing up NAT networking

### DIFF
--- a/doc/nat-networking.md
+++ b/doc/nat-networking.md
@@ -1,0 +1,100 @@
+# Using NAT networking to run multiple ITS machines
+To run multiple different ITS machines on the same physical host you can use the NAT networking capability provided in SIMH and KLH10 PDP-10 emulators.
+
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I saw on IRC that @decuser managed to get his instance to do TCP/IP.  Good job!  Do you have any feedback about the networking.md file or advice to add there?

decuser:
> I saw on IRC that @decuser managed to get his instance to do TCP/IP. Good job! Do you have any feedback about the networking.md file or advice to add there?

I would add something to the effect that wifi bridging is not supported, you will need to bridge an actual ethernet connection.  I would also let the reader know that the bridge and tap are not persistent across reboots, they will need to run the script again if they reboot between runs. Otherwise, I was able to follow it, pretty much verbatim, and it works.

oilcan-productions:
Closing this one for now. I do not have the time to spend on this ight now and it has been around for a long time

Issue #2192, opened 2023-04-20
COMSAT crashes because it can't find local host name


bictorv:
**Problem:**

Since a few weeks, right after boot, I get a message on the console:
```TOP LEVEL INTERRUPT 200 DETACHED JOB # 3, USR:TARAKA CNAVRL 18:53:55```
which means COMSAT crashed, and mail doesn't work, which is quite a drawback.

**Diagnosis:**

Doing :UJOB and :PROBE reveals it crashed at DEATHV+1, and going back up the call chain shows that RESOLV"HSTSRC gave a non-skip return at MAPI20+4. This means it couldn't find the name for our IP address! How come? 
The address (in OWNHST) is gotten from RESOLV"OWNHST (in the MYADDR routine), which reads it from ITS by reading the IMPUS3 symbol.

The IMPUS3 symbol is defined in SYSTEM;CONFIG, but that definition is no longer used for the IP address of the system (since #2061 the IP is received from the IMP at runtime). So when I changed the IP address of my ITS system recently, I just updated the address in my KLH10 init file, and everything just worked - except COMSAT. :-(

**Solution:**
One possibility would be to change RESOLV so it reads the local IP address from the NIFIPA location in ITS instead of from the IMPUS3 symbol. This however would require that all programs using the RESOLV library be recompiled.

Another possibility is to change SYSTEM;IMP which reads the IP address from the IMP (since #2061 ) to also patch the ITS symbol table and update the IMPUS3 symbol with the right value (having read it from the IMP). This would be similar to how #2038 updates MYCHAD after reading the Chaosnet address from the CH11 device..

**Patch:**
Ready, will make a pull request.

larsbrinkhoff:
Option number two would be nice if any programs were to use IMPUS3 directly instead of going through RESOLV.  Or if an IMP doesn't support the IP query extension.

@rcornwell, I think your IMP doesn't support querying the host IP number, right?

bictorv:
> Option number two would be nice if any programs were to use IMPUS3 directly instead of going through RESOLV. Or if an IMP doesn't support the IP query extension.

I agree, so that's why #2193 does it that way.

bictorv:
> So when I changed the IP address of my ITS system recently, I just updated the address in my KLH10 init file, and everything just worked - except COMSAT. :-(

[...]

PR #2193, opened 2023-04-20
Patch the value of IMPUS3 when IP is read from the IMP

bictorv:
Fix for issue #2192.

### DIFF
--- a/src/system/imp.369
+++ b/src/system/imp.369
@@ -629,6 +629,19 @@ IMPIN:	LDB T,[241000,,IMPILB+0]	;get network byte (ip[0])
 	MOVE T,NIFIPA		; make network part of address
 	AND T,NIFIPM
 	MOVEM T,NIFIPN		; save it
+	;; Backwards compatibility for RESOLV"OWNHST (used by e.g. COMSAT):
+	;; patch the IMPUS3 symbol
+	PUSH P,A		;Save the ACs clobbered by symlk
+	PUSH P,B
+	PUSH P,C
+	MOVE B,[SQUOZE 0,IMPUS3]
+	PUSHJ P,SYMLK
+	 BUG HALT,[IMPUS3 symbol not found]
+	MOVE A,NIFIPA
+	MOVEM A,(C)		;Zap the symbol value
+	POP P,C
+	POP P,B
+	POP P,A
 	JRST IMPIRT		;One more NOP from IMP
 
 ;;; IMP->Host Type 5 - RFNM (Ready For Next Message)


bictorv:
Someone may decide this is worth bumping a version number for? Please do.

eswenson1 (approved):
Wow! I didn't know IMPUS3 was no longer used.  I've been careful to set it properly for quite some time for each system I've set up!

eswenson1:
Yes, please *do* increment the version number!

Not doing so makes it difficult to know, by looking at file versions, whether a system is up-to-date.

bictorv:
Good question - let me explain by citing `SYSTEM;INET >`:
```
NIFIPA:	; IP Address for network interface
	IFN IMPP, IMPUS3	; IP address on IMP interface
	IFN CHAOSP, IMPUS4	; IP address on CHAOS interface
IFN .-NIFIPA-NNIFS, .ERR Wrong size table - NIFIPA
```
Here you can see that `NIFIPA` is a location in ITS, while `IMPUS3` is _just a symbol_, defined in `SYSTEM;CONFIG >`, e.g. through
```
DEFOPT IMPUS3==<IPADDR 192,168,1,100>	; IP address
```
In this case, `IMPUS3=30052000544` (the IP address). This symbol is read, e.g. in RESOLV"OWNHST, using the `.EVAL` UUO which examines the ITS symbol table. To modify the _symbol_ value, you need to look up the symbol in the symbol table (which "inside ITS" can be done by the `SYMLK` routine (see `SYSTEM;ITS >`) rather than the `.EVAL` UUO), and then store the new value in the address returned in C (the address of the symbol in the symbol table).

If you were to implement approach One in #2192, RESOLV would have to first use `.EVAL` to read the value of the `NIFIPA` symbol, and then read that location in ITS - every time a program calls RESOLV"OWNHST. Another advantage of approach Two is that the symbol table modification is done only once (when reading the IP address from the IMP), and the lookup (in RESOLV"OWNHST) can be slightly quicker.

Please ask if anything is unclear!

bictorv:
> Wow! I didn't know IMPUS3 was no longer used. I've been careful to set it properly for quite some time for each system I've set up!

Well, it turns out it was still used, but now it's only used "indirectly". Except in `SYSTEM;NCP >`, but that isn't used anymore - or perhaps Lars would say "not yet". ;-)

bictorv:
> Yes, please _do_ increment the version number!

Please do!

larsbrinkhoff:
@bictorv, thanks for the explanation (quite obvious in hindsight).  I have updated the file version.  FYI, here's how to do it:

```
$ git checkout bv-impus3
$ git mv src/system/imp.368 src/system/imp.369
$ git push -f
```

bictorv:
> > Wow! I didn't know IMPUS3 was no longer used. I've been careful to set it properly for quite some time for each system I've set up!
> 
> Well, it turns out it was still used, but now it's only used "indirectly". Except in `SYSTEM;NCP >`, but that isn't used anymore - or perhaps Lars would say "not yet". ;-)

But @eswenson1, please note: in old KLH10 versions, or other emulators, the IP address isn't read from the IMP and the setting of IMPUS3 is still important!

eswenson1:
Thanks, Björn. I think I'm running a reasonably current version of KLH10 for ES and EX, but I'll make sure.

PR #2200, opened 2023-07-04
Build TECO-6 from reconstructed source code

larsbrinkhoff:
The @ TECO binary is replaced with a program built from source reconstructed from old ITS TECO sources.


eswenson1 (commented):
Is the right thing here really to move the entirety of the Muddle build (muddle.tcl) forward, or wouldn’t it be better to simply move the STINK build out of muddle.tcl and before you need it for pdp-6 TECO?

larsbrinkhoff:
I thought about something like that, but then we'd have to decide on a new location or name for STINK so I punted on that.  Suggestions?

eswenson1:
Or you could keep the name and location the same as where muddle.tcl puts it. Just move the required lines from muddle.tcl to the TCL that builds 6 TECO.

eswenson1:
Or make it STKnnn (whatever version it is) and put in SYS3 (and update muddle.tcl) to reference it.

larsbrinkhoff:
Ok, I did the first thing.

eswenson1 (approved):
I assume the MUDSYS directory pre-exists from the tape load so you didn't have to create it?

eswenson1 (approved):

larsbrinkhoff:
The directory isn't created by the build scripts.  I think there's some MUDSYS files on the big "sources" tape.

PR #2205, opened 2023-08-01
use open-simh, not OG simh

athornton:
https://groups.io/g/simh/topic/91528716

In short: Mark Pizzolato made some decisions that almost everyone else, including Bob Supnik, disagreed with vigorously, and put the project under an extremely weird license:

https://github.com/simh/simh/blob/master/LICENSE.txt

Open SIMH is the resulting fork, under an MIT license, and it appears to be where the most actual development is happening.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,7 +12,7 @@
 	url = https://github.com/PDP-10/klh10
 [submodule "tools/simh"]
 	path = tools/simh
-	url = https://github.com/simh/simh
+	url = https://github.com/open-simh/simh
 [submodule "tools/supdup"]
 	path = tools/supdup
 	url = https://github.com/PDP-10/supdup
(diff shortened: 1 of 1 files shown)


rcornwell:
I've thought about merging. However I do testing in sims and when stable push over to open-simh. So there is something to be said to keep them separate.

larsbrinkhoff:
ITS has been held back to an old version of SIMH because newer versions wouldn't build on the KS10.  If OpenSIMH now works, that's great and we should switch over, and also reduce the two copies to one.  I will do some more checking first.

larsbrinkhoff:
With regards to "sims", that is the upstream for KA10 and KL10, so I want to keep that as long as @rcornwell is keeping the arrangement.

larsbrinkhoff:
Well duh, just changing the submodule repository and not the commit will have no effect in what emulator version is run.  So this is fine.

larsbrinkhoff (approved):

PR #2209, opened 2023-09-07
Hello world for MACRO-10 and FAIL.

larsbrinkhoff:
Adding UUOSYM so MACRO program can SEARCH it.  Assemling the UUOSYM universal requires MACTEN.

With UUOSYM in place it's possible to assemble a "hello world" with MACRO.  I also added one for FAIL.  I'm not sure how FAIL programs usually get their system symbols, so for now it includes some OPDEFs.

Finally added XFILE scripts to automatically build the programs.  They are put inside an archive to reduce clutter.  Some scripts use `:sleep` to wait for a compiler running in the background.  I don't like that, but I don't know how to do it better.


eswenson1 (approved):

rmaldersoniii:
> Adding UUOSYM so MACRO program can SEARCH it. Assemling the UUOSYM universal requires MACTEN.

> With UUOSYM in place it's possible to assemble a "hello world" with MACRO. I also added one for FAIL. I'm not sure how FAIL programs usually get their system symbols, so for now it includes some OPDEFs.

FAIL assembles universal files the same way that Macro-10 does (UNIVERSAL UUOSYM in place of TITLE statement), and produces a .FUN symbol file (rather than .UNV).  You probably need to massage the input file for FAIL syntax, if that's what you want to do.

larsbrinkhoff:
> You probably need to massage the input file for FAIL syntax, if that's what you want to do.

Thanks.  Searching the Saildart archive I indeed see a lot of .FUN files.  However, it seems FAIL programs normally don't have any SEARCH statement.  Reading the FAIL source code it looks like it provides built-in symbols for system calls.  But the set of symbols match what it's built for, so on ITS it doesn't have the WAITS UUOs.  The ITS port of SUDS uses an .INSRT file with OPDEFS for the UUOs.  I think, at least for now, having OPDEFs in the FAIL hello world is good enough.

PR #2211, opened 2023-09-09
Start a local Chaosnet

larsbrinkhoff:
`./start chaosnet` starts the Chaosnet bridge (CC @bictorv) with configuration for the "local" subnet 376.  The bridge is 177001 and ITS is 177002.  ITS and emulators are configured for this host address.


larsbrinkhoff:
@bictorv, no hurry for this iteration, but in the future, do you think it would be possible to have some kind of Chaosnet DNS name for some "known" local subnet nodes like 177002?  Or have a local configuration to resolve a name like DB, DB-ITS, LOCAL-ITS, or whatevs.  Something to allow typing `supdup db-its` or similar rather than using the octal address.

build/klh10/start
```diff
@@ -21,11 +21,17 @@ stop() {
     echo "OK"
 }
 
+chaosnet() {
+    (sleep 2; tools/cbridge/cbridge -vs -c build/cbridge.conf >cbridge.log 2>&1) &
```
bictorv, line 25: I wouldn't recommend -vs, this gives so much (irrelevant) chatter that it might slow the net down, not to mention filling the disk. If what you really want is a "startup printout", that can be arranged - or tell me what you're really after?

build/klh10/#dskdmp.ini#
bictorv, line 1: Isn't this an auto-save file from emacs? (And by the way, I suggest using tunaddr rather than gwaddr for the imp devdef, as discussed earlier.)

bictorv:
> the "local" subnet 354

You mean 376, right?

bictorv:
> do you think it would be possible to have some kind of Chaosnet DNS name for some "known" local subnet nodes like 177002? Or have a local configuration to resolve a name like DB, DB-ITS, LOCAL-ITS, or whatevs. Something to allow typing `supdup db-its` or similar rather than using the octal address.

Certainly possible, but let's discuss in other channels.

build/klh10/start
larsbrinkhoff: Ideally I'd like to see a small amount of logging that is useful to see what is happening on the network.  Maybe a line for each connection started or dropped.

build/klh10/#dskdmp.ini#
larsbrinkhoff: Yes, a few files got added by mistake.  I removed them.

larsbrinkhoff:
> You mean 376, right?

Yes; fixed.

build/klh10/start
bictorv: Connections only exist in the NCP, not in the plain bridge. Or do you mean "connection" as in "the first arrival of traffic on a link"?

Another option is using the hostat program, if you're just interested in seeing how much traffic there is (and how many packets are dropped etc)?

larsbrinkhoff: I'm intentionally vague here.  I'm not sure exactly what I want, but it's good to see e.g. incoming supdup connections.

bictorv: I would suggest leaving the -vs out until we figure this out, in the interest of wear and tear on SD cards. Sustainability matters, even in the small.

Supdup connections would not be incoming to cbridge, but to ITS. You can probably do an iptables/netfilter rule to find SUPDUP RFCs, if you like.

bictorv: If you only want Supdup connections (which DO come to cbridge when using the NCP), try "trace on" for the ncp config line and see if that's more like what you wanted? Still gives a lot of output I guess, and iptables/netfilter rules are easier to tweak/rate limit etc.

larsbrinkhoff: It's not overly important, so let's figure it out later.  If the need comes up.

ams:
I am getting:

```
%SIM-ERROR: Eth: open error - No support for tap: devices
```

build/klh10/config.203
```diff
@@ -993,10 +998,6 @@ DEFOPT IMPUS==100.	; IMP net host number (old-style) (100 octal)
 DEFOPT IMPUS3==<IPADDR %IP%>	; IP address
 DEFOPT NM%IMP==<IPADDR %NETMASK%>	; Subnet mask
 DEFOPT PKTTRC==-1	;Packet tracing code enabled
-DEFOPT CHAOSP==%CHAOSP% ;Has CHAOS net
```
eswenson1, line 996: So we can no longer build DB for a specified chaosnet address and only for the local chaosnet address?  This seems lame to me.  If I don't want to use a local chaosnet, but rather have my own bridge or want to use the global router, I may have a chaosnet address and want to use it.  It appears that now I cannot.

build/klh10/config.203
```diff
@@ -967,6 +967,11 @@ DEFOPT XBL==30.		; # TCP network channels
 DEFOPT IMPP==1		; Has an IMP interface
 DEFOPT SSIMP==1		; Hooked to "Simulated Simple IMP"
 
+DEFOPT CHAOSP==1	;Has CHAOS net
+DEFOPT MYCHAD==177002	;CHAOS net address
```
eswenson1, line 971: The chaosnet address of the host should not be hardcoded, but should come from build-time options.

build/mchn/DB/boot
```diff
@@ -2,6 +2,10 @@ set console wru=034
 set cpu its
 set cpu idle
 set tim y2k
+set ch enabled
+set ch node=177002
```
eswenson1, line 6: node and peer should not be hardcoded, but should come from build-time options. If I have my own cbridge and subnet, these will not be good defaults. Just like we allow overriding of the IP address, we should allow overriding of the chaosnet address.

build/mchn/DB/dskdmp.txt
```diff
@@ -20,7 +20,7 @@ devdef imp  ub3   lhdh   addr=767600 br=6 vec=250 ipaddr=%IP% gwaddr=%GW%
 ; Dummy definitions.  Only one DZ is still (apparently) needed.
 devdef dz0  ub3   dz11   addr=760010 br=5 vec=340
 ;devdef dz1  ub3   dz11   addr=760020 br=5 vec=350
-%CHAOSP%devdef chaos ub3  ch11   addr=764140 br=6 vec=270 %CHAOSA%
+devdef chaos ub3  ch11   addr=764140 br=6 vec=270 myaddr=177002 chudpport=44042 chip=177001/localhost:44041
```
eswenson1, line 23: Should not have hardcoded addresses here. Should come from build options.

build/mchn/KA/boot
```diff
@@ -14,6 +14,8 @@ set dtc dis
 set dk dis
 set pd ena
 set pd on
+set ch enabled
+set ch node=177002
```
eswenson1, line 18: Shouldn't be hardcoded.

build/mchn/KA/run
```diff
@@ -41,6 +41,10 @@ set imp ip=192.168.2.101/24
 set imp gw=172.31.1.100
 set imp host=10.3.0.6
 at imp tap:tap0
+set ch enabled
+set ch node=177002
```
eswenson1, line 45: Shouldn't be hardcoded.

build/pdp10-ka/config.202
```diff
@@ -242,10 +242,10 @@ DEFOPT NNETCH==20.	; # NCP net channels
 DEFOPT INETP==1		; Include Internet code
 DEFOPT TCPP==1		; Include TCP code
 DEFOPT XBL==20.		; # TCP network channels
-DEFOPT CHAOSP==0	;CHAOS NET
-DEFOPT MYCHAD==3114	;CHAOS NET ADDRESS
+DEFOPT CHAOSP==1	;CHAOS NET
+DEFOPT MYCHAD==177002	;CHAOS NET ADDRESS
```
eswenson1, line 246: Shouldn't be hardcoded.

build/pdp10-kl/boot2
```diff
@@ -7,6 +7,8 @@ set lpt dis
 set dc disable
 set pd ena
 set pd on
+set ch enabled
+set ch node=177002
```
eswenson1, line 11: Shouldn't be hardcoded.

build/pdp10-kl/config.202
```diff
@@ -515,9 +515,10 @@ DEFOPT IMPUS==106	; ARPA net host number
 DEFOPT IMPUS3==<IPADDR 192,168,1,100>	; Internet host number
 DEFOPT NCPP==0		; Flush NCP code
 DEFOPT NNETCH==30.	; # NCP network channels
-DEFOPT CHAOSP==0	;NO CHAOS NET
-DEFOPT MYCHAD==1440	;CHAOS NET ADDRESS
+DEFOPT CHAOSP==1	;NO CHAOS NET
+DEFOPT MYCHAD==177002	;CHAOS NET ADDRESS
```
eswenson1, line 519: Ditto.

build/pdp10-kl/run
```diff
@@ -24,5 +24,9 @@ set rpa2 rp04
 at rpa0 out/pdp10-kl/rp04.0
 at rpa1 out/pdp10-kl/rp04.1
 at rpa2 out/pdp10-kl/rp04.2
+set ch enabled
+set ch node=177002
```
eswenson1, line 28: Ditto.

build/pdp10-ks/boot
```diff
@@ -3,6 +3,10 @@ set cpu its
 set cpu idle
 set dz 8b lines=8
 at -u dz0 10004
+set ch enabled
+set ch node=177002
```
eswenson1, line 7: Ditto.  Also, for all the other hardcoded chaosnet and bridge addresses that follow in this PR.

drboone:
> + sed -i -e 's/set dpy enabled/set dpy disabled/' -e 's/set wcnsls enabled joystick cscope/set wcnsls enabled joystick/' out/pdp10-ka/run
> sed: -e: No such file or directory

Missing argument to -i (extension to append when renaming the old file).

build/klh10/config.203
larsbrinkhoff: That's right, *DB* ITS will be hardcoded to 177002.  The upside of this is that all beginner users can get started running ITS with a local Chaos network without any modifications.  Users that want their own Chaosnet address and connect to the global network can still do so by setting up another ITS.

build/klh10/config.203
larsbrinkhoff: The main goal for this pull request is to make it easy for people to get started with a Chaosnet.  It's still possible for experts to make your own ITS with another network configuration.

build/mchn/DB/boot
larsbrinkhoff: You're not running DB ITS, but ES, EX, etc, right?

build/pdp10-ks/boot
larsbrinkhoff: I'm going to go ahead with this.  If you want to propose a better way later, please do so!

larsbrinkhoff:
I'm going to merge this now, because I want to move forward letting people use Chaosnet.  This doesn't mean anything is set in stone from now forward.  If someone wants to suggest another mechanism, I'm all ears.

eswenson1:
But you could have done both -- use make/env variables that DEFAULT to the hardcoded addresses but could be overridden by the user.  This is what we do for IP addresses.  

And the answer to your question, above, is that I actually build/use a DB build (for EX).  If you connect to EX, you'll see it is just a DB.

build/pdp10-kl/config.202
bictorv: (At least on klh10, this address is overridden by the devdef address.)

Issue #2210, opened 2023-09-09
Zork crash


13spacemen:
Zork (zork^k) crashes when going east (e) from the Loud Room. It's supposed to go to the Ancient Chasm I think, but it just crashes and asks user to submit a bug report, then *KILL itself.

In the MDL Zork C++ port https://github.com/jclaar/zork it does NOT crash, and goes to the Ancient Chasm

Isn't the source for both PDP-10 ITS Zork and C++ Zork the same 1981 MDL source code?

Why does it crash on ITS?

![zorkcrash](https://github.com/PDP-10/its/assets/46101244/ed9b8963-d13f-4203-af3e-3e44675562b5)

Notes for ITS:

I am running Zork in the GT40 which has done ```GTLOAD^K GT40;VT07 BIN29``` beforehand

But even if I don't use a GUI terminal and just use the regular linux bash, it crashes

In Zork, do these commands from start of game:

- n
- e
- open
- w
- w
- take all
- turn on lamp
- move rug
- open trapdoor
- d
- e
- kill troll (repeat till he dies)
- n
- n
- d
- u
- e
- e
- s [we are now at Loud Room]
- e [CRASH ON PDP-10 ITS] {Ancient Chasm on C++ MDL Zork port}

It no longer crashes on ITS if I do "echo" before doing "e", while in Loud Room.
Is this behavior accurate? Is it supposed to crash if you don't do echo first? Is the fact that C++ Zork does not crash without doing echo first, a bugfix/change by jclaar?

eswenson1:
I can debug and fix this. I’ve played through the whole game but not with this exact sequence.

larsbrinkhoff:
FYI @heasm66

heasm66:
The *Confusion* version of this don't have this problem.

If I do the sequence in my version in PDP-10 ITS I get:
```
>s
Loud Room
This is a large room with a ceiling which cannot be detected from
the ground. There is a narrow passage from east to west and a stone
stairway leading upward.  The room is extremely noisy.  In fact, it is
difficult to hear yourself think.
There is a large platinum bar here.
get bar
get bar
e
I'm sorry, you seem to have encountered an error in the program.
^Z
```

But, if you do the `echo` first it works:
```
Damp Cave
This is a cave.  Passages exit to the south and to the east, but
the cave narrows to a crack to the west.  The earth is particularly
damp here.
>s
Loud Room
This is a large room with a ceiling which cannot be detected from
[...]

PR #2212, opened 2023-09-11
Fixed issue with ECHO-ROOM treating <PRSO> as a vector

eswenson1:
`<PRSO>` can be a scalar, and attempts to use EMPTY? on `<PRSO>` can cause an exception.  Addresses #2210.


larsbrinkhoff:
Nice to see Zork still getting updates!  And this one is making it simpler, too.

larsbrinkhoff (approved):

Issue #2217, opened 2023-09-16
DUMP remote tape drive can only be six characters


larsbrinkhoff:
The Chaosnet remote tape protocol allows any number of ASCII characters in the name specifying the drive.  However, DUMP is using the GETSYL routine to input a single SIXBIT word.  This is then converted to a ASCII drive name.

It's possible to have a remote tape server running on the same host as a PDP-10 emulator.  For the purpose of using the drive name to specify a tape image file, it would be good to allow more than six characters, and also lower case letters.  It shouldn't be very hard to replace GETSYL with something to input ASCII characters directly.

PR #2218, opened 2023-09-18
DUMP updates for remote tape code.

### LINKED ISSUE #2216: DUMP remote tape code uses DMOVE
The magtape DUMP program has DMOVE instructions in the remote tape code.  That doesn't work on the KA10.

larsbrinkhoff:
This is to fix #2216 and #2217.

- DMOVE instructions are replaced with two MOVEs to make it run on KA10s.
- The input specifying the remote tape drive can now be any ASCII string.


src/syseng/dump.448
```diff
@@ -8017,7 +8066,8 @@ READ9:	POPAE P,[TT,T,E,D,C,B]
 ;B - source index in BUFFER
 ;C - number of 4-word blocks
 ;Arguments get advanced, T,TT are clobbered
-RDCNV:	DMOVE T,BUFFER(B)	;First two 32-bit words
+RDCNV:	MOVE T,BUFFER(B)	;First 32-bit word
```
eswenson1, line 8069: I think this is brittle. If someone defines TT to be more than one register away from T it will break. Better to use T+1 in second MOVE.

src/syseng/dump.448
```diff
@@ -8545,7 +8596,8 @@ SCOPY2:	PUSHAE P,[D,E]
 	LDB D,[360600,,B]	;Byte position of source
 	MOVEI TT,34		;Compute left shift required to align it
 	SUB TT,D
-SCOPY3:	DMOVE D,(B)
+SCOPY3:	MOVE D,(B)
```
eswenson1, line 8599: This is much better.

eswenson1 (changes requested):

src/syseng/dump.448
larsbrinkhoff: I understand your point, but it's not that simple.  The code below relies on it being TT.  But it's also doing LSHC and ROTC, so really TT must be T+1.

eswenson1:
Ok. Is there a comment at the definition of TT?

eswenson1 (approved):

larsbrinkhoff:
TT=T+1 is a convention in much ITS code, and it usually is commented so, but here it isn't.  I added an assertion to check it, with an .ERR to signal if there is a problem.

src/syseng/dump.448
larsbrinkhoff: Resolved by asserting TT=T+1 at assembly time.

eswenson1:
Great! Thanks.

PR #2219, opened 2023-09-19
Document terminal ports available for emulators.

diego0001:
Localhost port is now defined by default as 10004, not 10003, as per:

```
at -u dz0 10004
```

on `its/out/simh/boot`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,17 @@ a list of useful commands.
 Alternatively to logging in directly in the system console window, 
 which will always be displaying daemon messages, you may prefer to 
 use a seperate terminal session for logging in.  To do this just
-run `telnet localhost 10003` from another shell window and press
-<kbd>CTRL</kbd><kbd>Z</kbd> to log in there.
+run `telnet localhost <port>` from another shell window and press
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Good catch, but the fix isn't correct since other emulators are configured to listen to 10003.

Here is the allocation of port number to various terminals/controllers: https://github.com/PDP-10/its/pull/1190#issuecomment-417269344

larsbrinkhoff (changes requested):
Either this should mention both 10003 and 10004, or we need to think about having a single port number across emulators.  Also, note that some emulators have many ports open.

larsbrinkhoff:
Probably the README should be updated to document all the ports as per the comment I linked to.  And check if that info is still valid.  Would you like to do that, @diego0001.

diego0001:
> Probably the README should be updated to document all the ports as per the comment I linked to. And check if that info is still valid. Would you like to do that, @diego0001?

Sure, just a moment.

larsbrinkhoff (approved):

Issue #2213, opened 2023-09-12
Make NETIME more useful


larsbrinkhoff:
NETIME has a hardcoded list of hosts which it asks for the current date and time.  None of those are now available, especially not on any Chaosnet.  Maybe we should add some host on the global Chaosnet, and also run a local subnet 376 time server.

Maybe also look into #485.

larsbrinkhoff:
TIMES also has outdated server lists.

larsbrinkhoff:
I understand the cbridge NCP serves TIME requests, right @bictorv?  So for the local network case, we can just add 177001 as a time server.

To make things look nicer, and to support the global Chaosnet, let's introduce two logical names in ITS' hosts file.  Something like `local-time-server` and `global-time-server`.  The former would be 177001, and the latter maybe Router.Chaosnet.NET.

bictorv:
Well, if you are NOT on the private subnet, trying to contact 177001 will take time to fail. For the non-private case, TIME.Chaosnet.NET already exists (but it might be better to use the local cbridge). (Hmm, didn't we discuss this ages ago?)

Perhaps the best would be to have a "time-server" entry in the ITS hosts file, which points to 177001 initially but with a comment about how to change it after connecting to the global net? The NETIME could have the same source code for everyone, perhaps with a clear comment that it's a good idea to add your local cbridge to the list?

Side note: ITS (at least under klh10) seem to have a slow clock (and thus be poor time servers), and for some reason TT currently is 6 hours off. (Test by "bhostat.py -s time -1"). Otherwise one option would be to make a BRD connection to ask for the time and simply use the first response.

larsbrinkhoff:
I think NETIME asks all servers in parallel.

bictorv:
OK, then perhaps "local-time-server" (with instructions in the hosts file that it's a good idea to update to your new local cbridge address when you attach to the global net) plus TIME.Chaosnet.NET?

larsbrinkhoff:
We're also close to getting a remote tape server, see https://github.com/bictorv/chaosnet-bridge/pull/31  So it would be good to add a name for that too, but "local-tape-server" is a bit much to type.  That could be the primary name, but a shorter alias would be good.

eswenson1:
There are issues with TIMES and NETIME.  First of all, even with TZONE set (in my case to PT), when I run TIMES, I get this:

```
times↑K!
Checking the time using CHAOS protocol: (^S to stop):
[...]

PR #2225, opened 2023-09-24
Updated netime and times to handle daily savings time.

eswenson1:
Also updated list of hosts so that these can be used today.

Addresses #2213


larsbrinkhoff (approved):
I split it into two commits.

src/sysen1/netime.128
```diff
@@ -100,45 +100,80 @@ $forma:	save a
...
 hosts:
 
+; Ask the reliable machines first.  They are all on the global
+; Chaosnet and tend to answer quickly.
+
+	[asciz "LOCAL-TIME-SERVER"]
+	[asciz "TIME.CHAOSNET.NET"]
+	[asciz "CHAOS.SWENSON.ORG"]
+	[asciz "CHAOS2.SWENSON.ORG"]
+	[asciz "ROUTER.CHAOSNET.NET"]
```
bictorv, line 110: Remove ROUTER since it is the same machine as TIME.

src/sysen1/netime.128
```diff
@@ -100,45 +100,80 @@ $forma:	save a
...
+	<(sixbit /ES/)>,,[asciz "ES"]
```
bictorv, line 111: Emulated ITS machines are poor time servers (in particular TT ;-)) so comment them out here.

src/sysen1/netime.128
```diff
@@ -150,7 +185,7 @@ wait:	20.*60.			; Wait 20 seconds polling other hosts.
 numerq:	2			; Ceiling of 2/3 of the hosts that respond
 denomq:	3			; must agree on the time.
 
-minq:	4			; Insist on at least 4 hosts with times in
+minq:	3			; Insist on at least 4 hosts with times in
```
bictorv, line 188: Make the comment match the value. But really, shouldn't this be a function of the length of the list of servers?

src/sysen1/netime.128
larsbrinkhoff: Right, I declare TT to be generally unreliable for anything.  It means Temporary Testing, after all.

src/sysen1/netime.128
```diff
@@ -100,43 +100,58 @@ $forma:	save a
...
+
+	[asciz "LOCAL-TIME-SERVER"]
+	[asciz "TIME.CHAOSNET.NET"]
+	[asciz "CHAOS.SWENSON.ORG"]
+	[asciz "CHAOS2.SWENSON.ORG"]
+	[asciz "ROUTER.CHAOSNET.NET"]
+	<(sixbit /ES/)>,,[asciz "ES"]
+	<(sixbit /EX/)>,,[asciz "EX"]
+	<(sixbit /KA/)>,,[asciz "KA"]
+	<(sixbit /KL/)>,,[asciz "KL"]
```
larsbrinkhoff, line 114: KA and KL aren't globally known hosts (along with DB).

src/sysnet/times.74
```diff
@@ -425,16 +454,16 @@ NHOSTS==100.			   ;Maximum number of hosts we can query.
...
-	[ASCIZ /CP/]
-	[ASCIZ /PI/]
-	[ASCIZ /DX/]
-	[ASCIZ /DU/]
-	[ASCIZ /KN/]
-	[ASCIZ /MD/]
-	[ASCIZ /NX/]
-	[ASCIZ /AXPKLH/]
-	[ASCIZ /PCKLH/]
+TCPTAB:	[ASCIZ /ES/]
```
bictorv, line 457: Emulated ITS systems are still poor time servers, but it's not as bad here since TIMES doesn't set the local time (it doesn't, right?)

src/sysnet/times.74
```diff
@@ -462,12 +491,23 @@ TCPTBL==.-TCPTAB
...
-	[ASCIZ /DX/]
-	[ASCIZ /DU/]
-	[ASCIZ /KN/]
-	[ASCIZ /MD/]
-	[ASCIZ /NX/]
+	[asciz "LOCAL-TIME-SERVER"]
+	[asciz "TIME.CHAOSNET.NET"]
+	[asciz "CHAOS.SWENSON.ORG"]
+	[asciz "CHAOS2.SWENSON.ORG"]
+	[asciz "ROUTER.CHAOSNET.NET"]
```
bictorv, line 498: Remove ROUTER, as above.

src/sysnet/times.74
bictorv: Plus I don't get an answer from ES. I know UP doesn't have a TCP TIME server currently.

src/sysnet/times.74
```diff
@@ -511,7 +569,7 @@ PDL:	BLOCK PDLLEN	;The Stack.
 	0
 
 DEBUG:	0		;-1 => Debugging.
-TCPP:	-1		;-1 => Try TCP time servers.
+TCPP:	0		;-1 => Try TCP time servers.
```
bictorv, line 572: This makes ITS try NCP (see ARTIM), which I don't think works. Either TCPP or CHAOSP should be -1, right?

src/sysen1/netime.128
larsbrinkhoff: LOCAL-TIME-SERVER isn't in the H3TEXT file.  Maybe do it in #2226?

bictorv: Or add it in this commit?

bictorv (changes requested):
Please look at the comments I made.

eswenson1:
Right! I forgot about doing just that. Thanks.

src/sysen1/netime.128
eswenson1: I stupidly already committed after seeing Lars’ approval and before getting through my email. I’ll add another PR removing router.chaosnet.org and adding local-time-server to h3text.

src/sysnet/times.74
eswenson1: ES will answer via chaosnet, currently not via TCP because my firewall is blocking whatever port is used. I can open it up.

src/sysnet/times.74
eswenson1: Will do.

eswenson1:
@bictorv @larsbrinkhoff Since I merged this already before seeing these comments and just your approval, I’ll address all these comments in a new PR. To be done soon. Ok?

src/sysen1/netime.128
eswenson1: I am going to leave ES, EX, and UP.  At least ES and EX are getting time updates and should have good clocks. I don't know about UP.  However, it is good to have a few more hosts. It doesn't slow things down much.  In fact, when I've been running this during test, all my systems seem to get their times from EX rather than the bridges.  Of course, EX's time is also as good as the bridges.

src/sysen1/netime.128
eswenson1: If the number of servers is less than minq, then minq is adjusted to be that number.  So I think it is fine.  However, I wasn't reliably able to get times when I used minq with the list of hosts. So I think 3 is good. I'm fixing the comment.

src/sysnet/times.74
eswenson1: Yes. I didn't realize that using the JNAME of CTIMES automatically set CHAOSP, so leaving TCPP set to -1 is the right thing. I'll fix.

src/sysen1/netime.128
bictorv: Well, I guess it depends on the emulator and the uptime. Currently, UP and BV have been up a couple of weeks and are over 2 minutes ahead. I recall seeing your hosts also being ahead (the easiest way is "bhostat.py -s time -1").

But then again, UP and BV don't have NETIME configured to work, they pick up the time at boot and start drifting.

eswenson1: Well, NETIME has a way of voting to weed out host times that are off. So if you do have time.chaosnet.net and local-host-time responding, in addition to UP with a way-off clock, NETIME won’t use that time.

However, if you can’t get a time from more reliable sources, sometimes it is better to get some (inaccurate) time.  In my case, when you boot my KS instance, pdp10-is cannot provide a clock value to ITS, so without some time from NETIME, it would not get any time set. I think it is better to get a drifted clock than no clock.

PR #2226, opened 2023-09-25
Add TIME.Chaosnet.NET as alias of MX12.Victor.SE

bictorv:


larsbrinkhoff:
Maybe add LOCAL-TIME-SERVER too, when you're at it?

larsbrinkhoff:
Or was it just LOCAL-TIME?

Issue #2231, opened 2023-10-04
MLDEV doesn't work with 3-character host names


eswenson1:
Using MLDEV with "devices" like EXA (three-character hostname), EXADIR, DIREXA, XEXA (experimental) don't work. Also, it is not possible to invoke a remote SHOE device to a three-character hostname (e.g. EXASHOE -- because that would be greater than 6 characters).

larsbrinkhoff:
There is no need to post an issue together with a pull request.  Just the pull request is enough.

larsbrinkhoff:
Thanks!  Searching applies to both pull requests and issues, so things show up doubly, or you have to look at both the issue *and* the pull request to get the full picture.  So it's less work for everyone if there's just a pull request.  Of course, if there's no quick fix there will probably be an issue (hopefully)  followed by a pull request later.

PR #2232, opened 2023-10-04
Updated MLDEV device to handle 2- and 3-char ITS hostnames.

eswenson1:
Addresses #2231.

Also a prior change to support the ITS ITSNMS table broke some of the case handling in MLDEV. This PR fixes those as well.

### DIFF
--- a/src/sysen2/mldev.108
+++ b/src/sysen2/mldev.108
@@ -129,11 +129,9 @@ nox:				; jump here if hostname minus first "x" is not a known ITS
 				; is it hhhdir or hhh
 	hrlz c,a		; get device in c
 	jumpe c, host3		; jump if no device
-	jrst hst2ck		;  have a device, nope, check for hhddd case
 
-hst2ck:				; check for hhddd case
 	ldb c,[000600,,a]	; can only be hhddd if last sixbit character is null
-	jumpn c,hhddd1		; handle hhddd case if last character is NUL
+	jumpe c,hhddd1		; handle hhddd case if last character is NUL
 	
 	hrlz c,a		; check for hhhddd, get ddd in c
 	jumpn e,hhhddd		; is there a ddd?, if so handle it
@@ -176,16 +174,16 @@ notits:				; here we've determined that hostname is invalid, check for special d
 	CAIN C,'XGP	;XGP THROUGH THE NET TO AI FROM ML/MC/DM
 	 JRST [	MOVEI A,'AI
 		JRST GOTHSX]
-	CAIN A,'GLP		;GOULD THROUGH THE NET TO MX
+	CAIN C,'GLP		;GOULD THROUGH THE NET TO MX
 	 JRST [ MOVEI A,'MX
 		JRST GOTHSX]
-	CAIN A,'DVR	;DVR THROUGH THE NET TO MC FROM AI/ML/DM
+	CAIN C,'DVR	;DVR THROUGH THE NET TO MC FROM AI/ML/DM
 	 JRST [ MOVEI A,'MC
 		JRST GOTHSX]
-	CAIN A,'DVS	;DVS THROUGH THE NET TO MC FROM ML/DM
+	CAIN C,'DVS	;DVS THROUGH THE NET TO MC FROM ML/DM
 	 JRST [	MOVEI A,'MC
 		JRST GOTHSX]
-	CAIN A,'TPL	;TPL THROUGH THE NETWORK FOR MC
+	CAIN C,'TPL	;TPL THROUGH THE NETWORK FOR MC
 	 JRST [ LDB B,[000400,,JBCOP]
 		CAIN B,2
 		 JRST TPLLK	;TRYING TO LINK ON TPL: LOSES
@@ -210,8 +208,9 @@ hhhddd:				; get device from last 3 characters
 doshoe:				; device is SHOE
 	lsh c,12.		; left justfify device
 	movem c,jbcdev		; set remote device
-	ldb b,[301400,,a]	; get 2-character hostname
-	lsh b,30		; left justify hostname
+	ldb a,[301400,,a]	; get 2-character hostname
+	lsh a,30		; left justify hostname
+	jrst gothst		; process host
 
 dodir:	hrlz b,a		;a is for example 'DIREX', or 'DIREXA' ; b becomes 'EX' or 'EXA'
 	hrlzm c,jbcdev		; save left-justified 'DIR' as remote device
--- a/src/sysen2/mldev.108
+++ b/src/sysen2/mldev.108
@@ -55,7 +55,6 @@ FILBLN:	0	;FILE LENGTH, IN BYTES OF SIZE WRITTEN IN.
 FILBSZ:	0	;BYTE SIZE FILE WRITTEN IN.
 FILDTP:	0	;-1 => WE KNOW THE CREATION DATE OF THIS FILE.
 FILDAT:	0	;FILE CREATION DATE (IF KNOWN).
-3char:  0	;-1 if hostname is 3 characters long, else 2 characters long
 dirdev: sixbit /DIR/	; DIR device
 dskdev: sixbit /DSK/	; DSK device
 
@@ -99,7 +98,7 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
 	 .value
 	;; check for 4-character devices. Only one checked for is SHOE, which only works with 2-character
 	;;  hostnames.
-	ldb c,[003000,,a]	; get last four characters
+	ldb c,[003000,,a]		; get last four characters
 	camn c,[sixbit /  SHOE/]	; is it the SHOE device?
 	  jrst doshoe
 
@@ -112,77 +111,67 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
 	ldb c,[.bp (770000),a]
 	cain c,'X
 	 jrst [	move b,a
-		lsh b,6              ;get rid of X
-		pushj p,isitsp       ; see if we now have a known ITS name
-		 jrst nox            ;  nope, skip handling as experimental case
-		move c,dskdev        ; assume DSK device
-		movem c,jbcdev       ; store as remote device
-		move a,b	         ; get host in A for gothst
-		jrst gothst ]        ; and go process request for this host
+		lsh b,6		;get rid of X
+		pushj p,isitsp	; see if we now have a known ITS name
+		 jrst nox	;  nope, skip handling as experimental case
+		move c,dskdev	; assume DSK device
+		movem c,jbcdev	; store as remote device
+		move a,b	; get host in A for gothst
+		jrst gothst ]	; and go process request for this host
 
-nox:                         ; jump here if hostname minus first "x" is not a known ITS
-                             ; check to see if format is hhDIR
+nox:				; jump here if hostname minus first "x" is not a known ITS
+				; check to see if format is hhDIR
 	move c,a
-	lsh c,12.                ; move past hostname
-	camn c,dirdev            ; is what remains the DIR device?
-	  jrst dirchk            ;  yes, continue checking hhDIR case
+	lsh c,12.		; move past hostname
+	camn c,dirdev		; is what remains the DIR device?
+	  jrst dirchk		;  yes, continue checking hhDIR case
 
-                             ; is it hhhdir or hhh
-	hrlz c,a
-	jumpe c, host3		     ; jump if no device
-	  jrst hst2ck            ;  have a device, nope, check for hhddd case
-	movem c,jbcdev           ; we have a device; save it as remote device
-	jrst chkits              ; see if host part (hhh) is a known ITS machine
+				; is it hhhdir or hhh
+	hrlz c,a		; get device in c
+	jumpe c, host3		; jump if no device
+	jrst hst2ck		;  have a device, nope, check for hhddd case
 
-hst2ck:	                     ; check for hhddd case
-	ldb c,[000600,,a]        ; can only be hhddd if last sixbit character is null
-	cain c,0                 ; is last character null?
-	  jrst hhddd1            ;  yes, handle hhddd case
+hst2ck:				; check for hhddd case
+	ldb c,[000600,,a]	; can only be hhddd if last sixbit character is null
+	jumpn c,hhddd1		; handle hhddd case if last character is NUL
 	
-	move c,a                 ; check for hhhddd
-	lsh c,18.                ; get ddd in c
-	caie c,0                 ; is there a ddd?
-	  jrst hhhddd            ;  yes, handle it
-	jrst chkits              ; nope, assume that we just have hhh
+	hrlz c,a		; check for hhhddd, get ddd in c
+	jumpn e,hhhddd		; is there a ddd?, if so handle it
+	jrst chkits		; nope, assume that we just have hhh
 
-host3:                       ; seems like hhh, but may be special, check to see if ITS
-	move b,a                 ; hostname in A
-	pushj p,isitsp           ; is it a known ITS
-	  jrst notits            ;  nope, see if special device
-	move c,dskdev	         ; we have an ITS name, so assume DSK device
-	movem c,jbcdev           ; save remote device
-	jrst gothst              ; and process our host
+host3:				; seems like hhh, but may be special, check to see if ITS
+	move b,a		; hostname in A
+	pushj p,isitsp		; is it a known ITS
+	  jrst notits		;  nope, see if special device
+	move c,dskdev		; we have an ITS name, so assume DSK device
+	movem c,jbcdev		; save remote device
+	jrst gothst		; and process our host
 
-dirchk:	                     ; 2-char hostname with DIR device
-	ldb b,[.bp (777700),a]   ; get hh part in b
-	lsh b,24.                ; left justify hostname
-dirch2:	move c,dirdev        ; set remote device as DIR device
+dirchk:				; 2-char hostname with DIR device
+	ldb b,[.bp (777700),a]	; get hh part in b
+	lsh b,24.		; left justify hostname
+dirch2:	move c,dirdev		; set remote device as DIR device
 	movem c,jbcdev
-	jrst chkhst              ;and check our host
-
-dirch3:	                     ; 3-char hostname with DIR device
-	ldb b,[222200,,a]	     ; get hhh part in b
-	lsh b,18.                ; left-justify hostname
-	jrst dirch2              ; set remote device as DIR device and process host
+	move a,b		; get host in a
+	jrst gothst		; process host
 
 hhddd2: ldb a,[301400,,jbcdev]	; get hostname in A
-	lsh a,24.		            ; left justify hostname
-	movem c,jbcdev		        ; set remote device from C
-	jrst gothst                 ; process host
+	lsh a,24.		; left justify hostname
+	movem c,jbcdev		; set remote device from C
+	jrst gothst		; process host
 
-hhddd1:                      ; continue check for hhddd
-	ldb c,[003000,,a]        ; get device portion
-	lsh c,12.                ; left-justify
-	caie c,0                 ; no device?
-	  jrst hhddd2            ;   nope, have devie; get hostname and set remote device
+hhddd1:				; continue check for hhddd
+	ldb c,[003000,,a]	; get device portion
+	lsh c,12.		; left-justify
+	jumpn c,hhddd2		; jump if we have a device; get hostname and set remote device
 
-	; fall through; assume just an ITS name
+	;; fall through; assume just an ITS name
 chkits:	move b,a
-	pushj p,isitsp           ; is b an ITS name?
-	 jrst notits             ;  no, process as special device
-    jrst gothst              ; yes, process this host
+	pushj p,isitsp		; is b an ITS name?
+	 jrst notits		;  no, process as special device
+        jrst gothst		; yes, process this host
 
-notits:	                     ; here we've determined that hostname is invalid, check for special device
+notits:				; here we've determined that hostname is invalid, check for special device
 	MOVS C,JBCDEV
 	CAIN C,'XGP	;XGP THROUGH THE NET TO AI FROM ML/MC/DM
 	 JRST [	MOVEI A,'AI
@@ -212,57 +201,39 @@ notits:	                     ; here we've determined that hostname is invalid, c
 NSD:	MOVSI A,%ENSDV
 	JRST NOGO2
 
-hhhddd:                 ; get device from last 3 characters
-	ldb c,[002200,,a]   ; get device
-	hrlz c,c            ; left-jusify
-	movem c,jbcdev      ; save remote device
-	ldb b,[222200,,a]   ; get hostname
-	hrlz a,b            ; left-justify
-	jrst gothst         ; process host
+hhhddd:				; get device from last 3 characters
+	hrlz c,a		; get device and left justify
+	movem c,jbcdev		; save remote device
+	hllzs a			; get hostname
+	jrst gothst		; process host
 
-doshoe:                 ; device is SHOE
-	lsh c,12.	        ; left justfify device
+doshoe:				; device is SHOE
+	lsh c,12.		; left justfify device
 	movem c,jbcdev		; set remote device
 	ldb b,[301400,,a]	; get 2-character hostname
-	lsh b,30		    ; left justify hostname
+	lsh b,30		; left justify hostname
 
-chkhst:                 ; b has host
-	pushj p,isitsp      ; is it a known ITS machine
-	  jrst .+1          ; ???
-	move a,b	        ; get host in a
-	jrst gothst         ; process host
-
-dodir:	hrlz b,a	    ;a is for example 'DIREX', or 'DIREXA' ; b becomes 'EX' or 'EXA'
-	pushj p,isitsp	    ;is it an ITS host that we know about
-	  jrst .+1          ; ???
-	lsh c,18.		    ; left-justift 'DIR'
-	movem c,jbcdev	    ; save remote device
-	move a,b	        ; get host in A for gothst
-	jrst gothst         ; process host
+dodir:	hrlz b,a		;a is for example 'DIREX', or 'DIREXA' ; b becomes 'EX' or 'EXA'
+	hrlzm c,jbcdev		; save left-justified 'DIR' as remote device
+	move a,b		; get host in A for gothst
+	jrst gothst		; process host
 
 ;; Check if B has the name of an ITS, if so, skip return
 ;; Messes up T
 isitsp:	movsi t,-nitses
 isitz0:	camn b,itses(t)
-	  jrst popj1	   ;match, skip return
+	  jrst popj1		;match, skip return
 	skipe itses(t)
 	aobjn t,isitz0
 	popj p,
 
-gothsx: ; a contains sixbit 2-char host, right-justified
+gothsx:	; a contains sixbit 2-char host, right-justified
 	lsh a,24.		; left-justify host
-	lsh c,18.       ; left justify device
-	movem c,jbcdev  ; save remote device
+	hrlzm c,jbcdev		; save left-justified remote device
 	;; fall through
 
 	; a contains sixbit host, left-justified
-GOTHST:	;; determine whether this is a 2-char or 3-char hostname
-	ldb d,[220600,,a]        ; check third character for nul
-	jumpe d,.+3              ; jump if 2-char host
-    setom 3char              ; mark as 3-char host
-	skipa                    ; skip
-goths0:	setzm 3char          ; mark as 3-char host
-	MOVE B,[440700,,BUF]		;Make ASCIZ host name
+GOTHST:	MOVE B,[440700,,BUF]	;Make ASCIZ host name
 GOTHS1:	LDB T,[360600,,A]
 	JUMPE T,GOTHS2
 	ADDI T,40
--- a/src/sysen2/mldev.108
+++ b/src/sysen2/mldev.108
@@ -55,6 +55,9 @@ FILBLN:	0	;FILE LENGTH, IN BYTES OF SIZE WRITTEN IN.
 FILBSZ:	0	;BYTE SIZE FILE WRITTEN IN.
 FILDTP:	0	;-1 => WE KNOW THE CREATION DATE OF THIS FILE.
 FILDAT:	0	;FILE CREATION DATE (IF KNOWN).
+3char:  0	;-1 if hostname is 3 characters long, else 2 characters long
+dirdev: sixbit /DIR/	; DIR device
+dskdev: sixbit /DSK/	; DSK device
 
 nitses==:40
 itses: block nitses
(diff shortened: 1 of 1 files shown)


src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
 	cain c,'X
 	 jrst [	move b,a
-		lsh b,6		;get rid of X
-		pushj p,isitsp
-		 jrst .+1
-		movem b,jbcdev
-		move a,b	;get host in A for gothst
-		jrst gothst ]
-	;; is it xxDIR?
+		lsh b,6              ;get rid of X
```
larsbrinkhoff, line 115: Not overly important, but the convention is to place tabs before the comment, not spaces.

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
-	 jrst notits
-	  jrst gothst
+	lsh c,12.                ; move past hostname
+	camn c,dirdev            ; is what remains the DIR device?
+	  jrst dirchk            ;  yes, continue checking hhDIR case
+
+                             ; is it hhhdir or hhh
+	hrlz c,a
+	jumpe c, host3		     ; jump if no device
+	  jrst hst2ck            ;  have a device, nope, check for hhddd case
```
larsbrinkhoff, line 133: JUMPE isn't a a skip instruction, so JRST shouldn't be indented.

larsbrinkhoff:
This is a big change which is difficult to understand fully.  I'm posting a few pro-forma trite remarks.

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	movem c,jbcdev           ; we have a device; save it as remote device
+	jrst chkits              ; see if host part (hhh) is a known ITS machine
+
+hst2ck:	                     ; check for hhddd case
+	ldb c,[000600,,a]        ; can only be hhddd if last sixbit character is null
+	cain c,0                 ; is last character null?
+	  jrst hhddd1            ;  yes, handle hhddd case
```
larsbrinkhoff, line 140: Suggest replacing two instructions with `jumpe c,hhddd1`.

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
 	
-notits:	;; nothing came of it, fall through
-	MOVS A,JBCDEV
-	CAIN A,'XGP	;XGP THROUGH THE NET TO AI FROM ML/MC/DM
-	 JRST [	MOVS A,[SIXBIT/AIXGP/]
-		MOVSM A,JBCDEV
-		MOVEI A,'AI
-		JRST GOTHST]
+	move c,a                 ; check for hhhddd
+	lsh c,18.                ; get ddd in c
```
larsbrinkhoff, line 143: Two instructions → `hrlz c,a`

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	caie c,0                 ; is there a ddd?
+	  jrst hhhddd            ;  yes, handle it
```
larsbrinkhoff, line 145: Two instructions → `jumpn e,hhhddd`

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+dirchk:	                     ; 2-char hostname with DIR device
+	ldb b,[.bp (777700),a]   ; get hh part in b
+	lsh b,24.                ; left justify hostname
+dirch2:	move c,dirdev        ; set remote device as DIR device
+	movem c,jbcdev
+	jrst chkhst              ;and check our host
+
+dirch3:	                     ; 3-char hostname with DIR device
+	ldb b,[222200,,a]	     ; get hhh part in b
+	lsh b,18.                ; left-justify hostname
```
larsbrinkhoff, line 165: Two instructions → `hrlz b,a`

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	ldb c,[003000,,a]        ; get device portion
+	lsh c,12.                ; left-justify
+	caie c,0                 ; no device?
+	  jrst hhddd2            ;   nope, have devie; get hostname and set remote device
+
+	; fall through; assume just an ITS name
+chkits:	move b,a
+	pushj p,isitsp           ; is b an ITS name?
+	 jrst notits             ;  no, process as special device
+    jrst gothst              ; yes, process this host
```
larsbrinkhoff, line 183: Tab before jrst, please.

src/sysen2/mldev.108
```diff
@@ -166,11 +207,39 @@ notits:	;; nothing came of it, fall through
...
 		MOVEI A,'ML		;USE THE MATHLAB LINE PRINTER
-		JRST GOTHST ]
+		JRST GOTHSX ]
 ;HOST NOT RECOGNIZED
 NSD:	MOVSI A,%ENSDV
 	JRST NOGO2
 
+hhhddd:                 ; get device from last 3 characters
+	ldb c,[002200,,a]   ; get device
+	hrlz c,c            ; left-jusify
```
larsbrinkhoff, line 217: Two instructions → `hrlz c,a`

src/sysen2/mldev.108
```diff
@@ -166,11 +207,39 @@ notits:	;; nothing came of it, fall through
...
+	movem c,jbcdev      ; save remote device
+	ldb b,[222200,,a]   ; get hostname
+	hrlz a,b            ; left-justify
```
larsbrinkhoff, line 220: Two instructions → `hllzs a`  (I chedked that `b` isn't an input to `gothst`)

src/sysen2/mldev.108
```diff
@@ -180,8 +249,20 @@ isitz0:	camn b,itses(t)
...
+	lsh a,24.		; left-justify host
+	lsh c,18.       ; left justify device
+	movem c,jbcdev  ; save remote device
+	;; fall through
+
+	; a contains sixbit host, left-justified
+GOTHST:	;; determine whether this is a 2-char or 3-char hostname
+	ldb d,[220600,,a]        ; check third character for nul
+	jumpe d,.+3              ; jump if 2-char host
+    setom 3char              ; mark as 3-char host
```
larsbrinkhoff, line 262: Tab before `setom`, please.

src/sysen2/mldev.108
```diff
@@ -166,11 +207,39 @@ notits:	;; nothing came of it, fall through
...
+	pushj p,isitsp      ; is it a known ITS machine
+	  jrst .+1          ; ???
+	move a,b	        ; get host in a
+	jrst gothst         ; process host
+
+dodir:	hrlz b,a	    ;a is for example 'DIREX', or 'DIREXA' ; b becomes 'EX' or 'EXA'
+	pushj p,isitsp	    ;is it an ITS host that we know about
+	  jrst .+1          ; ???
+	lsh c,18.		    ; left-justift 'DIR'
+	movem c,jbcdev	    ; save remote device
```
larsbrinkhoff, line 239: Two instructions → `hrlzm c,jbcdev`

src/sysen2/mldev.108
```diff
@@ -180,8 +249,20 @@ isitz0:	camn b,itses(t)
 	aobjn t,isitz0
 	popj p,
 
-;A CONTAINS SIXBIT HOST, RIGHT-JUSTIFIED
-GOTHST:	MOVE B,[440700,,BUF]		;Make ASCIZ host name
+gothsx: ; a contains sixbit 2-char host, right-justified
+	lsh a,24.		; left-justify host
+	lsh c,18.       ; left justify device
+	movem c,jbcdev  ; save remote device
```
larsbrinkhoff, line 255: Two instructions → `hrlzm c,jbcdev`

src/sysen2/mldev.108
```diff
@@ -55,6 +55,9 @@ FILBLN:	0	;FILE LENGTH, IN BYTES OF SIZE WRITTEN IN.
 FILBSZ:	0	;BYTE SIZE FILE WRITTEN IN.
 FILDTP:	0	;-1 => WE KNOW THE CREATION DATE OF THIS FILE.
 FILDAT:	0	;FILE CREATION DATE (IF KNOWN).
+3char:  0	;-1 if hostname is 3 characters long, else 2 characters long
```
larsbrinkhoff, line 58: I see `3char` is set to 0 or -1, but I don't see that the value is checked later.

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	lsh c,12.                ; move past hostname
+	camn c,dirdev            ; is what remains the DIR device?
+	  jrst dirchk            ;  yes, continue checking hhDIR case
+
+                             ; is it hhhdir or hhh
+	hrlz c,a
+	jumpe c, host3		     ; jump if no device
+	  jrst hst2ck            ;  have a device, nope, check for hhddd case
+	movem c,jbcdev           ; we have a device; save it as remote device
+	jrst chkits              ; see if host part (hhh) is a known ITS machine
```
larsbrinkhoff, line 135: These two instructions are unreachable.

src/sysen2/mldev.108
```diff
@@ -94,69 +97,107 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	jrst gothst              ; and process our host
+
+dirchk:	                     ; 2-char hostname with DIR device
+	ldb b,[.bp (777700),a]   ; get hh part in b
+	lsh b,24.                ; left justify hostname
+dirch2:	move c,dirdev        ; set remote device as DIR device
+	movem c,jbcdev
+	jrst chkhst              ;and check our host
+
+dirch3:	                     ; 3-char hostname with DIR device
```
larsbrinkhoff, line 163: I don't see that the `dirch3` label is jumped to anywhere.

src/sysen2/mldev.108
```diff
@@ -180,8 +249,20 @@ isitz0:	camn b,itses(t)
...
+	movem c,jbcdev  ; save remote device
+	;; fall through
+
+	; a contains sixbit host, left-justified
+GOTHST:	;; determine whether this is a 2-char or 3-char hostname
+	ldb d,[220600,,a]        ; check third character for nul
+	jumpe d,.+3              ; jump if 2-char host
+    setom 3char              ; mark as 3-char host
+	skipa                    ; skip
+goths0:	setzm 3char          ; mark as 3-char host
```
larsbrinkhoff, line 264: The `goths0` label isn't used anywhere.

src/sysen2/mldev.108
```diff
@@ -166,11 +207,39 @@ notits:	;; nothing came of it, fall through
...
+
+doshoe:                 ; device is SHOE
+	lsh c,12.	        ; left justfify device
+	movem c,jbcdev		; set remote device
+	ldb b,[301400,,a]	; get 2-character hostname
+	lsh b,30		    ; left justify hostname
+
+chkhst:                 ; b has host
+	pushj p,isitsp      ; is it a known ITS machine
+	  jrst .+1          ; ???
```
larsbrinkhoff, line 231: No need to call `isitsp` if the result isn't used.

src/sysen2/mldev.108
```diff
@@ -166,11 +207,39 @@ notits:	;; nothing came of it, fall through
...
+	move a,b	        ; get host in a
+	jrst gothst         ; process host
+
+dodir:	hrlz b,a	    ;a is for example 'DIREX', or 'DIREXA' ; b becomes 'EX' or 'EXA'
+	pushj p,isitsp	    ;is it an ITS host that we know about
+	  jrst .+1          ; ???
```
larsbrinkhoff, line 237: No need to call `isitsp` if the result isn't used.

larsbrinkhoff:
I order to understand and better review the change, I jotted down some notes about what checks and actions happen:

1. Is it *host*+`SHOE`?  Then host=*host*, device=`SHOE`.
2. Is it `DIR`+*host*?  Then host=*host*, device=`DIR`.
3. Is it `X`+*host*?  Then host=*host*, device=`DSK`.
4. Is it *hh*+`DIR`?  Then host=*hh*, device=`DIR`.
5. Is it *hhh* and a known host?  Then host=*hhh*, device=`DSK`.
6. Is it *ddd* and one of a plethora of special devices?  Then host=*special*, device=*special*.
7. Is it *hh*+*device*?  Then host=*hh*, device=*device*.
8. Is it *hhh*+*device*?  Then host=*hhh*, device=*device*.
9. Else it must be just host=*host*.  Not sure device is set?

larsbrinkhoff:
I notice there are three routines involved in checking a host name.

1. `isitsp` runs through the internal list of known ITS hosts.
2. `chkhst` calls `isitsp` **but ignores the result** and then calls `gothst`.
3. `gothst` sets a flag whether the host name is two or three characters **but the flag is unused**.  It then looks up the hostname in the hosts table.

It seems to me `chkhst` is somewhat unnecessary and we could probably call `gothst` directly.  Some other places call `isitsp` for good reason, but having `chkhst` to call `isitsp` (whether it ignores the result or not) and then `gothst` looking up the hostname seems like the two checks mostly overlap.

bictorv:
> I order to understand and better review the change, I jotted down some notes about what checks and actions happen:

I did something similar but from scratch, and coded it up. It looks much simpler in my mind.
0. Check for xxSHOE
1. If RH is empty: Check for HH, HHH, or xHH
2. If len = 5, check for HHDDD or DDDHH
3. else check HHHDDD or DDDHHH

See if you agree? Note: I haven't even compiled the code.
[mldev.108.txt](https://github.com/PDP-10/its/files/12818362/mldev.108.txt)

src/sysen2/mldev.108
eswenson1: Crap. This is what I get for editing in emacs on host — I have the option set to convert tabs to spaces. I’ll address this and the other comments you provided. Thanks.

src/sysen2/mldev.108
eswenson1: The old code did in a bunch of places. Didn’t really know why. Left it.

src/sysen2/mldev.108
```diff
@@ -112,77 +111,67 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	  jrst dirchk		;  yes, continue checking hhDIR case
+
+				; is it hhhdir or hhh
+	hrlz c,a		; get device in c
+	jumpe c, host3		; jump if no device
+	jrst hst2ck		;  have a device, nope, check for hhddd case
+
+hst2ck:				; check for hhddd case
+	ldb c,[000600,,a]	; can only be hhddd if last sixbit character is null
+	jumpn c,hhddd1		; handle hhddd case if last character is NUL
```
larsbrinkhoff, line 136: Did I say `jumpn` here?  Sorry if so.  It should be `jumpe` right?

eswenson1:
> I order to understand and better review the change, I jotted down some notes about what checks and actions happen:
> 
> 1. Is it _host_+`SHOE`?  Then host=_host_, device=`SHOE`.
> 2. Is it `DIR`+_host_?  Then host=_host_, device=`DIR`.
> 3. Is it `X`+_host_?  Then host=_host_, device=`DSK`.
> 4. Is it _hh_+`DIR`?  Then host=_hh_, device=`DIR`.
> 5. Is it _hhh_ and a known host?  Then host=_hhh_, device=`DSK`.
> 6. Is it _ddd_ and one of a plethora of special devices?  Then host=_special_, device=_special_.
> 7. Is it _hh_+_device_?  Then host=_hh_, device=_device_.
> 8. Is it _hhh_+_device_?  Then host=_hhh_, device=_device_.
> 9. Else it must be just host=_host_.  Not sure device is set?

This looks right. You may be right that the fallback case at the end doesn’t set device. I’ll check on this. I wonder if we ever actually reach that case?

eswenson1:
> I notice there are three routines involved in checking a host name.
> 
> 1. `isitsp` runs through the internal list of known ITS hosts.
> 2. `chkhst` calls `isitsp` **but ignores the result** and then calls `gothst`.
> 3. `gothst` sets a flag whether the host name is two or three characters **but the flag is unused**.  It then looks up the hostname in the hosts table.
> 
> It seems to me `chkhst` is somewhat unnecessary and we could probably call `gothst` directly. Some other places call `isitsp` for good reason, but having `chkhst` to call `isitsp` (whether it ignores the result or not) and then `gothst` looking up the hostname seems like the two checks mostly overlap.

I got rid of chkhst — in an earlier version it did more, but I whittled it down and you are right, it is not useful. The calls to isitsp where the error return just proceeds were in the original code. I left them there but after your review, I got rid of them since, as you pointed out, they are really a noop.

src/sysen2/mldev.108
eswenson1: I don’t recall what you said, but I tested all use cases and they worked. I think. I’ll check again and look at that code.

eswenson1: Yes, should be a jumpe.  I wonder why I was able to test the EXSHO case, and have it work?  I'll step through that again and see why the JUMPN seemed to work just fine.  May be a place for more optimization.  

There is some extra case handling that could probably be removed now -- the old code handled the hhhDIR case specially, but now the more general hhhDDD handling shouid cover that case.  Same with the hhDIR versus hhDDD cases.  These were separate before and I left them separate.  May optimize this later.

eswenson1: I tested the code again and the HHDDD case (e.g. EXSHO) case failed due to the JUMPN (which should have been a JUMPE).  I realized that I had tested this case out using a slightly older version of the ATSIGN MLDEV binary.  I'm running through all test cases again.  I also got rid of the hst2ck label, since it wasn't used.

eswenson1:
I updated this PR again to fix a few bugs in MLDEV introduced by last update.

Tested these cases: hhSHOE, DIRhh, DIRhhh, hhDIR, hhhDIR, Xhh, Xhhh, hhDDD, hhhDDD, xgp, dvr, dvs, tpl, glp, and xxx, where xxx is an unknown (bad) device.

larsbrinkhoff:
Seems to cover most cases!
Screenshot at 2023-10-05 21-02-24

larsbrinkhoff (approved):
Looks good to me.

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
-		pushj p,isitsp
-		 jrst .+1
-		move a,b	; get host in A
-		jrst gothst ]		
-	move b,a
-	;; is b an ITS name?
-	pushj p,isitsp
-	 jrst notits
-	  jrst gothst
+	lsh c,12.		; move past hostname
```
bictorv, line 125: This only handles HHDIR, not HHHDIR. Why is DIR treated specially, shouldn't all devices (including DIR) be treated uniformly?

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	camn c,dirdev		; is what remains the DIR device?
+	  jrst dirchk		;  yes, continue checking hhDIR case
+
+				; is it hhhdir or hhh
+	hrlz c,a		; get device in c
```
bictorv, line 130: This tests is the same as the one just before host3 (in reverse)?

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	jumpe c, host3		; jump if no device
+
+	ldb c,[000600,,a]	; can only be hhddd if last sixbit character is null
```
bictorv, line 133: Here (in hhddd1) you could handle HHDIR too.

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
-	 JRST [	MOVS A,[SIXBIT/MCDVS/]
-		MOVSM A,JBCDEV
-		MOVEI A,'MC
-		JRST GOTHST]
-	CAIN A,'TPL	;TPL THROUGH THE NETWORK FOR MC
-	 JRST [	MOVS A,[SIXBIT/MLTPL/]
-		MOVSM A,JBCDEV
-		LDB B,[000400,,JBCOP]
+	hrlz c,a		; check for hhhddd, get ddd in c
+	jumpn e,hhhddd		; is there a ddd?, if so handle it
```
bictorv, line 137: should be "jumpn c,hhhddd", right?

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	jrst chkits		; nope, assume that we just have hhh
```
bictorv, line 138: chkits does what host3 does, right? So you can remove this line and the chkits code. And neither code fills in dskdev for device, does it?

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	move b,a		; hostname in A
+	pushj p,isitsp		; is it a known ITS
+	  jrst notits		;  nope, see if special device
+	move c,dskdev		; we have an ITS name, so assume DSK device
+	movem c,jbcdev		; save remote device
+	jrst gothst		; and process our host
+
+dirchk:				; 2-char hostname with DIR device
+	ldb b,[.bp (777700),a]	; get hh part in b
+	lsh b,24.		; left justify hostname
```
bictorv, line 150: But you haven't tested (with isitsp) that it is an ITS name. Also this code is very similar to hhddd1/hhddd2. Here you have tested that the device is DIR and then store dirdev (DIR) as device, while in hhddd1/2 you just take whatever device there is, which seems more general?

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+dirch2:	move c,dirdev		; set remote device as DIR device
+	movem c,jbcdev
+	move a,b		; get host in a
+	jrst gothst		; process host
+
+hhddd2: ldb a,[301400,,jbcdev]	; get hostname in A
+	lsh a,24.		; left justify hostname
+	movem c,jbcdev		; set remote device from C
+	jrst gothst		; process host
```
bictorv, line 159: Also here you haven't tested if it is an ITS name.

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+
+hhddd1:				; continue check for hhddd
```
bictorv, line 161: Stylistic, but I would suggest hhddd1 before hhddd2 (and changing the test "jumpn c,hhddd2").

src/sysen2/mldev.108
```diff
@@ -94,69 +96,95 @@ GO:	MOVE P,[-LPDLL-1,,PDL-1]
...
+	jumpn c,hhddd2		; jump if we have a device; get hostname and set remote device
+
+	;; fall through; assume just an ITS name
+chkits:	move b,a
+	pushj p,isitsp		; is b an ITS name?
+	 jrst notits		;  no, process as special device
+        jrst gothst		; yes, process this host
+
+notits:				; here we've determined that hostname is invalid, check for special device
+	MOVS C,JBCDEV
```
bictorv, line 173: Since these machines, in these configurations, no longer exist and probably won't, I suggest commenting out this hackery.

src/sysen2/mldev.108
```diff
@@ -166,22 +194,45 @@ notits:	;; nothing came of it, fall through
...
+		JRST GOTHSX ]
 ;HOST NOT RECOGNIZED
 NSD:	MOVSI A,%ENSDV
 	JRST NOGO2
 
+hhhddd:				; get device from last 3 characters
+	hrlz c,a		; get device and left justify
+	movem c,jbcdev		; save remote device
+	hllzs a			; get hostname
+	jrst gothst		; process host
```
bictorv, line 206: Also here you haven't tested if it is an ITS name. Wouldn't it be "smoother" to change the "jumpn c,hhhddd" above to a "skipn c" + "move c,dskdev"?

src/sysen2/mldev.108
```diff
@@ -166,22 +194,45 @@ notits:	;; nothing came of it, fall through
...
+
+doshoe:				; device is SHOE
+	lsh c,12.		; left justfify device
+	movem c,jbcdev		; set remote device
+	ldb a,[301400,,a]	; get 2-character hostname
+	lsh a,30		; left justify hostname
+	jrst gothst		; process host
```
bictorv, line 213: Maybe "gothst" should test if it is an ITS name, but I'm not sure.

bictorv (changes requested):
I'd prefer if we had another round at this?

src/sysen2/mldev.108
eswenson1: You must be looking at the old code.  Please look again at the latest.  That code fragment no longer appears in MLDEV 108.

src/sysen2/mldev.108
eswenson1: Again, you're commenting on old code.

src/sysen2/mldev.108
eswenson1: Same -- please review latest code.  At hhddd1, we now have:

```
hhddd1:				; continue check for hhddd
	ldb c,[003000,,a]	; get device portion
	lsh c,12.		; left-justify
	jumpn c,hhddd2		; jump if we have a device; get hostname and set remote device

	;; Fall through; assume just an ITS name
```
which does handle the hhddd case.

src/sysen2/mldev.108
eswenson1: Yes. This was already fixed in later commit.

src/sysen2/mldev.108
eswenson1: CHKITS assumes that we've already set the remote device (JBCDEV), while host3 actually forces the DSK device. CHKITS is used in two places -- both times to check to see if the hostname is a known ITS system, and if not, jump to the code that checks for the hardcoded, different-machine devices (e.g. XGP, GLP, DVR, DVS, and TPL). If the hostname *is* a known ITS machine, it jumps to GOTHST.

src/sysen2/mldev.108
eswenson1: That is true.  However, the old code didn't either.  I could add a check here -- but if the host is not known, then the host lookup will simply fail and we'll get a NO SUCH DEVICE error.  The only time we *really* need to call CHKITS is when there may be some ambiguity regarding the assumptions of what part of the original device is a hostname and what part is a device name.

src/sysen2/mldev.108
eswenson1: Well, I thought about this -- however it is good documentation as to where these additional devices used to live. Also, we may implement one or more of these devices at some point (TPL, XGP), and it may be that we can only implement them on a KA (or KL or KS) and if that is the case, this logic might still make some sense.  I don't see what the harm is in leaving it.

src/sysen2/mldev.108
eswenson1: Not sure. You are reviewing old code.  Please take a look at the latest.

src/sysen2/mldev.108
eswenson1: It doesn't really matter.  GOTHST will convert the hostname to ascii and then look it up in the host table. If this fails, then we'll fail the call with a "NO SUCH DEVICE".  As mentioned above, the reason for CHKITS is really to handle the cases of ambiguity.  I find the chkits (and itsnms) a little "hacky" anyway -- since it requires rebuilding ITS in order to add a new known host.  I think it is actually better to simply use a hostname lookup to determine if the host is a known host.  The failure of MLDEV in either case is the same.  The only thing that CHKITS really gives us is some control over ambiguity and a way to "escape" to the "translated" device handling (e.g. TPL, XGP, etc.).

src/sysen2/mldev.108
larsbrinkhoff: I like to entertain the idea that some day the classic AI, DM, ML, MC machines be brought back under simulation.  I also think think they are have an important place in ITS history and should stay in ITS source code files; I believe this is in the spirit of ITS hacking.

larsbrinkhoff:
FYI, I'm going to squash those fix commits into the first commit.  Force push alert!

bictorv:
It's really weird about me commenting on old code: I'm pretty sure I clicked on "files changed" and commented on the code presented there. Also when I look at the new code in the master branch, e.g. "jumpn e,hhhddd" still appears, and the other code I commented on also. Please double check that what appears in the repo is what you intended?

I think I agree on skipping ITSNMS, but then I think it should be skipped consistently. Better to check for a valid host name with HSTLOOK. 

For the XGP etc hackery, what I suggested was to comment it out, not remove it. But as it is, it prevents us from using configs other than what is hardcoded here, doesn't it? Hardcoded machine names should not be used, I think, except for "aliases" such as local-time-server etc.

eswenson1:
@bictorv: I just pulled master branch and I do see alignment between your comments and the code -- I'm not sure why I didn't think it matched before.  I'm going to review all these comments again to make sure they don't reveal outstanding issues. 

I also compared the current mldev.108 in master branch with the mldev.108 that I'm running on ES (where I did the development and testing) and they match as well.  So I did test what I committed and what is on the master branch now.  Still, I'll re-review your comments.

src/sysen2/mldev.108
eswenson1: At this point, if C was 'EXDIR ', the LSH makes it 'DIR', the JRST DIRCHK continues the check for HHDIR, since we know that we have 'xxDIR ' at this point.  If we skip, however, which would be the case with, HHHDIR, we continue on with the following code:
```
        lsh c,12.               ; move past hostname
        camn c,dirdev           ; is what remains the DIR device?
          jrst dirchk           ;  yes, continue checking hhDIR case

                                ; is it hhhdir or hhh
        hrlz c,a                ; get device in c
        jumpe c, host3          ; jump if no device
     ```
     
     This part distinguishes between HHHDIR and HHH cases (we know we don't have HHDIR at this point). 
     
    But you are right that DIR is being handled specially here, but the HHDIR and HHSHO (for example) cases should be treated the same -- e.g. as HHDDD.  The reason DIR is handled specially is because in the original code it was as well.  But you are probably right that I could just check for HHDDD and HHHDDD where DDD = 0, and jump to code that would handle any device, including DIR.

File: src/sysen2/mldev.108 (hunk 21 above)
Comment (eswenson1) [reply src/sysen2/mldev.108:130]: This is the case where we tried to parse as HHDIR, but in fact the input was HHHDIR.  So we HRLZ to get the DIR (or DDD) in c, check for zero, and jump to HOST3 if there is no device.  This is the HHH case.
```
        move c,a
        lsh c,12.               ; move past hostname
        camn c,dirdev           ; is what remains the DIR device?
          jrst dirchk           ;  yes, continue checking hhDIR case

                                ; is it hhhdir or hhh
        hrlz c,a                ; get device in c
        jumpe c, host3          ; jump if no device
```

src/sysen2/mldev.108
eswenson1: Oh!  Yes.  You were commenting on the wrong register not the JUMPN/JUMPE sense, which was corrected already.  Yes, this is wrong.  I wonder why that bug allowed all tests to pass?  I'll fix that, and also get rid of the DIR special-casing.  But I'm going to test the existing code -- that specific spot too to see why it "worked".  But yes, that should be "C".

eswenson1: Interesting. I tested EXADIR as the device, and it didn't work. I'm pretty sure I did test this case earlier, so that means I wasn't testing the code I thought I was.  I'll redo all the testing after fixing this JUMPN E.  I may also get rid of the DIR special-casing (not DIRhh and DIRhhh, but the HHDIR and HHHDIR cases).

eswenson1:
I have a vastly simplified version of MLDEV now, that appears to work.  I've tested EX, EXA, EXDIR, EXADIR, DIREX, DIREXA, XGP, EXSHOE, EXSHO, XEX, XEXA.  

Have I forgotten any cases?  I'll submit a new PR for this version and WON'T MERGE IT until both Björn and Lars are happy.  I'll also run with it and give it more of a workout on my various ITS systems.

PR #2233, opened 2023-10-05
Network booting for MINITS

larsbrinkhoff:
- 11LOAD for serving MINITS network booting on Chaosnet/Ethernet.
- PDP-11 ROM for network booting using 11LOAD and an Interlan network interface board.


eswenson1 (approved):

eswenson1:
@larsbrinkhoff What capabilities does this PR give us?  Anything I should try?  Why would I care to try?

larsbrinkhoff:
There is currently nothing that can use these capabilities.  The boot ROM is for an Interlan board, which is not (yet) supported by SIMH PDP-11.  It would be interesting to add that support (see e.g. https://github.com/simh/simh/issues/380).  It's an Ethernet interface, which apparently was used for Chaosnet traffic, so that's another complication: we would need to pass traffic between Ethernet and a pure Chaos network.  I think @bictorv's Chaosnet bridge can do that, so it's not a big complication.

bictorv:
I'm not sure what you mean by "pass traffic between Ethernet and a pure Chaos network", i.e., what do you mean by "pure Chaos network"? Original coax-based Chaos hardly exists these days?

My Chaosnet bridge supports Chaosnet on Ethernet (protocol 0x0804) and several other encapsulations.

larsbrinkhoff:
The network simulated by the CH11 and CH10 devices in SIMH is what I call "pure Chaos network" in the sense that there are just Chaosnet packets.  The Interlan board is for Ethernet, so there will be some encapsulation (hence not "pure"); hopefully EtherType 0x0804.  So MINITS with an Interlan board wouldn't be able to talk directly to ITS using CH10/CH11.

larsbrinkhoff:
Confirmed that BOOTIL and MINITS use EtherType 0x0804 for Chaosnet over Ethernet.

So, assuming SIMH (or other PDP-11 emulator) was supporting the Interlan board, it should be possible to set up a PDP-11 with the Interlan boot ROM using Ethernet (possibly a virtual bridge) and Chaosnet-bridge that to a "pure" Chaos network with an ITS host serving 11LOAD.

PR #2242, opened 2023-10-11
Add a 1 second delay before accessing the boostrap disk.

gwright83:
On my FreeBSD CI system, on a relatively old server, I was seeing "disk controller not ready" errors when trying the format the bootstrap disk pack when building ITS with simh.  This error occurred consistently when running on CI server, but did not happen when entering the commands manually.

The fix is to add a 1 second delay before trying to format the bootstrap pack.  This fixes the problem seen in CI and has no effect when building on may (much faster Apple M1) laptop.

### DIFF
--- a/build/ks10/include.tcl
+++ b/build/ks10/include.tcl
@@ -22,6 +22,7 @@ proc start_dskdmp_its {} {
 }
 
 proc mark_pack {unit pack id} {
+    after 1000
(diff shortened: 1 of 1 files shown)


larsbrinkhoff (approved):
I was considering something like this myself, because the same thing happens with GitHub Actions.  Thanks!

Issue #1683, opened 2019-08-09
Forcibly shut down right now


larsbrinkhoff:
Log out all users immediately and shut down right away.

Needs a hairy program to identify all logged in users and gun them down.  Should be possible to borrow code from PEEK.

CC @ashkitten

larsbrinkhoff:
Word on the street is @eswenson1 went ahead and implemented this.

eswenson1:
I've written two programs.  The first is an command that guns down all jobs, then schedules a shutdown for 5 minutes from the current time, then guns its own HACTRN.  This program (if useful) could be used with an expect script from a host to force ITS to shut down immediately.  

The second is a TCP server.  You connect to port 511. (0777) on ITS and it guns all jobs and schedules the system to shut down. Since there are no jobs left, ITS shuts down immediately.

Not sure what to do with either of these. The server is dangerous -- anyone who connects to port 511 on the ITS system will initiate this immediate shutdown. One idea is to require the client that connects (e.g. with TELNET) to send a "password" and have the server reject any password except one burned into the server.  

What do we think we should do with either or both of these programs?

@larsbrinkhoff @ashkitten

larsbrinkhoff:
Some ideas:
- The two program could possibly be merged into one.
- Chaosnet and NCP could be alternative network methods.  ARPANET security is quite good these days.
- Yet another entry point could be a JOBDEV.  But that may be a lot of work.
- I don't see a problem adding this now, but I don't think it should be enabled by default.  Provide instructions for those who need it.

larsbrinkhoff:
How about instead of/in addition to network access, hook up a signal handler?  This requires changes to the simulator, of course.

eswenson1:
What do you mean by a signal handler?  How would the updated emulator "signal" to ITS?  Would it generate an interrupt or fault?  And wouldn't we have to add support to ITS to handle the signal?  Or do you mean that we'd add a new device to the emulator that could be attached by an ITS daemon (or program) and it could listen for interrupts on a channel and handle those?

larsbrinkhoff:
Something like the latter.  The signal would be to the emulator.  This would maybe be more convenient, and it's not available to do from outside the host machine.

I'm not proposing you change your program, just jotting down an idea.

larsbrinkhoff:
@eswenson1, @ashkitten would like to try your program.

eswenson1:
[...]

PR #2246, opened 2023-10-21
Shutdown server.

larsbrinkhoff:
This is a Chaosnet server for quickly shutting down ITS.  For e.g. the PiDP-10, users are assumed to power off the host computer and expect ITS disks to be in a consistent time next time it boots.  Thus we need a mechanism for the host to tell ITS to shut down quickly and orderly.  See also #1683

Posting a draft for comments.

@ams suggested the shutdown request to include information on how to shut down, or rather, reboot.  There could be an optional character, or string, for this.

More important is some kind of security token.  Something to restrict who can shut down the target.

If we want to overengineer things, a shutdown time could be included.

### DIFF
--- a/build/build.tcl
+++ b/build/build.tcl
@@ -195,6 +195,10 @@ respond "*" ":copy sys; ts ddt, dsk0: backup;\r"
 respond "*" ":copy sys; ts dump, dsk0: backup;\r"
 respond "*" ":copy sys; ts midas, dsk0: backup;\r"
 
+if [file exists $build/mchn/$mchn/custom.tcl] {
+    source $build/mchn/$mchn/custom.tcl
+}
+
 if {![info exists env(NODUMP)]} {
     # make output.tape
     respond "*" $emulator_escape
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
As per @ams's suggestion, I added the security measure "check the source address".  Maybe that is enough for the PiDP-10 use case, in which 177001 is the bridge, which is the source for NCP packets.

larsbrinkhoff:
In order to enable this feature on some machines, I have introduced a MCHN-specific custom.tcl script that is run at the end.  Only TT has the shutdown server enabled.

I also added a commit to customize TT and HX to use the trivial gunner.

larsbrinkhoff:
This server goes together with https://github.com/Chaosnet/chaosnet-tools/blob/master/shutdown.c

eswenson1 (approved):
That’s a nice/tiny implementation!

larsbrinkhoff:
Right, it doesn't have to do much!

PR #2265, opened 2024-01-04
Update deploy-ftp.sh

oilcan-productions:
changing hostname hactrn.kostersitz.com

### DIFF
--- a/build/deploy-ftp.sh
+++ b/build/deploy-ftp.sh
@@ -2,9 +2,9 @@
 
 set -e
 
-HOST="hactrn.kostersitz.com"
-USER="hactrn@kostersitz.com"
-DIR="images"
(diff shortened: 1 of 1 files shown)


oilcan-productions:
Updated FTP_SECRET with the new password. Once this is merged we should be able to re-publish

oilcan-productions:
Note that DNS replication for the change is still in progress and might take 48 hours to complete before it all starts working again

larsbrinkhoff:
Thanks!  I squashed your three commits together.  The real test will be when we push to master, that's when the deploy script kicks in.

larsbrinkhoff (approved):

larsbrinkhoff:
Ok, let's give this a try.

Issue #1969, opened 2020-11-19
Check build dependencies


larsbrinkhoff:
Add a small script, run early, to check for build dependencies.  It should cover all "supported" host systems:

- Debian
- Ubuntu
- MacOS
- FreeBSD

And maybe some more... @ams how's the OpenBSD build these days?

CC @pgalbraith

ams:
Still broken last time I checked -- I will try again.

pgalbraith:
Thanks @larsbrinkhoff , I'd say add them to the main readme first, after all I'm sure any change in dependencies is going to necessitate either a change to the readme, or a manual change to this proposed script, and the former seems easier.

PR #2268, opened 2024-01-23
Update dependencies.sh

oilcan-productions:
Fixing up Mac OS dependencies to make sure all sdl2 bits and tools are there

### DIFF
--- a/build/dependencies.sh
+++ b/build/dependencies.sh
@@ -33,9 +33,9 @@ install_freebsd() {
 install_osx() {
     brew update
     case "$EMULATOR" in
-        simh*) brew install sdl2;;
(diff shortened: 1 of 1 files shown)


oilcan-productions:
Related to https://github.com/PDP-10/its/issues/1969

larsbrinkhoff:
Thanks!  The dependencies.sh script is mostly for continuous integration.  I gave up on MacOS, because I don't have such a machine myself.  Maybe it's time to give it another try?

Another thing, KLH10 doesn't need SDL2 since it doesn't have any graphics.

larsbrinkhoff (approved):

larsbrinkhoff:
Ok merged, thanks!

PR #2273, opened 2024-02-20
Update README.md

jordanhubbard:
install_macosx function has been renamed to install_osx

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ the script takes 2 parameters to run:
 `install_<os>` and `EMULATOR=<emnulatorname>`
 To install the dependencies 
 - on MacOS for the PDP-10 KA run
-  - `./dependencies.sh install_macosx EMULATOR=pdp10-ka`
+  - `./dependencies.sh install_osx EMULATOR=pdp10-ka`
 - on Ubuntu run
   - `./dependencies.sh install_linux EMULATOR=pdp10-ka`
 - on FreeBSD run


eswenson1:
Why the rename to install_osx? OS/X is an old/obsolete name for the Mac operating system. The correct name for the OS is now macOS (spelled that way). 

I’d recommend “install_macOS”.

jordanhubbard:
Because that is the name of the function in the shell script - I assumed someone changed the name in the script without changing the documentation.  As long as both are in sync, it’s fine.

ams:
> I’d recommend “install_macOS”.

Lower case please... since this is something a user is going to type.

jordanhubbard:
I didn’t mean to start a naming war, particularly since I’m a little too close to macOS to be opinionated about it, I am merely trying to align README.md and build/dependencies.sh since they are currently out of sync.

FWIW, the official Apple brand is “macOS” - capital O, capital S.  As one of the creators of macOS, however, I don’t want to dictate how others choose to spell it.

- Jordan

jordanhubbard:
FWIW, if we're "voting" on this, I would suggest a single change to name it macOS in both places.  If everyone can align on that, I'll submit two updated PRs that commits *just* the renaming change.

ams:
Typing capital letters for one system is annoying though.  Should it now be Linux? What about

larsbrinkhoff:
This wasn't really meant to be a user facing name.  `dependencies.sh` was just a helper for the continuous integration scripts.  But now the cat's out of the bag, and maybe that's just as well.

The name is part of an identifier, which ends up being called inside a shell script.  Stylistically, I think lower case works better.  Maybe this design is questionable, and the script could just as well take a name as an argument, and act accordingly.  In that case it could even be case insensitive.  If someone wants to do a minor rewrite of the script, give it a go.

larsbrinkhoff (approved):

Issue #2220, opened 2023-09-19
ITS sometimes hangs on shutdown and never returns to NTS DDT


eswenson1:
I've seen the issue on both a KA (pdp10-ka) and a DB (klh10) ITS systems. When you attempt to shut down ITS with LOCK you may run into similar console output:

```
DB ITS going down in 4:59
q
:KILL
*$$u
 LOGOUT EJS    0 16:29:58

DB ITS 1651 SYSTEM JOB USING THIS CONSOLE. 16:29:58

DB ITS 1651 GOING DOWN IN 00:04:56.   16:29:58

DB ITS 1651 NOT IN OPERATION AT 16:29:58
PFTHMG DRAGON CHANNA _DRGN_ TIMES  WRITE  16:30:00
PFTHMG DRAGON CHANNA LOGOUT TIMES  DELRNM 16:30:00
IT IS NOW  4:31:16 PM EDT, MONDAY, SEP 18, 2023
```

ITS does not return to NTS DDT but continues "running" indefinitely.  Analysis of the situation shows that ITS is waiting for all TUTs to get written out to disk, but the code that writes out the TUTs is waiting for the `QACTTM` counter to become positive.  `QACTTM` is set to -1, and then at various times, incremented (with AOS).  Even during the "shutdown hang", it can be seen to increment past 0 (I've seen it go to 1 and 2), but always seems to get reset to -1, and possibly incremented to 1, before the TUT writing code checks it.  

As a result, the TUTs are never written out, and because there may be one or more TUTs marked as needing to get written out, ITS will not execute the logic that returns to NTS DDT.  It is waiting on the TUTs to get written out.

The issue appears to be a timing issue.  At first, I noticed the issue on a pdp10-ka (KA) instance, where I had added disks to make the total number of disks 5. I thought it had something to do with the addition of the 5th disk.  But then, I ran into the same issue with a newly-built KLH10 DB instance, where there had been no changes made -- the build was a pristine one.

There appears to be a negative interaction between the slow clock (which, when in the "shutdown" mode, is the one that checks to see if the TUTs are all written out before returning to NTS DDT), and the disk interrupts and fast clock used to ensure disk i/o is done.  

I added a fix to SYSTEM; DISK >, which guarantees that TUTs will be written out if doing so was skipped over due to `QACTTM` being non-positive more than 10 times.  This allows the system to shut down in the above situation.

eswenson1:
The slow clock runs this code:

```
[...]

PR #2276, opened 2024-02-22
Add fix to ITS to ensure it doesn't hang waiting forever for TUTs to be written to disk.

eswenson1:
Fix guarantees that TUTs will get written out if doing so was skipped due to QACTIM being non-positive more than 10 times.

Addresses #2220.


eswenson1:
@larsbrinkhoff I'm not suggesting we actually merge this yet.  I'd like @desaster to try it out to see if it resolves their shutdown issue.  We can also discuss the fix here to see if we want to change it.

desaster:
I think I managed to install this for testing. So far after a few tries my system has successfully shut down every time:

```
SHUTDOWN COMPLETE
PI LEVEL 7 BUGDDT.  TYPE <ALTMODE>P TO CONTINUE.
YOU ARE NOW IN DDT.
BUGPC/   CAI COFFI4+1   $Q-2/   JRST COFFI7
```

But I'll keep testing more as I keep exploring ITS. I'll also keep testing the old build to see that the issue is still there.

The way I installed it, hopefully actually applying it correctly:

* Copied the file to `FIX; DISK 1228` using the [itstar instructions](https://github.com/PDP-10/its/blob/master/doc/DUMP-itstar.md)
* Copied it to `SYSTEM` with `:copy fix; disk,system; disk`
* Assembled ITS using the [assembling instructions](https://github.com/PDP-10/its/blob/master/doc/NITS.md), skipping disk re-configuration and skipping updating (N)SALV
* Booted the new ITS from DSKDMP as instructed, and later copied it to `DSK0:.; @ ITS`

eswenson1:
@desaster Seems to me like you did it correctly.  Glad it is working for you.  Keep using that ITS (I'm doing the same) so we get some experience with it.  Please let me know if you see the shutdown hang again (when using the fix).  Thanks very much for trying it out.

larsbrinkhoff:
I have a vague feeling this has become more prominent with later KA10 emulator updates.  I will try to look into *why* QACTIM isn't behaving as expected.

larsbrinkhoff (changes requested):
I'm reviewing this as "request changes" to be explicit about waiting to merge something.

eswenson1:
Updated this branch to use the "option 2" fix -- we simply force writing of TUTs and disk buffers on slow clock if we are shutting down.

larsbrinkhoff (approved):

larsbrinkhoff:
All ready to merge now.  Thanks so much for working on this, Eric!

PR #2277, opened 2024-02-23
Rework the way dependencies are executed and documented

jordanhubbard:
1. Renamed all of the OS targets to be compatible with the output of `uname -s` making this less subjective and more declarative.
2. Hid this from being required at all by having a Makefile target ("deps") run the dependencies once (until clean target destroys $(OUT).
3. Fix some bugs with the FreeBSD dependency install (pkg(1) requires sudo).
4. Changed the documentation to reflect these changes and also make FreeBSD users a bit more successful


larsbrinkhoff:
Technically, this looks good.  (Except minor things; the CI configuration in .github need to be updated in concert, some commits could be squashed, etc.)

However, I'm not keen on this short-manned project shouldering the burden of maintaining scripts for installing dependencies on various operating systems.  The dependencies.sh script was only meant to facilitate continuous integration.  Maybe it was a mistake to mention it in the README.md file in the first place.

jordanhubbard:
I do understand your trepidation and can only suggest that it might be a good time to kind of step back (for the project) and ask yourselves whether you want to even support multiple platforms.

I am capable of validating all three of project’s current “choices", for example, and can definitely say that it’s far from “type make and be done” on all three of them.

	1. FreeBSD, for example, requires kernel parameter tuning before it will actually not blow up in various ways during the build, and the simh submodule that has been bundled in also does not build on FreeBSD-current due to a symbol redefinition problem.

	2 Darwin on Mac silicon currently dies here:

(schemedump)
This is SCHEME (DSK:SCHEME;NSCHSY 99 compiled by DB) running in LISP 2156.
   The moon is 6 days, 6 hours, 49 minutes, and 38 seconds past the first quarter.
   
;-1.1920929E-7 NEG ARG - SQRT

;BKPT WRNG-TYPE-ARG

The last command timed out.
make: *** [out/pdp10-ka/rp03.2] Error 1

Based on the evidence I have collected so far, Linux is the only actual target that works which means the others have probably not been tested for some time.

Perhaps simply declaring Linux as the only supported platform would simplify your lives?  It is trivial to create a virtual machine on macOS and FreeBSD to run Linux - people do this all the time.  With a little effort (it doesn’t work quite yet, I’ve tried) it might even be possible to run on WSL2 under Windows.

"Why am I even bothering with all of this?” might also be a legitimate question you’re asking right now.  One simple reason:  I wanted to see if I could get the current Zork sources to load, since it requires an MDL environment.  I’m also an ex-LUSER of MIT-AI, MIT-ML and MIT-DM (HUBRD) and there is some nostalgia involved.  That’s really it, however.  If I could pick the “default user story” it would be “I want an ITS prompt, ideally with all of the cool stuff that used to be installed on one of those three machines" (or perhaps all of them, since I see Tcl files for each of the major applications / subsystems offered).

That’s it, really.  Tell me what modern OS to install, let me type “make” with a few parameters, load up ITS.  If I then want to customize my ITS, tell me what files to hack on.

That’s the user story.  I don’t know what the maintainer stories are.

- Jordan

eswenson1:
The ITS build generally works on macOS, as well as Linux.  These are the only two platforms I use and I'm successfully running ITS under various simulators on these platforms.  (There is a current issue regarding running the KLH10 simulator on versions of macOS with Ventura or later releases, but there is a workaround that makes it work just fine).

jordanhubbard:
> The ITS build generally works on macOS, as well as Linux. These are the only two platforms I use and I'm successfully running ITS under various simulators on these platforms. (There is a current issue regarding running the KLH10 simulator on versions of macOS with Ventura or later releases, but there is a workaround that makes it work just fine).

The problem is that many mac users have moved on - I run Sonoma on everything and it's hard to tell users to downgrade.  I think the question about this project's identity and goals remain, but I will also suggest that if there is a known WAR to the problem on later macOS versions, it should be committed.

eswenson1:
I never suggested that KLH10’s not working on current macOS releases was acceptable. I and others have moved on and are running the latest releases of macOS.

The specific issue of KLH10 and macOS I mentioned was a recently reported bug and no one had investigated it until a few days ago. It was a macOS change (in Ventura) that broke things. We have a solution now and will be updating the build doc to tell users about the prerequisites to running KLH10 on macOS. The “fix” cannot be made part of the build because it entirely depends on a user’s environment — how much shared memory has been allocated and how much is free.

Only the KLH10 emulator has this issue. The other emulators run just fine on macOS. Most of us run ITS under Linux or RPI OSes and for that reason, this issue with macOS was not detected until recently.

larsbrinkhoff:
I will note there is no official stance on what is supported or not.  The project, such that it is, is a vaguely connected set of people with write access to this repository, all with slightly different ideas.  My stance is that I can only really offer (limited) support for the set of operating system that I happen to use on my handful of machines, which is mostly various versions of Ubuntu.  So I wouldn't even say e.g. Red Hat is supported.  Eric and others can somtimes help out in the Mac department.  We do have a CI build on FreeBSD that seems to chug along nicely, but I don't think anyone is actively testing or using the result.

On the other hand, we don't say platforms are explicitly *not* supported.  We are happy to receive improvements if people were to offer them.

larsbrinkhoff:
> If I could pick the “default user story” it would be “I want an ITS prompt, ideally with all of the cool stuff that used to be installed on one of those three machines" (or perhaps all of them, since I see Tcl files for each of the major applications / subsystems offered).

The build script currently picks and chooses things across all machines, and all over the timeline.  The name Frankenstein comes to mind.  There's a tiny bit of provision for dividing things with an eye towards, maybe sometime, building something closer to historically correct machine configurations.  It wouldn't be overly difficult to do this, but currently the default is to stuff in as much as possible.

> Tell me what modern OS to install

I think a recent Ubuntu is the easiest.  Raspberry Pi OS should be fine too.  There should be installation scripts for Rasberry Pi forthcoming out of the PiDP-10 project soon.

> I don’t know what the maintainer stories are.

To tell the truth, I think it's mostly "we want to hack on fun stuff".  There's some sense that it's good to provide a bit of computer history to the public.  But support is very much on a volunteer, limited spare time, best-effort basis.

larsbrinkhoff:
Closing in favor of #2315.

PR #2279, opened 2024-02-25
Fix .TRANS in TJ6.

### LINKED ISSUE #925: TJ6 problems when building TJ6MEM
There are a few problems with our current version of TJ6 that are visible in TJ6MEM:

- .TRANS doesn't work as described in the manual. The example in the manual (page 15.1) ends up setting up a translation that turns all the commas in the rest of the manual into `e`s!

- ~~^D (insert the date) formats the year as 1918.~~ Fixed in #947.

- ~~TJ6 does superscript and subscript by changing the XGP's font baseline adjustment. The XGP documentation says that the baseline adjustment is only reset to 0 when you change font. However, in the superscript examples in the manual (e.g. page 16.1), TJ6 emits an extra baseline adjustment after the font change, leaving the baseline where it was until another font change happens to reset it some time later. So either TJ6 is confused, or the XGP doesn't work exactly the way the documentation describes.~~ Fixed in #924.

larsbrinkhoff:
Must read a new character after the comma, before looping around.  Fixes #925.


larsbrinkhoff:
Fixed.  This previously said "Delsh" and "commande."
tj6

eswenson1 (approved):

Issue #2280, opened 2024-03-02
L;STRUCT 659 was updated without bumping up version number


eswenson1:
4 months ago, a commit updated L;STRUCT 659 without increasing the version number.

This makes it really hard for people keeping their ITS systems up-to-date with latest changes from this repo.

This ticket calls for changing the filename to L;STRUCT 660.

PR #2281, opened 2024-03-02
Bump version number for files updated a while back, without a version bump

eswenson1:
The files L;STRUCT 659, SYSEN1;PWORD 2664, SYSENG;DUMP 448, 
SYSNET;TELSER 174, and SYSTEM;TTYTYP 322  were changed in a commit from 
about 4 months ago, but the version numbers were not updated.

This makes it really hard for those trying to update their existing ITS
systems with changes from this repository.

Addresses #2280, #2282, #2283, #2284, and #2285.


larsbrinkhoff:
It seems these missing version bumps will keep happening, and slipping through review.

larsbrinkhoff (approved):

foxkit:
I was always fond of the "name >" file name convention.  It used up a lot
of extra numbers, but the version number incremented every time a file was
saved.

PR #2287, opened 2024-03-08
Add support for XGP user in INQUIR database.

eswenson1:
This PR changes the way that the INQUIR database is installed into ITS. Previously, we included a binary INQUIR;LSR 1 database on the tape we load.  Now, we build the INQUPD program (and INQUIR;DIRS BIN) database, which is required by INQUPD, when run in LSRINI mode.  LSRINI mode, invoked by a JNAME of LSRINI for INQUPD, generates an empty .TEMP.;LSR1 EMPTY database for INQUIR.  The build scripts generate this empty database and copy it to the appropriate plact (INQUIR;LSR 1).

Then, the INQUPD program is run in "normal" mode, such that it processes files in the INQUIR directory with FN1 = .UPD1., and we provide such a file with the XGP user defined.  We can, of course add other users that we wish to be "pre-loaded" in the INQUIR database.


larsbrinkhoff (approved):

larsbrinkhoff:
Right, I understand the GUNNER problem is something else.  This is still a great enhancement.  E.g. XGP can log in and automatically run the .XGPR.; XGP LOGIN script.

I updated the XGP information per a real LSR1 1 file.  I will make a separate pull request to add various system users.

larsbrinkhoff:
Verified that logging in as XGP runs .XGPR.;XGP LOGIN.

PR #2288, opened 2024-03-09
Add INQUIR entries for system users and directories.

larsbrinkhoff:
These entries were added, based on information from backed up LSR1 files: APPLE2, DRAGON, DRAW, GFR, GT40, MINI, PDP11, PEEK, TARAKA, and TEACH.


eswenson1 (approved):

larsbrinkhoff:
Only a draft so far.  I tried putting all users in a single .UPD1. file, but that doesn't seem to work.  Maybe they have to be separate files.

eswenson1:
Really? I’m pretty sure you can have multiple entries. Separated by the correct separator.

if not, we can clearly have multiple .UPD1. files with different FN2s.

larsbrinkhoff:
I tried using a ^_ separator, and I tried having just a blank line.  Neither worked.  Having separate files work.

larsbrinkhoff:
Another aspect is that FILDI may have @ to indicate a directory on a particular machine.  There can be multiple directories, separated by comma.  I'm guessing the first directory is the home directory for the user.  If the @ indicates a machine other than the current one, apparently strange things can happen.  What happened in the build script is that the TEACH directory was TEACH@MC, and when the commands `TEACH$^S` and `EMACS^K` were run, the right EMACS init file was not picked up.  So I think it's best to drop all @\<machine>.

PR #2291, opened 2024-03-13
Updated daylight savings time calculations for the US to reflect current rules

eswenson1:
In 2017, the rules changed from a start date of the 1st Sunday in April to the 2nd Sunday in March. The end date changed from the last Sunday in October to the 1st Sunday in November.


eswenson1:
@bictorv and @larsbrinkhoff (and anyone else) please review the changes for correctness.  I *think* these are correct, but I'm having a hard time wrapping my head around the increment 6, rather than the increment 7, as reflecting the "second Sunday".   I think it is correct as is, but I'd appreciate another set of eyes.

Needless to say, that booting ITS with this change does the correct thing now (considers the time zone to be daylight time rather than standard time), but the devil is in the boundary conditions!

larsbrinkhoff:
I can't say I understand this very well.  E.g. why write `<6+7>` instead of plainly `<13>`?  Probably @bictorv is the one to review this.

eswenson1:
I followed Björn’s convention. 6 days past the start date. Written as n+6. Or n+6.

larsbrinkhoff (approved):
Looks good to me, but @bictorv should review.

src/system/time.953
```diff
@@ -228,24 +228,24 @@ IFDEF DSTEU,[IFN DSTEU,[
 	;; else it's obviously DST
 ]]
 .ELSE [
-	CAML A,[<APR 1>*SPD+7200.]	;IF BEFORE 2AM APR 1,
-	CAML A,[<OCT 31.>*SPD+3600.]	;OR IF AFTER 1AM STANDARD TIME OCT 31,
+	CAML A,[<MAR <1+6>>*SPD+7200.]	;IF BEFORE 2AM MAR 7,
```
bictorv, line 231: With this usage, <MAR 7> is better/clearer.

src/system/time.953
```diff
@@ -228,24 +228,24 @@ IFDEF DSTEU,[IFN DSTEU,[
...
+	CAML A,[<NOV 1>*SPD+3600.]	;OR IF AFTER 1AM STANDARD TIME NOV 1,
 	 POPJ P,			;THEN OBVIOUSLY STANDARD TIME IS IN EFFECT
-	CAML A,[<APR 7>*SPD+7200.]	;IF BEFORE 2AM APR 7 STANDARD TIME,
-	CAML A,[<OCT 25.>*SPD+3600.]	;OR IF AFTER 1AM STANDARD TIME OCTOBER 25,
+	CAML A,[<MAR <7+6>>*SPD+7200.]	;IF BEFORE 2AM MAR 13 STANDARD TIME,
```
bictorv, line 234: As above (this and next line).

src/system/time.953
```diff
@@ -228,24 +228,24 @@ IFDEF DSTEU,[IFN DSTEU,[
...

-		;IT IS NOW SOMETIME DURING THE WEEKS IN APR AND OCTOBER
+		;IT IS NOW SOMETIME DURING THE WEEKS IN MAR AND NOV
 		;WHEN IT MAY BE EITHER STANDARD OR DAYLIGHT SAVINGS TIME
 
 GLPDT3:	PUSH P,A	;SAVE # SECS
 	CAMG A,[JUL*SPD]	;IF NOT YET JULY,
-	 TDZA D,D	;THEN IT MUST BE APR, SET INDEX
-	  MOVEI D,1	;OCTOBER, SET INDEX
+	 TDZA D,D	;THEN IT MUST BE MAR, SET INDEX
```
bictorv, line 247: The comments only reflect the non-EU case.

bictorv (changes requested):

bictorv:
Sometimes I think it would be good to not only test for DSTEU, but also what year we're dealing with, since rules change. We might get a change in EU soon:ish, and then it would be nice to have the code in place, triggering the right case the right year. But we can hope to remember to do this in time for the change. ;-)

src/system/time.953
bictorv: And the original comments are a bit misleading. If it's not yet July, it's the first half of the year (spring on the Northern hemisphere), otherwise it's the second half (fall). This is what is tested, to check which index in CRDSB to use.

bictorv (approved):

PR #2295, opened 2024-03-26
Light show.

larsbrinkhoff:
A program which has a framework for displaying various light patterns on the memory indicators.  It's interrupt-driven to update 60 times a second.  The same binary runs both on ITS and out of timesharing.  Patterns are called as co-routines so they can maintain an internal state and proceed where they last left off.

The program sets ITS location MIPGDS to -1 to have ITS leave the memory indicators alone.  It should be reset to 0 by the user after running this program.

One of the patterns displays data from a pseudo-random number generator.  The algorithm is a 36-bit version of xoroshiro+ with 2<sup>72</sup>-1 period.

### DIFF
--- a/src/lars/lites.79
+++ b/src/lars/lites.79
@@ -1,0 +1,132 @@
+	TITLE Light Show
+
+;Default is to build for both TS and NTS.
+IFNDEF TS,TS==1
+IFNDEF NTS,NTS==1
+
+A=1
+B=2
+T=10
+TT=11
+Q=16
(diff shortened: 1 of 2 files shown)


eswenson1 (approved):
Looks good. But why did you mix submodule updates (with no mention in PR description) and no separation from other two commits)?

larsbrinkhoff:
Oops, the submodules should't be there.  Thanks for checking!

PR #2303, opened 2024-05-02
Set SYS;RECORD TIME timestamp for current uptime record.

larsbrinkhoff:
The TIME program checks the SYS;RECORD TIME timestamp to see if a new uptime record has been set.  It's not a normal timestamp, but is relative to day 0 of month 0 in year 1900.

Thanks to Heinz-Bernd Eggenstein (@Bikeman) for discovering this.

### DIFF
--- a/build/basics.tcl
+++ b/build/basics.tcl
@@ -375,4 +375,9 @@ expect ":KILL"
 respond "*" ":palx gt40;_gt40;bootvt\r"
 expect ":KILL"
 
+# PLAN/CREATE
+respond "*" ":midas sys3;ts create_syseng;create\r"
+expect ":KILL"
+respond "*" ":link sys1;ts plan,sys3;ts create\r"
+
 processor_basics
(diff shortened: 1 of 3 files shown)


eswenson1 (approved):

larsbrinkhoff:
No, that's not what the timestamp says.  The file timestamp is the uptime counted relative the non-existing date 0/0 in the year 1900.  It's not really a date at all.  It's abusing the file timestamp to record a time value.  The right half hour/minute/second fields are the same as in a regular timestamp, but the left half is just number of days.

Your SYS;RECORD TIME timestamp doesn't make sense as a uptime record.  It's probably just from when the file was copied from tape in 1992, and the original timestamp was lost.

Bikeman (changes requested):
Is there a typo in the create command? (timer / time)

eswenson1 (changes requested):
Update the :CREATE command in build.tcl to specify TIME rather than TIMER.

eswenson1:
> Is there a typo in the create command? (timer / time)

Yes, good eye!

larsbrinkhoff:
I didn't remove an r, but I added the missing backslash.

eswenson1 (approved):

PR #2306, opened 2024-05-06
Deploy ftp script updates

oilcan-productions:
Adding a change to ensure we recreate the tarball on retry as the last failures seemed to suggest that actually the created tarball could be corrupt and not the upload.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /build/*/stamp
 /build/klh10/dskdmp.ini
 /user/*/*
+test.sh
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
It looks to me like the previous version already does recreate the tarball every time.  I'll check.

larsbrinkhoff:
I checked, and the line `(cd out; tar cvzf $EMULATOR.tgz $EMULATOR)` does run every retry, and it recreates the tarball each time.

larsbrinkhoff:
Closing since the functionality already is there.

PR #2307, opened 2024-05-19
Fix MCL (macsyma source compiler) to actually work.

eswenson1:
It is invoked with :MAXTUL;MCL and can be used to compile any macsyma source.  Of course, if you change IN-CORE sources, you'll have to dump out a new macsyma, but for non-IN-CORE sources, you'll get the FASL and UNFASL files in the right places, and any macros loaded that are needed.

The original link of maxtul;.good. complr was incorrect -- it pointed to SYS;TS COMPLR. The link is supposed to point to the dumped out MCL image.

### DIFF
--- a/build/macsyma.tcl
+++ b/build/macsyma.tcl
@@ -216,9 +216,22 @@ respond "File name->" "\002"
 respond ";BKPT" "(quit)"
 
 respond "*" ":midas maxtul;ts mcl_mcldmp midas\r"
-respond "*" ":link maxtul;.good. complr,sys;ts complr\r"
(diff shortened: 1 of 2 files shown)


eswenson1:
Hmmm… now MCL works fine, but as configured, compile_lisp_file in MACSYMA doesn’t work. I’ll have to investigate further.

eswenson1:
I've updated this PR to correctly build UTMCL -- the lisp compiler used by COMPILE_LISP_FILE in Macsyma.

larsbrinkhoff (approved):

PR #2312, opened 2024-06-10
Update h3text.2018: add N3UC hosts

bictorv:
Michael Wyrick joined the Global Chaosnet.


larsbrinkhoff:
Apparently there's a problem building on FreeBSD with CirrusCI.  Looking into that.

larsbrinkhoff (approved):

PR #2315, opened 2024-06-15
The dependencies.sh script is only for CI.

larsbrinkhoff:
Removed the README.md text about dependencies.sh because it's only to aid continuous integration builds.  We don't want the maintenance burden of keeping this up to date for everyone's operating system release.


eswenson1 (approved):
I’ll trust your judgement, but I think it should have stayed, with perhaps a caveat added to allay your concerns. For the most widely used OSes, it is helpful for people to know what the dependencies are and how to get them installed.

larsbrinkhoff:
The suggested change in #2277 is adding more support for various targets.  I for one am not ready to take on the maintenance of such things.  If someone is, I suggest starting a separate repository with helpful scripts.  Depending on how that turns out, it could be folded back into this repository.

PR #2319, opened 2024-07-23
Updated SLAVE interface to Type 340 display to correct an obvious bug.

eswenson1:
The value of the timeout was getting overwritten by incorrect code. I was seeing timeout errors while trying to write to the display from SHRDLU.  The fix seems to help a lot.


larsbrinkhoff:
My initial concern was that `MOVEI T,100.` was overridden by the subsequent `MOVEI T,1`.  Your fix doesn't address that.

However, I now see it's more complicated than that.  There is also a `MOVEI T,10.`, and `T` is used as a loop counter.  There's also a *second* loop nested around that.  So now I'm not sure at all what this code is supposed to do, only that there are certainly multiple conflicts around the use of `T`.  We'll have to rethink this, and the fix, a bit.

eswenson1:
I did fix the overwrite of T -- it no longer does that.  What I noted, from another similar sleep loop, in the code, which used the D register for the .sleep argument, was that the T value was supposed to be a loop counter, not the value to be passed to .sleep.  So we initialize the loop counter (T) with either 20 (for PDP6) or 100 (for PDP10) and loop (over the .sleep) a maximum of that many times before giving up.

larsbrinkhoff:
Ok thanks, I understand now and agree with your fix.

larsbrinkhoff (approved):

PR #2320, opened 2024-07-25
Fixed various bugs with the current version of SHRDLU.

eswenson1:
Also added some descriptive files with info about the various files in the source directory.


larsbrinkhoff:
I split off the formatting/whitespace changes into its own commit.  The end result is identical to the orginal commits.

src/shrdlu/blockl.6
```diff
@@ -250,13 +250,15 @@
 (DEFUN PACKO
        (OBJ TYPE)
        (PROG (XX)
+	     (setq xx nil)
```
larsbrinkhoff: PROG should bind XX to NIL already (per the Pitmanual).  Is this necessary?

src/shrdlu/blockl.6
```diff
@@ -250,13 +250,15 @@
...
 	     (MAPC (FUNCTION (LAMBDA (X)
 				     (AND (THVAL (QUOTE (THGOAL (!IS $?X
 								     $E
 								     TYPE)))
 						 (LIST (LIST (QUOTE X) X)))
 					  (SETQ XX (PACKORD X (SIZE X) XX)))))
-		   OBJ)
+		   (cond ((atom obj) (list obj))
+			 (t obj)))
```
larsbrinkhoff: SHRDLU has a function LISTIFY does this.

src/shrdlu/blockp.5
```diff
@@ -493,7 +493,9 @@ THEOREM)
   (X Y BLOCKS PYR (WHY (EV)) EV)
   (!STACKUP $?X)
   (OR (LESSP (APPLY (QUOTE PLUS)
-		    (MAPCAR (QUOTE (LAMBDA (X) (CADDR (SIZE X)))) $?X))
+		    (MAPCAR (QUOTE (LAMBDA (X) (CADDR (SIZE X))))
+			    (cond ((atom $?x) (list $?x))
+				  (t $?x))))
```
larsbrinkhoff: Can use LISTIFY.

src/shrdlu/blockl.6
eswenson1: I was getting an XX unbound variable here and the change appeared to fix it, but I was stumped because both the circa 197x LISP and the current LISP do bind prog variables to nil. I can try to make it through the script without that change and we can see if it made any difference.

src/shrdlu/blockl.6
eswenson1: I’ll check out listify and change to use it.

eswenson1:
I've addressed your review comments, @larsbrinkhoff.

larsbrinkhoff (approved):
Approved.

PR #2323, opened 2024-08-13
Fixed lots of bugs and made it so SHRDLU and PLNR can be compiled.

eswenson1:
Now, we build SHRDLU;TS SHRDLU and SHRDLU;TS PLNR from compiled lisp code.  This should allow it to run faster and have fewer issues with garbage collection and running out of LIST space.

Found lots of bugs going through the compilation effort.  Also fixed several issues that were resulting in weird/incorrect responses to standard demo prompts.

Below is a list of prompts that I have tested:

pick up a big red block.
grasp the pyramid
find a block that is taller than the one you are holding and put it into the box. what does the box contain?
what is the pyramid supported by?
how many blocks are not in the box?
is the red cube supported?
can the table pick up blocks?
can a pyramid be supported by a block?
can a pyramid support a pyramid?
stack up two pyramids.
stack up two red blocks.
put the green pyramid on the red cube.
which cube is sitting on the table?
is there a large block behind a pyramid?
put a small one onto the green cube that supports a pyramid. put the littlest pyramid on top of it.
does the red cube support anything?
what color is the block that supports the green pyramid? how many things are on top of green cubes?
had you touched any pyramid before you put the green pyramid on the little cube? when did you pick it up?
why did you pick it up?
why did you clear off that cube?
how did you clean off the red cube?
how many objects did you touch while you were doing it? put the blue pyramid on the block in the box.
is there anything which is bigger than every pyramid but is not as wide as the thing that supports it? thank you.

Of course, lots more things are possible.

It is still best to do this without answering "Y" to the Type 340 display prompt at the start.  There is either an issue with the GRAPHF module or the SLAVE module (or perhaps with the Type 340 simulator) that is causing crashes after several successful commands with graphic output.


larsbrinkhoff (approved):
Approved.

PR #2325, opened 2024-08-17
Fixed two issues with SHRDLU including the shape change on putting down an object

eswenson1:
Fixed an issue with SHRDLU where picking up the green pyramid and putting it down on top of the table or another object turned it into a block (as far as the display was concerned, not the PLNR state).

Also fixed an issue with SHRDLU where, if the answer to a question was supposed to be NOTHING, it sometimes became "THE IL".  This was because of a missing case being handled in LISTNAMES -- one to handle a NIL value passed in as the object.

### DIFF
--- a/src/shrdlu/newans.82
+++ b/src/shrdlu/newans.82
@@ -783,7 +783,8 @@ TEST-LOOP
        ;;ANSWER HAS BEEN DECIDED ON.
        (PROG (COUNT EXAM X RES ANS COMMA?) 
 	     (SETQ NAMES (MAPCAR '(LAMBDA (X) (NAMEOBJ X SPEC))
-				 (cond ((atom names) (list names))
+				 (cond ((null names) names)
+				       ((atom names) (list names))
 				       (t names))))
 	                                                               ;NAMEOBJ RETURNS A LIST OF THE OBJECT AND THE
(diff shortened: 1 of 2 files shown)


src/shrdlu/newans.82
```diff
@@ -783,7 +783,8 @@ TEST-LOOP
        ;;ANSWER HAS BEEN DECIDED ON.
        (PROG (COUNT EXAM X RES ANS COMMA?) 
 	     (SETQ NAMES (MAPCAR '(LAMBDA (X) (NAMEOBJ X SPEC))
-				 (cond ((atom names) (list names))
+				 (cond ((null names) names)
+				       ((atom names) (list names))
 				       (t names))))
```
larsbrinkhoff, line 788: Isn't this what LISTIFY (or LISTIFY2) should be doing?

larsbrinkhoff (commented):
Good job separating two commits for two different issues!  I'm just curious if the listify thing can be used consistently throughout the code.

larsbrinkhoff (approved):
Approved.  Take what you will from my comment.

eswenson1:
I’ll compare the two versions of listify (both different from the MacLISP listify), as well as this code to see if I should switch to one or the other. I remember noting that the two versions of listify in SHRDLU are different from each other.

PR #2326, opened 2024-08-17
SSIMP should use 18-bit mapping, not 16...

dseagrav:
...otherwise it will fail on a non-emulated KS10. Tested and confirmed on a non-emulated KS10.

### DIFF
--- a/src/system/imp.369
+++ b/src/system/imp.369
@@ -187,7 +187,8 @@ IF2,IFN IMPIBF&777,.FATAL IMPIBF not on DEC page boundary
 IMPINI:	SETOM IMPUP		;Not up yet,
 	SETOM IMPTCU		; but thinking about it.
 	MOVEI A,IMPIBF_-9.	;DEC page # of IMP buffer page
-	TRO A,%UQ16B\%UQVAL	;Valid mapping, 16 bit device
+IFE SSIMP,TRO A,%UQ16B\%UQVAL	;Valid mapping, 16 bit device (for real imp)
+.ELSE	TRO A,%UQVAL		;Valid mapping, 18 bit device (for ssimp)
 	IOWRI A,UBAPAG+IUIMPG_1	;Set up 1 DEC page of UBA mapping. Note that
 				; the second half of IUIMPG isn't mapped at all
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
As far as I can tell, we'll build ITS for KLH10 with SSIMP=1.  Wouldn't this change the behavior for KLH10 as well as your real KS10?  Will KLH10 still work?

larsbrinkhoff (approved):
Approved.

larsbrinkhoff:
@dseagrav explained elsewhere that KLH10, whilst indeed using the SSIMP interface, does not respect the 18-bit flag and sends the 16-bit format.

eswenson1:
Do all configs we build for set SSIMP==1 that need to? Looks like DB does, but what about others?

larsbrinkhoff:
Only KLH10 uses SSIMP==1.  All other KS10 varianse use SSIMP==0.  SSIMP only applies for KS10 using the ACC LH/DH interface.

dseagrav:
What exactly is going on is:

1: When SSIMP is nonzero, KLH10 simplifies the IMP model so the emulated KS does not have to rearrange bits for Unibus transfers. Instead of rearranging a 32-bit word into two 16-bit words, it sends a 32-bit word left justified in a 36-bit word.

2: Whether or not SSIMP is nonzero, the 16-bit transfer flag is set in the UBA mappings for the LH/DH interface.

3: KLH10 does not actually honor the 16-bit flag - All 18 bits of each unibus transfer are available even when the 16-bit flag is set. Therefore KLH10 always acts as though the 16-bit flag is not set.

I found this when making SSIMP run on a physical KS10. The change should not affect KLH10's behavior in any way, since it ignores the 16-bit flag. It only removes the gotcha for anyone who runs on physical hardware and implements SSIMP instead of the full IMP.

eswenson1:
Thanks for the explanations. I’m ok with this fix since you’ve verified that it works with KLH10 and doesn’t apply to the other simulators. I assume it works fine with pdp10-ks too?

eswenson1 (approved):

PR #2327, opened 2024-08-23
Various fixes and improvements to SHRDLU

eswenson1:
Fixed issue with incorrect indefinite name when no antecedent is found.

Fixed issue with the word ISN'T being output incorrectly in a message.

Updated shrdlu build script to pass the file to compile in JCL and to use new load functions.

Made it so that either DEBUGMODE or USERMODE could be used, and so that both modes would not display annoying timer messages.

Got rid of hack of setting CAR and CDR values to T.

The old Maclisp used when SHRDLU was created allowed one to take the CAR
and CDR of atoms.  The former returned internal bits associated with the
symbol, and the latter returned the property list of the symbol.

This was disabled in later verisons of MacLISP, but allowed to be enabled
by setting the value of CAR to T (and the value of CDR to T).

However, doing this masked coding errors that resulted from unintentionally
taking the CAR or CDR of a symbol, when a list or NUL was actually expected.

This commit removes the hack of setting CAR and CDR to T, and adds macros
to replace the use of CAR and CDR in this cases in PLNR and associated PLNR
logic.  These macros are found in the MACROS module.  Making this change,
and removing the duplicated $ reader macro from PLNR (it is already in
MACROS for the benefit of other files), required making changes to the
loader of SHRDLU and PLNR.  I removed the obsolete use of UREAD to load
interpreted files, and replaced with a new NEW-LOAD function.  UREAD was
unable to handle the (status macro $ 'thread) code that needed to be included.


larsbrinkhoff (approved):
Very nice; approved!

PR #2329, opened 2024-08-25
QR code generator.

larsbrinkhoff:
Draw a QR code on a TV display.  A command line may specify an input file, or else input is from the keyboard.


larsbrinkhoff:
I found two improvements I want to make, and made those in a separate commit.  When I have time to verify the program still works, I will squash the second commit.

eswenson1 (approved):

PR #2335, opened 2024-09-02
Updated build to make pdp10-ka-, pdp10-ks-, and pdp10-kl-based systems correctly set up IP network.

eswenson1:
Updated build to make pdp10-ka-, pdp10-ks-, and pdp10-kl-based systems correctly set up IP network.

Now, KA, DB, and KL systems will set their IP address to the value of the IP make variable. They wiil correctly define a host table entry for their machine. They will correctly set up .mail.;names >.
They will correctly configure COMSAT for networking.

The resulting systems should allow COMSAT to come up, and with appropriate IMP configuration in the "run" scripts to perform inbound and output IP network requests.


larsbrinkhoff:
Overall, this look good.  But as discussed on email, the check-dirs target fails in some way that is not immediately obvious why.

larsbrinkhoff (approved):
Approved.

PR #2336, opened 2024-09-04
In COMSAT, let BUGHST default to local host

### LINKED ISSUE #2331: COMSAT should default BUGHST to the local ITS
BUGHST in COMSAT is used as destination for mail to "BUG type recipients" which have no explicit entry in .MAIL.;NAMES. The default is `HN$MC`, i.e. MC.LCS.MIT.EDU at Chaosnet address 3131 (not in use on the global Chaosnet).

The build process sets BUGHST to the local ITS address, to keep things local. However, if your IP/Chaos address changes, you must remember to patch COMSAT with the new address, otherwise BUG email will be sent to the old address.

A simplification would be to let BUGHST default to 0, and change the code in RXPBUG to check for 0 and then use the local host address instead. (Currently it checks only if BUGHST is the local address or not.)

Thoughts?

bictorv:
BUGHST in COMSAT is used as destination for mail to "BUG type recipients" which have no explicit entry in .MAIL.;NAMES. The default is HN$MC, i.e. MC.LCS.MIT.EDU at Chaosnet address 3131 (not in use on the global Chaosnet).

The build process sets BUGHST to the local ITS address, to keep things local. However, if your IP/Chaos address changes, you must remember to patch COMSAT with the new address, otherwise BUG email will be sent to the old address.

This PR simplifies this by letting BUGHST default to 0, and change the code in RXPBUG to check for 0 and treat it in the same way as if it was (one of) the local address(es).

Closes #2331.

### DIFF
--- a/src/sysnet/comsat.585
+++ b/src/sysnet/comsat.585
@@ -8418,6 +8418,7 @@ RXPBUG:	PUSHAE P,[C,D,L]
 	  JRST [SETZ B, ? JRST RXPB90]	; Yikes, recursing!  Just ignore...
 	MOVE L,$ARLOC+EQVAR	; All stuff must be CONSed in EQVAR.
 	MOVE D,BUGHST		; Are we the magic host on which most program
+	JUMPE D,RXPB50		;  or is it unspecified?
 	CAME D,OWNHS2
 	CAMN D,OWNHST		;   maintainers reside?
 	 JRST RXPB50		; Yes, send to (BUG RANDOM-PROGRAM)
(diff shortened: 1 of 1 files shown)


eswenson1:
Did you intend to assign to me or to have me be a reviewer? If you did intend to assign to me (and Lars) just what are you expecting us to do?

eswenson1 (approved):

larsbrinkhoff (changes requested):
Would be approved, except the branch needs rebasing, and bumping file version.

larsbrinkhoff:
@eswenson1, @bictorv I suggest this and #2337 can be merged in quick succession with only one file version bump.  There's no added value in run-away version numbers.

bictorv:
Please go ahead with rebasing and version bumping etc? I'm worried I would do something bad - I'm not a git wiz.

larsbrinkhoff (approved):
Approved; I did the rebasing and bumping.

PR #2337, opened 2024-09-04
Let COMSAT recompile .MAIL.;NAMES when HOSTS3 has been updated

### LINKED ISSUE #2330: .MAIL.;NAMES needs reparsing when IP/Chaos addresses change (e.g. for INQUIR to work)
Typically, your system is set up to handle INQUIR updates locally, by defining `UPDATE-ITS-INQUIR` using `UPDATE-INQUIR@DB` (where `DB` is your local ITS name) in `.MAIL.;NAMES >` (henceforth NAMES).

If you change only the IP address of your ITS (after running COMSAT at least once, and update your H3TEXT accordingly, but not NAMES since the ITS name wasn't changed), INQUIR will still attempt to send updates to your old IP address, which typically fails.

The reason for this is that COMSAT uses the "compiled" version of NAMES (named `.MAIL.;  LIST    EQV`) if the "source file" (NAMES) hasn't been updated (by checking file creation date), and the compilation (for efficiency) saves the IP (or Chaos) addresses of hosts used in the source file, rather than their names. If NAMES has been updated, or the compiled file is missing, the source is recompiled. (See EQVCHK in COMSAT.)

In order to make COMSAT use the updated address, a workaround could be to make SYSHST;H3MAKE (or SYSHST;HOSTS3) force COMSAT to regenerate `  LIST    EQV`, e.g. by stopping COMSAT, removing `  LIST    EQV`, and restarting COMSAT (which happens every hour through DRAGON;HOURLY CNAVRL, but why wait?).

Before you suggest that COMSAT could check if the local host address has changed and if so recompile NAMES, note that the problem also occurs for any other host names in NAMES which had their addresses updated.

The suggested workaround wouldn't help if host/address translation was done using DNS or HOSTAB, but we're not there yet so let's keep it simple.

Thoughts?

bictorv:
NAMES is normally recompiled to a binary "EQV file" when either there was no such file, or NAMES has been updated later than the creation date of the EQV file.

However, the compilation translates host names in NAMES to binary host addresses in the EQV file, so if the address of some host name has changed since compilation, the new addresses won't be used.

This PR checks for this case: when NAMES would not normally need recompilation, check if SYSBIN;HOSTS3 has been updated since the last NAMES compilation and if so, recompile to get any fresh host addresses.

All changed code is under a `$$H3CHK` switch.

Closes #2330.

### DIFF
--- a/src/sysnet/comsat.585
+++ b/src/sysnet/comsat.585
@@ -155,6 +155,13 @@ T==:U1	; Alternate names for smashable ACs.
 TT==:U2 ; This is mostly for NETWRK/RESOLV.
 IFE U2-OC,.ERR RESOLV temp ACs lose!!
 
+	; Check if SYSTEM;HOSTS3 has been updated in EQVCHK, to ensure
+	; that current addresses are used for hosts in NAMES.
+	; Note: this only has the intended effect for hosts/addresses
+	; in SYSTEM;HOSTS3. Some other mechanism would be needed for
+	; hosts/addresses from DNS or HOSTAB servers.
+IFNDEF $$H3CHK,$$H3CHK==1
+
 $$DQ==1		; Domain interface code in use (mostly for NETRTS)
 $$DQRN==1	; Use RENMWO hack to cut down on resolver overhead
 ;; These are IFNDEFs for now so that I can play with them without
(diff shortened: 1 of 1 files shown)


eswenson1:
Did you intend to assign to me or to have me be a reviewer? If you did intend to assign to me (and Lars) just what are you expecting us to do?

eswenson1 (approved):

larsbrinkhoff:
Looks like @bictorv assigned us by mistake, and then assigned himself.

@eswenson1, you shouldn't approve this if you want file versions to be bumped.

larsbrinkhoff (approved):

eswenson1 (changes requested):
Please bump file FN2 versions.

Thanks @larsbrinkhoff

eswenson1 (changes requested):
Please bump FN2.

bictorv:
> Please bump FN2.

Lars did already do that? But only once for the two PRs.

eswenson1:
Right. It looks like this PR doesn’t have FN2 bumped — unless he merged this PR’s changed into the other PR. I can’t tell.

bictorv:
You can tell from the FN2 of the changed file. :-)

eswenson1 (commented):
Yes, but this PR still has version 585 and doesn’t get renamed. Did @larsbrinkhoff merge this change into the other PR?

eswenson1 (approved):
Ah! The other PR did the rename from 584 to 585 and this PR modified 585. So all good to go.

larsbrinkhoff:
Yes, version was bumped in the very previous pull request.  Merged now.

PR #2338, opened 2024-09-09
Make BUGHST, TCPGAT, DOMGAT, HDRGAT and GATHST be settable in NAMES.

bictorv:
When changing addresses of e.g. your host system, patching in any updates of the addresses of `TCPGAT` and `DOMGAT` in COMSAT is easy to forget. This PR makes them settable through `.MAIL.;NAMES >` (along with `BUGHST`, `HDRGAT` and `GATHST`).

The attached `NAMES` file documents the settings and has defaults for `*COMSAT-TCP-GATEWAY` (TCPGAT) and `*COMSAT-DOMAIN-GATEWAY` (DOMGAT). The Makefile and build/misc.tcl are updated to use this way of making the settings, instead of the old.

Resolves #2332.

PR #2339, opened 2024-09-13
Add a HOSTS3 entry for the gateway IP.

bictorv:
This is needed to make COMSAT parse the IP address properly (see #2338): the GW address must have a host entry to be parseable by `NETRTS"HANLYZ`.

Without the HOSTS3 entry, parsing e.g. "192.168.0.45" fails. With the generated entry, it succeeds.

I'd say the underlying bug is in `NETRTS` or `RESOLV` (which is called by NETRTS), but that's for another day.


larsbrinkhoff (approved):

eswenson1:
This is needed even if you use the @192.168.0.45 syntax? It seemed to work for me without that. I did add host table entries for my gateway, but I thought I did that only for aesthetics and only after I verified that it was working. I could be wrong.  In any case, is is a good idea to add the host table entries and use symbolic names.

eswenson1 (approved):

larsbrinkhoff:
Squashed.

bictorv:
> This is needed even if you use the @192.168.0.45 syntax? It seemed to work for me without that.

Yes. It didn't seem to work for me. Running with DEBUG non-zero shows debug printouts in STATS, and stepping it shows that HANLYZ (in NETRTS) returns failure.

larsbrinkhoff:
Merged.

PR #2342, opened 2024-11-20
Check checksum on CH11 under KLH10

bictorv:
There is some bug in the CH11 implementation for KLH10 which makes it corrupt packets - quite rarely, but making e.g. remote DUMPs (using the RTAPE protocol) unreliable. I've previously established that it must be the CH11, rather than cbridge or the RTAPE server, that is to blame. My hunch is that there is some synchronization problem in the CH11 implementation.

The CH11 implementation in KLH10 is based on the IMP implementation, and I did it some 20-25 years ago. It uses really hairy mechanisms for synchronization between the KLH10 end (dvch11) and the device process (dpchaos). Hacking this type of C code is far from my favourite thing to do, so I've been putting it off since there are so many other things to hack that are much more fun and rewarding. 

This PR provides a workaround: when running on a KLH10 that uses CH11, ITS computes the checksum when a packet has been received, and thus discovers corrupt packets and simply drops them (after counting it which will appear in STATUS replies). The performance price is really low.

The underlying bug in KLH10s CH11 should be fixed at some point, but that's a separate issue (and not for this repo).

Side note: is "ADDI J,(B)" (significantly) faster than "ADD J,B" on a KLH10?


larsbrinkhoff:
This triggers a HALT instruction on machines other than KS10.

src/system/chaos.291
```diff
@@ -2426,29 +2426,49 @@ BBLK
 	 JRST [	AOS CHNLOS	;No buffers, lose this packet
 		JRST CHXBK1 ]
 	HRLI A,-%CPKDT		;Copy out the packet header
+IFN KLH10P,[
```
larsbrinkhoff, line 2429: This needs a closing `]`.  Or the line could just be removed because the comment doesn't need to be inside a conditional.

src/system/chaos.291
```diff
@@ -2426,29 +2426,49 @@ BBLK
...
+;; The KLH10P-marked code is a workaround for a bug in the 
+;; KLH10 implementation of the CH11. It seems to be a synchronization
+;; bug which results in packets being corrupted after being received
+;; from the net, and before being read by the 10.
+;; The workaround is to compute the checksum here, and verify it.
+;; When someone either fixes the CH11 bug or simply reimplements it,
+;; this code should be taken out again.
+;; Note that the performance impact is very small.
+;; /BV November 2024.
+;; (Side note: is "ADDI J,(B)" faster than "ADD J,B"?)
```
larsbrinkhoff, line 2439: On some machines, potentially yes.  The reason being that the immediate operand can use the possibly streamlined 18-bit addressing circuitry.  I believe it is a little faster on a KA10, or maybe a PDP-6.  Certainly this type of optimization can be seen in early MIT code.

larsbrinkhoff: I don't think ITS source code comments are the right place for a Q and A, though.

larsbrinkhoff (approved):

bictorv:
> Anyway, the missing `]` effectively removes the rest of the CHAOS code, so that's probably why.

I wish MIDAS would complain about unbalanced `]`s. I don't think it's the first time this happens...

PR #2343, opened 2024-11-21
Problem getting SENDS in DDT

larsbrinkhoff:
There seems to be a problem getting SENDS in DDT using the <kbd>^A</kbd> command.  For example:

```
↑A
(Getting SENDS for LARS@ )
 : .TEMP.; LARS SENDS - NO SUCH DEVICE
```

### DIFF
--- a/src/sysen1/ddt.1549
+++ b/src/sysen1/ddt.1549
@@ -954,9 +954,10 @@ bret:	    aos -1(p)
 mchok0:	camn a,[sixbit /DSK/]  ;is this the local machine?
 	  jrst [ move a,itsnam ? jrst popj1]  ;then use that instead
 	movsi b,-mchcnt		;for all the machines
-mchok1:	camn a,mchtab(b)	;is it this one?
+mchok1:	skipn mchtab(b)		;empty table entry?
+	  ret			;  yes, exit early
+	camn a,mchtab(b)	;is it this one?
 	  jrst popj1		;  yes, it's OK
-	skipe mchtab(b)
 	aobjn b,mchok1		;no, try next
 	ret
 


bictorv:
> There seems to be a problem getting SENDS in DDT using the ^A command. For example:
> 
> ```
> ↑A
> (Getting SENDS for LARS@ )
>  : .TEMP.; LARS SENDS - NO SUCH DEVICE
> ```

Doesn't happen on UP, which is a PI-based machine.

bictorv:
> > There seems to be a problem getting SENDS in DDT using the ^A command. For example:
> > ```
> > ↑A
> > (Getting SENDS for LARS@ )
> >  : .TEMP.; LARS SENDS - NO SUCH DEVICE
> > ```
> 
> Doesn't happen on UP, which is a PI-based machine.

UP uses the old code, with an explicit list of ITS names at `mchtab`, which it loops over, and where `mchcnt` is precisely its length. Now `mchcnt` is the *max* length of `mchtab`, instead of the number of entries in it. 

Your fix is a good one!

bictorv (approved):

eswenson1:
> @eswenson, I see ES has the same issue ~so I'm guessing it's something that affects the PI distribution as well~.
> 
> EDIT: Strike that, maybe ES just has the GitHub version.

Indeed, I updated SENDS from GitHub. I generally do this all the time. ES should be running the latest GitHub code. But if something existed in PI that hasn’t been included in GitHub, I leave it be.

eswenson1 (approved):

bictorv:
(One idea would be to check for corresponding code in other programs that got updated to use the ITSNMS-generated list of ITSes. I don't have time right now.)

eswenson1:
@larsbrinkhoff I've been using :QSEND to send messages to other machines.  DDT's SNDFLG allows the use of :SEND to invoke an external program of the same name, and our build causes SYS; TS SEND to be linked with SYSBIN;SENDER BIN.  However, when I try :send ejs@ex <message>^C from ES, I get "Can't", and the ">>" prompt.  When I try :send ejs@es <message>^C on EX, it sends perfectly find and I receive the message. On both systems, SYS;TS SEND is a link to SYSBIN;SENDER BIN, and BINPRT says that both were assembled from SYSEN1;SENDER 398.

Because :SEND ejs@ex <message>^C on ES always says "Can't", I figured that it didn't support the chaos SEND protocol.  However, as seen by the success of the same thing from EX, that was a bad assumption.  I've always used :QSEND rather than :SEND when sending messages to other machines (I used to do that back in my MIT-MC days as well).  

Can you think of any reason why this works from one machine (EX/DB) and not from ES?  QSEND, of course, works perfectly fine on both machines for sending to the other (and other chaosnet hosts).

eswenson1:
For whatever reason, on ES attempting to use :SEND to send to EX, SENDER attempts to use the CLI device, rather than CHAOS.  This fails (and :send prints "Can't").  However, if after the failure, I use the "K" command to switch from CLI to Chaos, it works fine.  On EX, on the other hand, it doesn't attempt to use the CLI device, but chooses CHAOS, which works.

I don't understand why on ES, SEND tries to use the %CLI device, but not on EX.  Any ideas?

larsbrinkhoff:
My first idea is that you have forgotten to set SNDFLG to -1 on ES.

You should be able to make CLI work by having all your "local' machines in ITSNMS _and_ setting op MLDEV between them.

larsbrinkhoff:
In that case I think it's TS SEND (SENDER) printing "Can't" (and maybe something more?)  I'm guessing it can't open EX:.TEMP.;uname SENDS for some reason.

eswenson1:
Well, it's not that.  I can, from ES, write to EX:.temp.;ejs sends (using the :copy command).  And I can do the same from EX (writing to ES:.temp.ejs sends).  MLDEV is working perfectly fine from either machine to the other.

larsbrinkhoff:
ATSIGN DEVICE does some magic to make device names like EXCLI work.  I see it has ES built in, but not EX.  Unless you fixed that locally?

larsbrinkhoff:
Björn

bictorv:
> Björn [ Björn ](https://private-user-images.githubusercontent.com/775050/390855539-770b8bcf-d217-4598-9054-3ea45e720f01.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzI4MTYxODAsIm5iZiI6MTczMjgxNTg4MCwicGF0aCI6Ii83NzUwNTAvMzkwODU1NTM5LTc3MGI4YmNmLWQyMTctNDU5OC05MDU0LTNlYTQ1ZTcyMGYwMS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQxMTI4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MTEyOFQxNzQ0NDBaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hZDk3ZTFiNzEzMWViMTY1OGEwZTFmYTRhMjY1MGRiMmE1MTQ0NTQ3YWUxY2FmYWNkNTY4NWI2NDJkYWFkNGZjJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.A_OmOwzHu2NpwJ4_ojQfVLe0MsKoxJwXZVBtnJjEGc8) [ ](https://private-user-images.githubusercontent.com/775050/390855539-770b8bcf-d217-4598-9054-3ea45e720f01.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzI4MTYxODAsIm5iZiI6MTczMjgxNTg4MCwicGF0aCI6Ii83NzUwNTAvMzkwODU1NTM5LTc3MGI4YmNmLWQyMTctNDU5OC05MDU0LTNlYTQ1ZTcyMGYwMS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQxMTI4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MTEyOFQxNzQ0NDBaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hZDk3ZTFiNzEzMWViMTY1OGEwZTFmYTRhMjY1MGRiMmE1MTQ0NTQ3YWUxY2FmYWNkNTY4NWI2NDJkYWFkNGZjJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.A_OmOwzHu2NpwJ4_ojQfVLe0MsKoxJwXZVBtnJjEGc8)

HTH, som man säger. :-)

PR #642, opened 2024-12-02
NFILE - New file server

larsbrinkhoff:
BAWDEN; NFILE 185  
NETDOC; RFC037 NFILE


larsbrinkhoff:
@eswenson1, you think this is ok to merge?

eswenson1:
I don’t see why not — considering it is a new program being added.

do we have any NFILE clients we can use to test that the server works? It might be good to prove that it works before committing.

larsbrinkhoff:
If I understand correctly, NFILE was a Symbolics protocol.  I haven't found anything else, so I assume a Symbolics Lispm would be the only client.  And I don't have one of those.

eswenson1:
I do have a Symbolics client, but I haven't put it on the chaosnet yet.  Maybe I'll try to do that.

eswenson1 (approved):

larsbrinkhoff:
Merged.  If bug fixes or updates are needed, they can be done separately.

PR #2361, opened 2025-02-11
Build script updates for running MIDAS

larsbrinkhoff:
Following up on #2360, here are more updates to streamline the build script.


eswenson1 (approved):
Nice. Love all the new TCL macros you’ve been adding.

larsbrinkhoff:
By the way, I'm also considering adding something for other regular patterns in the scripts: deleting files and making links, purifying and dumping programs, dumping with Maclisp, calling COMPLR.

eswenson1:
> By the way, I'm also considering adding something for other regular patterns in the scripts: deleting files and making links, purifying and dumping programs, dumping with Maclisp, calling COMPLR.

I think that is a great idea. Anything we do multiple times that consists of other than a trivial one liner is a good first candidate.

PR #799, opened 2025-02-13
Neater build process for JOTTO

larsbrinkhoff:
build.tcl's process to dump TS JOTTO is messy because it expects to read its dictionary from a DECtape rather than DSK. It also doesn't open the TTY channels when started at BEG0.

One fix would be to make it read a single filename string, and open it with SOPEN. (This is roughly what SAIL did with their version.)

Alternatively, we could just make it read the dictionary file whenever it's started; this only takes a second or so under emulation.


larsbrinkhoff:
I saw this was lying around, so I fixed it using ^T filename translation.

eswenson1 (approved):

PR #2363, opened 2025-04-08
Dockerize

mfgamallo:
* Fix `dependencies.sh` so it works in Ubuntu 24.04
* Generate a Docker image per emulator: `KLH10`, `SIMH`, `pdp10-ka`, `pdp10-kl` and `pdp10-ks`.
* Increase timeout in `lisp.tcl` because half of the time the workflow would fail to build the image for `KLH10`
* Push images to GitHub Packages <-- WARNING! I only tested this in my repo!

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -1,0 +1,26 @@
+name: Build Docker
+
+on: [push, pull_request]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build-docker:
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04]
+        emulator: [klh10, simh, pdp10-ka, pdp10-kl, pdp10-ks]
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+      - run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+      - name: Build Docker image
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
Great pull request!  However, the deployment fails:
```
ERROR: invalid tag "ghcr.io/PDP-10/its-klh10:e19bd829e50d9cd9b0f63d740ae42cc1ca2add7d": repository name must be lowercase
```

eswenson1 (approved):
I was able to build and run the docker container successfully.  I didn't set up networking, so I don't know whether that works, but I was able to get to DSKDMP and list the files in the "." directory.  I was able to start up ITS and see messages from the operator console.  So I'm guessing it would work just fine.

larsbrinkhoff (changes requested):
Please look into why deployment fails.

mfgamallo:
I see! I naively used the name of the git repo as name for the Docker repo. My first impulse is to just lowercase said name, but let me take a look at the requirements, just in case there are other limitations (like no hyphens)

mfgamallo:
Aha. Looks like repos must follow the rule
```
Only contain lowercase letters, numbers, hyphens (-), and underscores (_)
```
(Taken from [these docs](https://docs.docker.com/docker-hub/repos/create/))

Cool, in that case using the name of the repo in lowercase should work. Let me fix that.

mfgamallo:
@eswenson1 Great news! Thanks for trying it out :-)

mfgamallo:
Hmmm. I don't seem to be able to add the new commits to this pull request. Weird, I would have sworn I did something like that in the past.
Well, never mind. I can close this PR and open a new one.

PR #2364, opened 2025-04-12
Fix stamp.sh error calling git

kahlkevin:
- Problem: $1 is empty string inside shell function "timestamps"
 - Solution: Use $list global computed previously

### DIFF
--- a/build/stamp.sh
+++ b/build/stamp.sh
@@ -18,7 +18,7 @@ timestamps(){
             t="$2"
             TZ=EST
         else
-            t=`git log -1 --date=format:%Y%m%d%H%M.%S --format=%cd -- "$1"`
+            t=`git log -1 --date=format:%Y%m%d%H%M.%S --format=%cd -- "$i"`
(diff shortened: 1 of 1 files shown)


kahlkevin:
Or, instead was the intent that the timestamp match the most recent commit on a _per file_ basis?

If so, I can patch this pull request (or submit a new one) to change `$list` to `$i`. Please advise if this is the maintainers' preference.

(Either way, the current `$1` is incorrect and leads to error output during make)

larsbrinkhoff:
Yes, I think `$i` was intended.  Thanks, good catch!

kahlkevin:
> Yes, I think `$i` was intended. Thanks, good catch!

Okay, I've made the change here (via rebase/force-push so we still only have a single commit).

Once you're happy with validation, I look forward to your approval and completion of the pull request.

larsbrinkhoff (approved):

larsbrinkhoff:
Merged, thanks!

PR #2365, opened 2025-04-14
Dockerize

mfgamallo:
* Fix `dependencies.sh` so it works in Ubuntu 24.04
* Generate a Docker image per emulator: `KLH10`, `SIMH`, `pdp10-ka`, `pdp10-kl` and `pdp10-ks`.
* Increase timeout in `lisp.tcl` because half of the time the workflow would fail to build the image for `KLH10`
* Push images to GitHub Packages <-- WARNING! I only tested this in my repo!

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -1,0 +1,26 @@
+name: Build Docker
+
+on: [push, pull_request]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build-docker:
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04]
+        emulator: [klh10, simh, pdp10-ka, pdp10-kl, pdp10-ks]
+    permissions:
+      contents: read
+      packages: write
+    steps:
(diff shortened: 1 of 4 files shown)


mfgamallo:
Ah, I can see it failed again
```
denied: installation not allowed to Create organization package
```
Right, I don't know enough about organizations. Let me see if I can create a testing organization local to me with which I can play to learn what needs to be done in order to be able to push to one :+1:

mfgamallo:
> Ah, I can see it failed again
> 
> ```
> denied: installation not allowed to Create organization package
> ```
> 
> Right, I don't know enough about organizations. Let me see if I can create a testing organization local to me with which I can play to learn what needs to be done in order to be able to push to one 👍

Alternatively, I may end up just creating a free account in Docker Hub and upload the images there. Whatever is easier :slightly_smiling_face:

mfgamallo:
Hi! :wave: Sorry for the huge delay. I created an organization of my own and was able to push Docker images to it. I needed to enable Public Package Creation permission, though.
Can you tell me if that permission exists and is sett for `PDP-10`?
In particular, I'm looking for this section:
Screenshot_20260218_224336

PR #2369, opened 2025-09-04
Adding two versions of HAKMEM.

eswenson1:
HAKMEM 141 seems to be later version, with a date in 1980. HAKMEM 17 appears older, but different in some respects.

Note: The timestamps on both of these files are suspect.  The date on HAKMEM 141 appears to be Jan 1, 1980 (no HHMMSS), and the date on HAKMEM 17 was in the year 2009 (which is probably the year it was extracted from ToTS).  I cannot find the correct date/time for this file.


eswenson1:
I’ll update the date, but I definitely saw 15 somewhere.

PR #1658, opened 2026-04-13
COLOR HACK

larsbrinkhoff:
PDP10; COLOR HACK  
PDP10; COLOR NOISE  
MB; COLOR 23

Uses the color scope.


larsbrinkhoff:
Another copy of the source code:  
https://github.com/larsbrinkhoff/mit-logo-and-sits-raw-files/blob/master/files/3100052/logo/color.23

larsbrinkhoff:
Checks whether it runs on PDP-6 or PDP-10.  If on 10, does .IOTLSR for running in timesharing.

eswenson1 (approved):

larsbrinkhoff:
Some of the hacks use switches as input.  However, it's not the regular console switches, but the "device 420" switches.  The program CONOs 40, so the Spacewar consoles and a bank of 36 switches (BIT SW'S below) can be used.  The latter is not implemented in the emulator.  If you have a PiDP-10, it's probably convenient to edit the program and replace all DATAI CSOPE with DATAI APR, and then you can use the console switches.

image

https://github.com/larsbrinkhoff/its-archives/blob/master/ailab/ITS_Hardware_Memo_6.pdf

PR #2387, opened 2026-06-10
Updated comsat to correctly initialize DOMGAT from COMSAT-DOMAIN-GATEWAY

eswenson1:
Previously, TCPGAT was initialized from *COMSAT-TCP-GATEWAY (correctly), and then replaced with the value from *COMSAT-DOMAIN-GATEWAY, leaving DOMGAT uninitialized.


larsbrinkhoff (approved):
Looks fine according to the context in the code.

eswenson1:
I discovered this while trying to use the feature when I changed VPSes for ES.

PR #2390, opened 2026-06-15
Need to link in JOBDAT in DECUUO programs, and use /B LOADER switch.

larsbrinkhoff:
I noticed TS BASIC only has the DECUUO symbols, but not any symbols from BASIC or HELPER.  I wanted those symbols to debug why BASIC can't find its .HLP file.  This is despite the S command being used (implicitly by D) in DECUUO to get the TOPS-10 symbol table and convert it to an ITS symbol table.  Checking DECUUO, it apparently takes the symbols from .JBSYM, location 116.  That location is always 0.  Running LOADER, I see it always leaves 116 zero.  Saying /S to LOADER doesn't help, it's still zero.  I tried to find a manual for loader, but it doesn't seem to be anywhere online.  I resorted to finding the complete list of commands from a listing: https://pdp-10.trailing-edge.com/decuslib10-04/01/43,50322/loader.lst.html  The /U command provided the help I needed: listing all undefined symbols.  This can be run anytime during the linking process, but doing it at the end provided the crucial clue: .JBSYM was undefined.  Linking in JOBDAT provides a definition.  HOWEVER, that still doesn't make 116 point to a symbol table.  Scanning the command list again, I found /B - "blocks down symbol table".  That sounds promising.  Adding it to the LOADER command line does the trick; now 116 has an interesting value.  And after running the DECUUO D command, :LISTS lists the BASIC and HELPER symbols.


larsbrinkhoff:
The /LINK switch for making an in-core symbol table is /SYMSEG.

I had to add JOBDAT.MAC and move the Fortran games down in the build sequence.

larsbrinkhoff:
@eswenson1, if you check the build logs for this pull request, you can see that the ADVENT and TREK games written in Fortran has undefined symbols.  Search for "undefined global(s)".  I'm guessing there is some library needed to resolve those.  Thoughts?

eswenson1:
I can look at this tomorrow. I assume we’ve always had these unresolved references, but not noticed them because we haven’t built with the new options you are now using.

larsbrinkhoff:
Right, we always had them.  Maybe they are not used, so might not be a problem.

larsbrinkhoff:
Additional build related fixes.

MACRO calls out to .HELPR, so needs linking with HELPER.  Typing /H should print the help text in MACRO.HLP, so I added that file too.

MACN11 uses some JOB symbols which are not defined in JOBDAT, ~~but in DECDFS.~~  I added the appropriate links for DECBTS and DECDFS, and added the DECSYS;DECBTS 232 file.

EDIT: DECDFS is a MIDAS file, not suitable for MACRO.  The undefined in MACN11 remain unresolved for now.

eswenson1:
Do we have sources for all the link-related DECSYS files (link.shr, lnklod.shr, lnkscn.shr, linxit.shr), and the loader? It would be nice to build those from source too like you've done for jobdat.rel.

eswenson1 (approved):

larsbrinkhoff:
I usually search like this: https://pdp-10.trailing-edge.com/cgi-bin/searchbyname?name=lnklod.mac

So yes, LINK sources seem available.  I vaguely recall trying to build it, but giving up... or maybe I misremember.  Anyway, historically DECSYS was mostly populated with binaries, so I haven't been vigilant in trying to build everything there from source.

PR #2391, opened 2026-06-17
Add help file for LOADER.

larsbrinkhoff:
I have been unable to find any manual for LOADER, so I made this.  It was created by looking at the LOADER.MAC source code, and written in the same style as MACRO.HLP.

FYI @jayjwa

### DIFF
--- a/doc/decsys/loader.hlp
+++ b/doc/decsys/loader.hlp
@@ -1,0 +1,30 @@
+Switches are :-
+/< beginning of overlay
+/= put out chain retaining symbols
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Maybe I should have looked harder, because here is the LOADER manual: https://bitsavers.org/pdf/dec/pdp10/1970_PDP-10_Ref/1970PDP10Ref_Part5.pdf

Anyway, the .HLP file is useful.

eswenson1 (approved):

Issue #853, opened 2018-04-30
The 2027 problem


atsampson:
From MC:SYSDOC;ITS RECENT:

> Date: Mon, 26 Oct 87 11:58:03 EST
> From: Alan Bawden
> [...]
> The other, more common, date format is the one used by the filesystem, and by programs that used the DATIME library; this is the format returned by the RQDATE system call.  This format packs the year, less 1900, in a 7-bit field, so this will last through 2027.  Since the bit to the left of the year field is unused, we can easily expand this to last through 2155.

It's looking like ITS will be alive and well come 2027, so this'll need fixing at some point.

Things that will need fixing (very incomplete!):

- [ ] SYSTEM
- [ ] SYSDOC; .CALLS
- [ ] TJ6
- [ ] TECO's FS FDCONV

larsbrinkhoff:
I looked at using eight or nine bits for the year field.  For eight, it's generally agreed that it's probably no problem.  For nine, I didn't see anything obviously damaging, but who knows.

Anyone volunteering to run a system with the clock set 300 yeas ahead?

atsampson:
The system's idea of the date is stored in QDATE, which is updated by CLCQDT (in TIME) and read along with TIMOFF by various syscalls in DISK (e.g. ARQDAT for .RQDATE). It looks like it would be straightforward to make CLCQDT write the year into 4.9-4.1 rather than 4.7-4.1.

The filesystem doesn't appear to use bits 4.8 or 4.9 in UNDATE or UNREF for anything.

However, QDATE is set to -1 on boot to indicate that the time isn't known yet, and there are a few tests that do things like `SKIPL A,QDATE`. These would need adjusting to test for -1 specifically if we wanted to use 4.9.

larsbrinkhoff:
Thanks for checking.  There's also a chance some user program may use a negative value to signal a special condition.

atsampson:
SYSDOC; .CALLS > will need updating for this; at the same time, we can fix the various places where it says that the year field is "mod 100" when it's actually "minus 1900".

larsbrinkhoff:
I set my TT ITS clock to 2028 and made a file with CREATE.  It got timestamped 1928.

larsbrinkhoff:
Must have clicked the wrong button.

larsbrinkhoff:
The TV-11 code claims it's limited to a 7-bit year.

larsbrinkhoff:
SYSENG; DATIME 74 has this:

```
[...]

Issue #1882, opened 2020-05-04
Y2K bug in Macsyma plot


larsbrinkhoff:
Guessing this is in APLOT2 function `date`.

![](https://user-images.githubusercontent.com/775050/80988232-7bd11e00-8e33-11ea-830b-4c22776102ed.png)

larsbrinkhoff:
Cleverly hidden here:

```
                              '(44. 32. 49. 57.) (explodec (car date))))
```

The four numbers are ASCII codes for ", 19".  The year comes from `(STATUS DATE)` which now returns 2020.  I believe it used to return year-1900.

larsbrinkhoff:
Fixed in #2392.

larsbrinkhoff:
Confimed fix.

<img width="620" height="459" alt="Image" src="https://github.com/user-attachments/assets/9ea7dac4-93a6-47cc-bd43-3bed6727e558" />

PR #2392, opened 2026-06-18
Fix Y2K bug in XGPSPL.

### LINKED ISSUE #2368: XGP output heading date is 100 years behind current date
The heading in the year field of the XGP output lists the year as 1925, versus 2025, even though the system time is correct. On June 5th I sent the following output to the printer. Result is the attached PNG. Note heading. 

```
*:print xgptst >

*:time
The time is 18:23:50 EDT.
Today is Thursday, the 5th of June, 2025.
KA ITS 1652 has run for 2 days, 6 hours, 55 minutes, 27 seconds.
Surpassing all previous KA records for uptime!
*
```

Image

larsbrinkhoff:
This should fix #2368.

I pre-emptively made the year field 9 bits in anticipation of the #853 fix.

### DIFF
--- a/src/share/aplot2.301
+++ b/src/share/aplot2.301
@@ -872,7 +872,7 @@
 						      (10. . "October")
 						      (11. . "November")
 						      (12. . "December")))))
-			      '(44. 32. 49. 57.) (explodec (car date))))
+			      '(44. 32.) (explodec (car date))))
 		     (t (append (exploden (caddr date)) '(46.)
 ;					    ((lambda (base)
 ;						     (mapcar (function
--- a/src/dcp/versa.212
+++ b/src/dcp/versa.212
@@ -945,7 +945,8 @@ gflag1:	movei x,[ [asciz /~%~T Processing DSK:.GLPR.;~6 ~6/]
 
 	move x,cdate
 	move a,p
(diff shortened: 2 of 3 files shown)


larsbrinkhoff:
Also made simlar fixes in VERSA and APLOT2.

src/share/aplot2.301
```diff
@@ -872,7 +872,7 @@
 						      (10. . "October")
 						      (11. . "November")
 						      (12. . "December")))))
-			      '(44. 32. 49. 57.) (explodec (car date))))
+			      '(44. 32.) (explodec (car date))))
```
eswenson1, line 875: What does this do?

larsbrinkhoff: It makes a list of ASCII characters.  44 32 49 57 are `, 19` so I dropped 19.  `(car date)` has the four digit year.

eswenson1 (approved):

larsbrinkhoff:
The APLOT2 bug is #1882.

larsbrinkhoff:
I have now checked that the fixes for XGPSPL and APLOT2 work.  I have not checked VERSA, because I don't know how to use it.

